PR #2259, opened 2015-01-04
Feature/wp9

matiu:
many wp releated fixes


matiu:
updated and ready for review.

PR #2286, opened 2015-01-07
Read and agree the terms before continue using Copay

### LINKED ISSUE #2284: Disclaimer before opening Copay
* Read and accept these terms

cmgustavo:
* Only for mobile devices (fixes #2284)


matiu:
Code looks good. I would like to  test it on different devices before merging it.

thoatbk:
Would make sense to make the text translateable, but for that the language selection should appear on the first login also and not be hidden in the settings.

PR #2310, opened 2015-01-14
Removes "Manage wallets" and moves info to "Settings"

### LINKED ISSUE #2250: Entering 'My Profile' tab before all wallets are loaded will show partial list of wallets
It should show all wallets and maybe some indication that the wallet is being loaded.

image

cmgustavo:
* Removes "Manage wallets"
* Wallet info inside "Setting"
* Backup/Delete wallet inside "Setting"
* Fix layout "Setting" page
* Fix PIN create and error message
* Fix spinner when signing, creating profile and logout

Fixes #2250 #2296


matiu:
Great work!!

PR #2366, opened 2015-01-29
Improve response time when press menu icon...

### LINKED ISSUE #2362: Mobile: use "ng-click" in the sidebar and to close popup
To avoid delay in mobile devices

cmgustavo:
and Remove transition effects when closing modals.

fixes #2362

### DIFF
--- a/index.html
+++ b/index.html
@@ -85,7 +85,7 @@
         <div ng-controller="SidebarController" ng-show="$root.iden && !$root.hideNavigation  && !$root.wpInputFocused">
           <nav class="tab-bar">
             <section class="left-small">
-              <a class="left-off-canvas-toggle menu-icon" ><span></span></a>
+              <a href class="menu-icon" ng-click="openMenu()"><span></span></a>
             </section>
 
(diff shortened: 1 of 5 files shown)


index.html
```diff
@@ -129,7 +129,7 @@ <h1 class="title ellipsis" ng-show="$root.iden && $root.wallet  && !$root.hideWa
           ng-class="{'main':$root.iden && !$root.starting && !$root.hideNavigation  && !$root.wpInputFocused}" 
           ng-view></section> 
 
-        <a class="exit-off-canvas"></a>
+        <a class="close-menu" ng-click="swipe()"></a>
```
matiu: we should decide whether wi will call swipe() functions or openMenu() functions on the HTML. It does not make sense to call swipe() and openMenu(). It should be `openMenu()` and `closeMenu()`.

cmgustavo: Updated. Thanks for your suggestion.

cmgustavo:
Updated

cmgustavo:
Updated

PR #2368, opened 2015-01-29
WIP: Fix rounding error in amount field

cmgustavo:
Javascript rounding error. The following operations should be equals:

```
101808500 * 1 / 100000000 = 1.018085
```
But not...
```
101808500 * 0.00000001 = 1.0180850000000001
```

### DIFF
--- a/js/controllers/send.js
+++ b/js/controllers/send.js
@@ -83,6 +83,7 @@ angular.module('copayApp.controllers').controller('SendController',
     $scope.setInputs = function() {
       var w = $rootScope.wallet;
       var unitToSat = w.settings.unitToSatoshi;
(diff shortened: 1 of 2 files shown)


matiu:
Please:

1) add test to prevent this to get broken one more time
2) remove `satToUnit` if it is not longer used.

cmgustavo:
Updated!

1.- Tests added. 
2.- Unused variable removed.

Note:
All floating point math is based on the IEEE 754 standard. JavaScript uses 64-bit floating point representation, which is the same as Java's double.

matiu:
Awesome, great work

MeniRosenfeld:
It seems the problem still/again exists.

image

PR #2379, opened 2015-01-31
Important fix: paste into input and textareas

cmgustavo:
The previous CSS code disables text selection on all screen but not allowed paste text on input and textareas. This fix enable the copy to inputs and textareas.


matiaspando:
Works great on Android!

PR #2407, opened 2015-02-09
fix incomple fix

matiu:
- fix incomplete wallet's forced sync.


cmgustavo:
ACK, all tests passed. 
Manual test with complicated wallets 2-3 and 2-2 OK

PR #2409, opened 2015-02-10
Pin screen

bechi:


cmgustavo:
LGTM. Tested with wp8.1, iphone4s and 6plus.

PR #2431, opened 2015-02-25
Integration of Wallet Name Service lookup

wdawg33:
- Implemented Wallet Name Service lookup
- Unit tests created for Wallet Name Service lookup

### DIFF
--- a/config.js
+++ b/config.js
@@ -92,6 +92,11 @@ var defaultConfig = {
     // clientId: '1036948132229-biqm3b8sirik9lt5rtvjo9kjjpotn4ac.apps.googleusercontent.com',
   },
   */
+
+  walletNameLookup: {
+    enabled: true,
+    url: 'https://pubapi.netki.com/api/wallet_lookup/'
+  }
 };
 if (typeof module !== 'undefined')
   module.exports = defaultConfig;
--- a/js/controllers/send.js
+++ b/js/controllers/send.js
@@ -3,7 +3,7 @@ var bitcore = require('bitcore');
 var preconditions = require('preconditions').singleton();
 
 angular.module('copayApp.controllers').controller('SendController',
-  function($scope, $rootScope, $window, $timeout, $modal, $filter, notification, isMobile, rateService, txStatus, isCordova) {
+  function($scope, $rootScope, $window, $timeout, $modal, $filter, notification, isMobile, rateService, txStatus, isCordova, walletNameLookupService) {
 
     $scope.init = function() {
       var w = $rootScope.wallet;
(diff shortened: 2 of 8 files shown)


martindale:
This is an interesting proposal.  Can you please provide a more detailed description of what this is intended to accomplish?

thoatbk:
We made an open protocol like that, called OpenAlias, its already supported by Electrum.

Check openalias.org for details, and maybe take a look at the electrum plugin which got merged into their master branch: https://github.com/spesmilo/electrum/blob/master/plugins/openalias.py

That might be worth implementing for copay as well as it doesnt rely on third party services, it uses DNSSEC and DNSCrypt for security.

fluffypony:
NACK, closed + centralised ripoff of the [OpenAlias standard](https://openalias.org)

martindale:
After review, I am disinclined to allow this into Copay as it depends on an outside web service.  This will cause information leakage for Copay users, as an API call is made for addresses they are viewing or using.  This is not an acceptable behavior for our intended audience.

Can an alternative be provided that is decentralized, and/or can be derived from a local blockchain?  OpenAlias seems like a reasonable approach that deserves further investigation.

fluffypony:
@martindale OpenAlias resources are spread thin as we're busy setting up DNS seeders for Namecoin Core, but will see if someone can build out OA support for CoPay after March 7th.

martindale:
Hey @wdawg33, we're going to keep this one open for now – addressing the above feedback will be needed before we consider merging it.

This would be a great opportunity to implement a plugin system where these types of integrations become opt-in for the user.  Discussion open.

wdawg33:
Thank you to everyone for taking the time to review our request and provide your feedback. Netki is a company that aims to solve the problem of unmemorable wallet addresses by developing, contributing, and supporting a decentralized standards based lookup. Similarly to OpenAlias, we believe privacy and security are as crucial to solving this problem as the problem itself. We offer our public API as a quick means of getting started, however we have published resolvers both for ICANN based domains and Namecoin’s blockchain, as well as a Python based API server to quickly get started with hosting your own decentralized lookup service.

DNSCrypt penetration is limited today, but we would love to help push this as a further means of privacy. Where we differ is we want to utilize technology that is deployed today such that we use the power of DNS deployment across the internet for redundancy, reliability, and scale. DNSSEC is the mechanism to ensure the chain of trust. Our belief is that lookups should be powered by the existing DNS infrastructure at large. We plan to include other established open protocols into our products to enhance user privacy and will update our open source offerings. 

We are going to drop this pull request and resubmit as an optional Bitcore plugin so that folks who would like to use this service have that option.

PR #2510, opened 2015-04-13
Feat/networking01

matiu:
- handle reconnections.


cmgustavo:
Manual test with WP: OK

PR #2529, opened 2015-04-16
Fix URI-payment for mobile

cmgustavo:


cmgustavo:
Only works on iOS and Android.

PR #2530, opened 2015-04-16
ignorePayPro verification in chrome

matiu:
- ignorePayPro verification in chrome, so chrome users can sign PayPros generated from other devices.


cmgustavo:
LGTM. Manual test: OK

PR #2535, opened 2015-04-17
Fixing barcode scanner (Android) with new cordova plugin

matiaspando:


cmgustavo:
ack. Manual test on iOS: OK

PR #2537, opened 2015-04-17
Layout details 03

bechi:


public/views/send.html
```diff
@@ -27,24 +27,12 @@ <h4 class="title m0">
...
-          <div class="box-notification" ng-show="send.error">
-            <div class="box-icon error">
-              <i class="fi-x size-24"></i>
-            </div>
-            <span class="text-warning size-14">
+          <div class="box-notification" ng-show="send.error" ng-click="send.resetError()">
+            <span class="text-warning">
               {{send.error|translate}}
             </span>
             <a class="close-notification  text-warning" ng-click="send.resetForm(sendForm)">&#215;</a>
```
cmgustavo: The same function is called twice

Issue #2561, opened 2015-04-19
Inviting copayers 


ajp8164:
The idea of sharing a secret is kind of intimidating and as a casual user you might be concerned with sharing "secrets" via email etc.  recommend this be relabeled to something like "Invite users" or "Invite copayers" or similar.  Remove word "secret" from screen.

dabura667:
+1

My wife actually just asked me that:

"ok, I signed up, but do I send you the secret or do I keep it secret???"

I was about to go into the difference between asymmetric and symmetric keys... but then just decided to say "it's our secret, so gimme :-P"

martindale:
+1 on this as well.  I always move through this stage really quickly when showing people how Copay works, because it's where people get stuck.

Maybe there's also an opportunity here to do local discovery on a LAN, or something – other Copayers can turn on "discovery mode", and when creating a wallet, you can see a list of these people and add them directly.

Issue #2553, opened 2015-04-19
Master extended private key setting


ajp8164:
0.10.3 - is not a user friendly setting name. I have no idea what this is for and suppose most users won't.

matiu:
Where you see that setting? In "Advanced" at wallet creation?

On Sun, Apr 19, 2015 at 9:26 AM, Andy Phillipson <notifications@github.com>
wrote:

> 0.10.3 - is not a user friendly setting name. I have no idea what this is
> for and suppose most users won't.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitpay/copay/issues/2553>.
>



-- 
BitPay.com

ajp8164:
Yes

PR #2569, opened 2015-04-20
better wording for wallet invitations

matiu:
fixes https://github.com/bitpay/copay/issues/2561

and fixes https://github.com/bitpay/copay/issues/2553

### DIFF
--- a/public/views/create.html
+++ b/public/views/create.html
@@ -61,6 +61,7 @@
             </label>
             <label for="ext-master" class="m10t">
               <span translate>Master extended private key</span>
+              <small translate>If not given, a random key will be generated</small>
               <input id="ext-master" 
               type="text" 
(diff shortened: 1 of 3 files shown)


public/views/create.html
```diff
@@ -61,6 +61,7 @@
             </label>
             <label for="ext-master" class="m10t">
               <span translate>Master extended private key</span>
+              <small translate>If not given, a random key will be generated</small>
```
dabura667: I could not find this string in the template.po file.

I am assuming this is because this fix and the templates.po fix were done on separate branches based off a commit without these changes.

cmgustavo: Yes. The template.pot files is out-of-date. I think we 're going to add more words in the following of days and updating the pot file. I know, it 's a hard work to maintenance, but we are currently changing words almost every days to do that all users understand Copay better.

PR #2619, opened 2015-04-24
Bug/legacy chromeimport

matiu:
- better client error message
- fix error when importing legacy local profiles in chrome app


cmgustavo:
ACK

PR #2625, opened 2015-04-24
Cordova file api 

matiu:
Replaces localStorage with device's file storage (on Android / iOS mobile devices)


cmgustavo:
ACK
Tested with: iOS 8.3 and WP8.1

PR #2638, opened 2015-04-26
Feat/gen addr anim

matiu:
-  'generating address' now only shows on 'receive' tab
 - better animation (no longer uses spinner dialog)


cmgustavo:
ACK

PR #2644, opened 2015-04-27
write file not as blobs

matiu:


matiu:
ack'ed offline by @matiaspando

Issue #2690, opened 2015-05-07
Balance Refresh


jodyrb:
How does the wallet balance refresh? Could there be a refresh button or, on the mobile apps, the ability to "pull down" to refresh? Thanks.

matiu:
+1 on this!

Currently the balance is refreshed with a wallet tx is detected thru a
socket connection to bitcore-wallet-server or when the connection is
interrupted or restored, but a manual trigger would be also fine, IMO.

On Thu, May 7, 2015 at 10:33 AM, jodyrb <notifications@github.com> wrote:

> How does the wallet balance refresh? Could there be a refresh button or,
> on the mobile apps, the ability to "pull down" to refresh? Thanks.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitpay/copay/issues/2690>.
>



-- 
BitPay.com

Issue #2617, opened 2015-04-24
Wrong error message when broadcast fails


matiaspando:
It should say something like 'Broadcast failed, please retry' instead of ETIMEDOUT

![screenshot_2015-04-24-11-16-04](https://cloud.githubusercontent.com/assets/2073486/7320656/58c5e7e6-ea74-11e4-8104-c52d248d2f50.png)

cmgustavo:
closes #2661

PR #2706, opened 2015-05-12
Feat/tap reload

matiu:
- Fixes https://github.com/bitpay/copay/issues/2690
 - Fixes update after delete TX
 - Fixes https://github.com/bitpay/copay/issues/2617


cmgustavo:
LGTM

PR #2755, opened 2015-05-20
Feat/create tab

### LINKED ISSUE #2753: personal wallet UX improvements
A few issues for personal wallet (non-multisig) UX:

## More intuitive personal wallet setup
In the `Create new wallet` view, could we add tabs (or some type of selection) for:

`Personal Wallet`  |  `Shared Wallet`

Personal Wallet could be selected by default, no further options necessary. 

If the user taps “Shared Wallet”, it could reveal the `Your Nickname` field and set the copayers to 2 of 3 (like the current default).

## Eliminate use of phrase `1 of 1`
We can remove all references to `1 of 1` throughout the app – if the user only has personal wallets, there's no need to complicate the view with references to multisig.

In the menu, the `1 of 1` subtitle can simply be removed, and the wallet title centered vertically. In the wallet `Home`, the `1 of 1` text can be replaced with the wallet name and centered vertically.

(For multisig wallets, mentioning the wallet's "security policy" – like `2 of 3` – removes user-uncertainty. :+1: )

cmgustavo:
fixes #2753

### DIFF
--- a/public/views/create.html
+++ b/public/views/create.html
@@ -21,15 +21,13 @@
     </div>
   </div>
 
-  <div class="row m20b">
+  <div class="create-tab">
     <a href 
-      class="small-6 medium-6 large-6 columns text-center" 
-      ng-class="{'text-gray': totalCopayers == 1}"
-      ng-click="create.setTotalCopayers(1)">Personal</a>
+      ng-class="{'selected': totalCopayers == 1}"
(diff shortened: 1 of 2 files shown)


matiu:
great job!!!

cmgustavo:
Thanks! Updated

PR #2770, opened 2015-05-22
Show memo in log

matiu:
depends on https://github.com/bitpay/bitcore-wallet-client/pull/79


matiu:
update, ready for review.

PR #2780, opened 2015-05-28
remove beta on splash

bechi:


cmgustavo:
ACK

Kirvx:
Sorry to intervene, but maybe it makes sense to change the icon color with the new blue color, no ?
Just asking.

matiu:
@Kirvx you are more that welcomed!

This PR should only remove the 'beta' tag in the banners, no design changes.

Maybe @bechi could respond you about the color suggestion you have.

Kirvx:
Thanks :)
You changed the color with a lighter blue, and left the older blue for https://github.com/bitpay/copay/blob/master/cordova/android/res/drawable/icon.png 
Anyway I tried to replace the color of the icon with the lighter blue and we don't really see much difference because of the size of the icon.

So you can forget what I said, sorry ^^

bechi:
@Kirvx thanks for your feedback.

The lighter blue is only for the wallets, and the older blue (dark blue) is the primary color for Copay. Anyway I will add more color for the wallets and also I will change the default color.

Thanks again :)

PR #2797, opened 2015-05-30
Fixed Japanese

dabura667:


matiu:
Thanks a lot for the contribution!!

dabura667:
Thanks a lot for all the hard work.

Just a little more push until 1.0.0

Then all that's left is maybe working on brushing up the usability based on user feedback.

The only problem with Copay is that there are soooo many ways you could do backups etc. and some of those are more reasonable for smaller wallets and some more reasonable for larger wallets etc.

It's a tough job to make decisions that some people will dislike for the sake of actually getting a product out. (trust me, that's like the story of my life) so I really do appreciate it.

The least I can do is pimp it to my peeps in Japan, and in order to do that, this must get done.

Keep up the good work, and let me know if you ever come to Japan.

matiu:
thanks a lot for your comments! we are really close to 1.0.. regarding your
question about iOS (on other thread), it will be there really soon.

I agree on the backups, we did not nail a great slution yet. we need to
think more and have more user feedback.

Sure we will let you know if we go to Japan!, and you let us know if you
came to Argerntina or Atlanta!

On Sat, May 30, 2015 at 11:20 AM, dabura667 <notifications@github.com>
wrote:

> Thanks a lot for all the hard work.
>
> Just a little more push until 1.0.0
>
> Then all that's left is maybe working on brushing up the usability based
> on user feedback.
>
> The only problem with Copay is that there are soooo many ways you could do
> backups etc. and some of those are more reasonable for smaller wallets and
> some more reasonable for larger wallets etc.
>
> It's a tough job to make decisions that some people will dislike for the
> sake of actually getting a product out. (trust me, that's like the story of
> my life) so I really do appreciate it.
>
> The least I can do is pimp it to my peeps in Japan, and in order to do
> that, this must get done.
>
> Keep up the good work, and let me know if you ever come to Japan.
>
> —
> <https://github.com/bitpay/copay/pull/2797#issuecomment-107044203>.
>



-- 
Matías Alejo Garcia
@ematiu

PR #2822, opened 2015-06-03
Updates description for stores

cmgustavo:
We had to do a full update of the description. Ping @Kirvx @dabura667


dabura667:
Sounds good, brb

Issue #2828, opened 2015-06-03
Button is mixing 3 strings in English order


dabura667:
![screenshot_2015-06-04-03-18-07](https://cloud.githubusercontent.com/assets/6586275/7967939/72921b0e-0a68-11e5-9fe7-abf541f08a99.png)

Create x-of-y wallet

In Japanese must be
x-of-y wallet create

How should I remedy this?

matiu:
that shouldnt be a problem, we should translate with place holder +
sprintf, something like
"Create %d of %d wallet"

dabura667:
https://github.com/bitpay/copay/blob/master/public/views/create.html#L109

I found the button. (I am not sure if there are similar constructs in other places.)

However, I am not good with html and css and I have no clue how to fix this.

If someone could to something similar to:
`"Create %d of %d wallet"` and generate a template.pot file from that and commit it, I would appreciate it.

PR #2845, opened 2015-06-04
Fixed String for Japanese word order.

dabura667:
this removed need for singular string "wallet"

Also, I am having trouble previewing this. If someone else could tell me whether this works or not, I would appreciate it.


dabura667:
#2828 will be fixed by this.

cmgustavo:
It works great! Thanks!

PR #2875, opened 2015-06-10
Clean last address and scann all addresses

### LINKED ISSUE #2871: When changing BWS setting, update address indexes in new BWS
It could happen that the new BWS has 'old' indexes,
so money transfered to cached address will be in a 'future' address.

Also, the address cache new to be cleaned with BWS setting is changed.

cmgustavo:
fixes #2871

### DIFF
--- a/src/js/controllers/index.js
+++ b/src/js/controllers/index.js
@@ -625,21 +625,6 @@ angular.module('copayApp.controllers').controller('indexController', function($r
     storageService.setCleanAndScanAddresses(function() {});
   });
 
-  $rootScope.$on('Local/CleanAndScanAddresses', function(event) {
-    storageService.getCleanAndScanAddresses(function(err, val) {
-      if (val) {
-        $log.debug('Clear last address cache and Scan');
-        var wallets = profileService.walletClients;
-        for (var walletId in wallets) {
-          storageService.clearLastAddress(walletId, function(err) {
-            self.startScan(walletId);
-          });
-        }
-        storageService.removeCleanAndScanAddresses(function() {});
-      }
-    });
-  });
-
   $rootScope.$on('Local/WalletCompleted', function(event) {
     self.setFocusedWallet();
     go.walletHome();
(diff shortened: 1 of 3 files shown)


src/js/controllers/index.js
```diff
@@ -549,9 +548,8 @@ angular.module('copayApp.controllers').controller('indexController', function($r
     c.startScan({
       includeCopayerBranches: true,
     }, function(err) {
-      if (err) {
-        if (self.walletId == walletId)
-          self.setOngoingProcess('scanning', false);
+      if (err && self.walletId == walletId) {
```
matiu: Not sure about this change. If an error occur and the wallet is focused it should be shown.

src/js/controllers/index.js
```diff
@@ -624,6 +622,22 @@ angular.module('copayApp.controllers').controller('indexController', function($r
 
   $rootScope.$on('Local/BWSUpdated', function(event) {
     profileService.applyConfig();
+    storageService.setCleanAndScanAddresses(function() {});
+  });
+
+  $rootScope.$on('Local/CleanAndScanAddresses', function(event) {
+    storageService.getCleanAndScanAddresses(function(err, val) {
```
matiu: I think this should not be an other event, but it should be in 'newfocusedwallet'

src/js/controllers/index.js
cmgustavo: That is exactly what does. If wallet is focused, display an error. If don't, not. For example, if we are trying to scan a wallet (not focused) that doesn't exist on the new BWS, it displays the message to recreate in the *focused* wallet.

src/js/controllers/index.js
cmgustavo: Good idea!

cmgustavo:
Updated.

src/js/controllers/index.js
```diff
@@ -769,6 +768,19 @@ angular.module('copayApp.controllers').controller('indexController', function($r
     self.setFocusedWallet();
     self.updateTxHistory();
     go.walletHome();
+    storageService.getCleanAndScanAddresses(function(err, val) {
+      if (val) {
+        $log.debug('Clear last address cache and Scan');
+        var wallets = profileService.walletClients;
+        for (var walletId in wallets) {
```
matiu: lodash.each(lodash.keys(profileService.walletClients),  function(walletId) {
   storageSe...
});

PR #2882, opened 2015-06-11
some German corrections

saschad:


saschad:
not need to "create 1-of-1 wallet" in personal wallet section. i used the title string "create new wallet", so this will work with all translations,

cmgustavo:
Awesome regarding 'Create new wallet'. Good catch. Thanks!

LGTM

PR #2891, opened 2015-06-12
Resume/OnLine events do not display notifications

cmgustavo:
* Quiet mode when Resume or OnLine events are triggered (for mobile).


matiu:
LGTM

isocolsky:
Working ok on Android!

matiu:
ACK, merging.

still needs a refactor:

1) we should not add the `quiet` parameter, better not to call `setOngoingProcess`
2) add `opts = opts || {}`

PR #2895, opened 2015-06-17
When running from browser for first time configCache.wallet is not found

greggzigler:
The first time I ran Copay as a server and opened localhost:3000 in Chrome on Mac OSX, the page did not render. The configService had an error, configCache.wallet is undefined, when looking for settings.unitCode. This change solves that error for me.


matiu:
thanks, good catch

PR #2901, opened 2015-06-19
Feature: send all and fee customization.

matiu:
- Allows send all available funds, with a single click
- Allows to configure the fee for outgoing transactions (Priority / Normal / Economy), values suggested by @dabura667 

(needs new BWS api)

### DIFF
--- a/src/js/controllers/preferencesFee.js
+++ b/src/js/controllers/preferencesFee.js
@@ -1,0 +1,35 @@
+'use strict';
+
+angular.module('copayApp.controllers').controller('preferencesFeeController',
+  function($rootScope, $scope, configService, go) {
+    var config = configService.getSync();
+    this.feeName = config.wallet.settings.feeName || 'Priority';
+    this.feeOpts = [{
+      name: 'Priority',
+      value: 100,
+    }, {
+      name: 'Normal',
+      value: 50,
+    }, {
+      name: 'Economy',
+      value: 10,
+    }];
+
+    this.save = function(newFee) {
+      var opts = {
+        wallet: {
+          settings: {
+            feeName: newFee.name,
+            feeValue: newFee.value * 100,
+          }
+        }
+      };
+      this.feeName = newFee.name;
+
+      configService.set(opts, function(err) {
+        if (err) console.log(err);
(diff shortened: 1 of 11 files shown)


cmgustavo:
LGTM

src/js/controllers/preferencesFee.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+angular.module('copayApp.controllers').controller('preferencesFeeController',
+  function($rootScope, $scope, configService, go) {
+    var config = configService.getSync();
+    this.feeName = config.wallet.settings.feeName || 'Priority';
+    this.feeOpts = [{
+      name: 'Priority',
+      value: 100,
```
dabura667: https://bitcoinfees.github.io/misc/profile/

Looking at the current "stress test" waiting queue, you would need at least 15000 satoshi per kB to get past the spam...

Maybe this is fine for everyday use... but I wonder if there is a simple way to make this value adjustable upward?

I am afraid if someday there is a longer spam attack and fees go up to like 40000 satoshis, this hard coded limit will slow down user's transactions. (ie. I can't use breadwallet right now, the 15 bit fees will slow me down to a grinding halt)

Also see peter's post https://gist.github.com/petertodd/8e87c782bdf342ef18fb#what-needs-to-be-done

Dynamic fee detecting procedures are way too advanced, but I think allowing the priority to be set to anything 100 or above (up to a limit)

What are your thoughts on possible future fee increases due to spam attacks like what's going on right now?

dabura667:
*Other than the mixed feelings brought on by recent events (fees pushed up by spam):*

LGTM

matiu:
Thanks for the info.

I think we can work of a dynamic fee approach if this problem becomes
usual. It should be that difficult if we trust BWS for the 'current fee
status'.



On Mon, Jun 22, 2015 at 2:14 PM, dabura667 <notifications@github.com> wrote:

> *Other than the mixed feelings brought on by recent events (fees pushed up
> by spam):*
>
> LGTM
>
> —
> <https://github.com/bitpay/copay/pull/2901#issuecomment-114185499>.
>



-- 
Matías Alejo Garcia
@ematiu

PR #2902, opened 2015-06-20
Send wallet

matiu:
Transfer fund between wallets using a popup in `send`

### DIFF
--- a/public/views/walletHome.html
+++ b/public/views/walletHome.html
@@ -266,7 +266,7 @@
         </div>
         <div class="text-center" ng-show="!home.generatingAddress && home.addr[index.walletId]">
           <a class="size-12" ng-click="home.openCustomizedAmountModal(home.addr[index.walletId])" translate>
-            Request a specific amount
+            Receive a specific amount
           </a>
         </div>
       </div>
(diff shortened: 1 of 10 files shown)


matiu:
updated and ready for review!

isocolsky:
Worked flawlessly in Android livenet. Great job!
A couple of notes:
- I think the link "send to my wallets" needs a better UI design. It doesn't look like something that can be tapped.
- After a wallet is selected from the list, the flow continues as if I had scanned an address. Do you think it would be useful to show some kind of note stating that the address belongs to the wallet I selected?

public/views/walletHome.html
```diff
@@ -327,6 +327,11 @@ <h4 class="title m0">
             <div ng-hide="home._paypro || home.hideAddress">
               <div class="row collapse">
 
+              <div class="right size-12 m0 text-gray" ng-click="openWalletsModal(index.otherWallets)" ng-if="index.otherWallets">
```
cmgustavo: You could add `pointer` class to do more user friendly for desktop versions

cmgustavo:
LGTM. manual test in iOS.

matiu:
updated.

PR #2909, opened 2015-06-24
Generate QR code with customized amount

cmgustavo:


public/views/modals/customized-amount.html
```diff
@@ -0,0 +1,80 @@
...
+            <i class="icon-checkmark-circle size-14"></i>
+          </small>
+        </div>
+        <div ng-if="!showAlternative">
+          <label for="amount">
+            <span translate>Amount</span>
+          </label>
+          <div class="input">
+            <input type="number" id="amount" name="amount" ng-attr-placeholder="{{'Amount'|translate}}"
+            ng-minlength="0.00000001" ng-maxlength="10000000000" ng-model="_customAmount" valid-amount required autocomplete="off">
```
isocolsky: I think `ng-minlength` & `ng-maxlength` are used to check string length. Are you looking for `ng-min` & `ng-max`? Also, the max value is 100 BTC when unit is set to bits. Is this a reasonable limit?

cmgustavo:
Updated.

isocolsky:
Super useful feature. Great job!

PR #2915, opened 2015-06-26
Disabling unnecessary online/offline events for mobile/desktop

cmgustavo:


src/js/controllers/walletHome.js
```diff
@@ -46,18 +46,6 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
     self.resetForm();
   });
 
-  var disableOnlineListener = $rootScope.$on('Local/Online', function() {
-    // This is needed then the apps go to sleep
-    self.bindTouchDown();
-  });
-
-  var disableResumeListener = $rootScope.$on('Local/Resume', function() {
```
matiu: I think this needs to stay. If you put the up to sleep, and the resume, the 'fast' buttons (menu button, "tap event on press") need to be rebinded.

src/js/controllers/walletHome.js
```diff
@@ -76,8 +64,6 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
     disablePaymentUriListener();
     disableTabListener();
     disableFocusListener();
-    disableResumeListener();
```
matiu: see abobe, probably need to stay.

src/js/controllers/walletHome.js
cmgustavo: "Tap event..." don't use events. This is called from view to the function: `index.openWallet(); index.updateTxHistory()`

PR #2921, opened 2015-06-28
Fix Japanese and English

dabura667:
Especially paymentUri.html was including the span tag and the empty spaces in front of the file, as well as \r which should not be translated around [testnet] in it. This must be fixed.

Also, I worried whether to include [testnet] as a translation or not. I leave it up to you.

Also I added Japanese translations.


matiu:
Good catches! I would say that Testnet should not be translated,but I dont have a strong opinion on that.

PR #2948, opened 2015-07-01
handle multiple outputs in transaction proposal

greggzigler:


matiu:
Looking good!

greggzigler:
There are two ways to test this code:

(1) insert the following code into src/js/controllers/index.js at line 405, at the top fo setPendingTxps:

    txps.push({
      message: 'test multi-output',
      fee: 1000,
      createdOn: new Date() / 1000,
      outputs: [
        { toAddress: '2N8bhEwbKtMvR2jqMRcTCQqzHP6zXGToXcK', amount: 700, message: 'one'},
        { toAddress: '2MuQrmRXQw3Njv4Wgw47evczYKFBHhAfKkB', amount: 700, message: 'two'},
        { toAddress: '2N6S3UXnyeGR9FxdhtfufWL9XyiWZpKUH9v', amount: 700, message: 'three'}
      ]
    });

(2) create a node project, make bitcore-wallet-client as a dependency, export read-only backup from copay wallet, import backup into node, create a txproposal with multiple outputs, send to the wallet.

matiu:
Is this still WIP?

On Tue, Jul 21, 2015 at 5:53 PM, Gregg Zigler <notifications@github.com>
wrote:

> There are two ways to test this code:
>
> (1) insert the following code into src/js/controllers/index.js at line
> 405, at the top fo setPendingTxps:
>
> txps.push({
>   message: 'test multi-output',
>   fee: 1000,
>   createdOn: new Date() / 1000,
>   outputs: [
>     { toAddress: '2N8bhEwbKtMvR2jqMRcTCQqzHP6zXGToXcK', amount: 700, message: 'one'},
>     { toAddress: '2MuQrmRXQw3Njv4Wgw47evczYKFBHhAfKkB', amount: 700, message: 'two'},
>     { toAddress: '2N6S3UXnyeGR9FxdhtfufWL9XyiWZpKUH9v', amount: 700, message: 'three'}
>   ]
> });
>
> (2) create a node project, make bitcore-wallet-client as a dependency,
> export read-only backup from copay wallet, import backup into node, create
> a txproposal with multiple outputs, send to the wallet.
>
> —
> <https://github.com/bitpay/copay/pull/2948#issuecomment-123474058>.
>



-- 
Matías Alejo Garcia
@ematiu

matiu:
reviewing.

matiu:
I am using the suggested option (1) to test it.

Clicking the black triangle would open only one extra output, i think it should open all.

Also, all outputs should be grouped somehow on the UX, @bechi can help there. The triangle is not something we are using at copay, maybe it should be rounded (Show all) button.

Leaving some comments on the code too.

src/js/controllers/index.js
```diff
@@ -406,14 +406,33 @@ angular.module('copayApp.controllers').controller('indexController', function($r
...
-      tx.alternativeAmountStr = tx.alternativeAmount + " " + config.alternativeIsoCode;
-      tx.alternativeIsoCode = config.alternativeIsoCode;
+      function formatAmount(obj, amount) {
+        obj.amountStr = profileService.formatAmount(obj.amount) + ' ' + config.unitName;
+        obj.alternativeAmount = rateService.toFiat(obj.amount, config.alternativeIsoCode) ? rateService.toFiat(obj.amount, config.alternativeIsoCode).toFixed(2) : 'N/A';
+        obj.alternativeAmountStr = obj.alternativeAmount + " " + config.alternativeIsoCode;
+      };
 
+      if (tx.outputs) {
+        tx.amount = lodash.reduce(tx.outputs, function(total, o) {
```
matiu: It could be simpler to have: 
`tx.summary`
and
`tx.formatedOutputs`
separated.

`summary` could have the accumulated amount, and the recipient count.
`formatedOutputs` could by a list with already formatted amounts and note. 

Then, the template would show the multioutput variant if `summary` exists, and if expanded, formatedOutput will be shown (not need to call expand, but only set a template variable).

greggzigler:
Thanks for the feedback. Regarding the black triangle and expanding one at a time:  the primary use case that I am designing for is a proposal with more than 100 outputs. The application that creates the proposal will build the list of outputs ordered by descending amount, so the largest outputs will appear first. If we expand the list of 100+ outputs with a single click, then the copay user will need to scroll down very far to reach the Accept button. I am open to alternative designs, as long as the experience is very good for my use case.

greggzigler:
I replaced the expand-one-output-at-a-time UX with a toggle show-all / hide-all experience.

A completely different solution might be to follow the UpdateTxHistory pattern, showing a "MORE" button after 5 outputs are displayed. 

I would welcome suggestions on icons and toggle-vs-more patterns.

src/js/controllers/index.js
```diff
@@ -406,14 +406,32 @@ angular.module('copayApp.controllers').controller('indexController', function($r
...
-      tx.alternativeAmountStr = tx.alternativeAmount + " " + config.alternativeIsoCode;
-      tx.alternativeIsoCode = config.alternativeIsoCode;
+      function formatAmount(obj, amount) {
+        obj.amountStr = profileService.formatAmount(obj.amount) + ' ' + config.unitName;
+        obj.alternativeAmount = rateService.toFiat(obj.amount, config.alternativeIsoCode) ? rateService.toFiat(obj.amount, config.alternativeIsoCode).toFixed(2) : 'N/A';
+        obj.alternativeAmountStr = obj.alternativeAmount + " " + config.alternativeIsoCode;
+      };
 
+      if (tx.outputs) {
+        tx.outputs.details = JSON.parse(JSON.stringify(tx.outputs));
```
matiu: just a details, I think you can use `lodash.clone` here.

matiu:
Thanks for the update. 

ACK with the code. I think that @bechi you take it from here and review a little the UX. 

Personally, I like the current toggle approach. I would just modified the view to group the outputs.

matiu:
Please note that TX history still does not handle multiple outputs, so this cannot be merged yet.

bechi:
ACK

matiu:
Great work @greggzigler @bechi.

PR #2954, opened 2015-07-03
History: export .csv file (compatibility with LibraTax)

### LINKED ISSUE #2853: add CSV download again
libratax___copay__integration__-_ematiu_gmail_com_-_gmail

cmgustavo:
Fixes #2853

### DIFF
--- a/src/js/controllers/index.js
+++ b/src/js/controllers/index.js
@@ -547,7 +547,7 @@ angular.module('copayApp.controllers').controller('indexController', function($r
         return '';
       }
 
-      return dateObj.toJSON().substring(0, 10);
+      return dateObj.toJSON();
     }
 
     function formatString(str) {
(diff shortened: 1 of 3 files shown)


dan-da:
Enterprise wallet user here, presently evaluating Copay.     It would be very useful for accounting purposes if the CSV download includes a column for the USD (local currency) equivalent amount at the time of the transaction.   Please consider adding this.

dabura667:
Yeah, no wallet stores historical USD prices... That always shocked me.

Having one column with value at time of send and one column with value at time of CSV output would make everyone's life soooooo much easier.

matiu:
We used to have exactly that on copay 0.8! We had to create a new xrate api for that. We may include that again ;-)

cmgustavo:
Updated

matiu:
Regarding fiat values, it is in a new ticket: https://github.com/bitpay/copay/issues/2961

src/js/controllers/index.js
```diff
@@ -535,6 +535,97 @@ angular.module('copayApp.controllers').controller('indexController', function($r
...
+  this.csvHistory = function() {
+
+    function saveFile(name,data) {
+      var chooser = document.querySelector(name);
+      chooser.addEventListener("change", function(evt) {
+        console.log(this.value); // get your file name
+        var fs = require('fs'); // save it now
+        fs.writeFile(this.value, data, function(err) {
+          if(err) {
+            alert("error"+err);
```
isocolsky: I get that this is supposed to be a rare event but maybe we can have better error communication with the user?

src/js/controllers/index.js
```diff
@@ -535,6 +535,97 @@ angular.module('copayApp.controllers').controller('indexController', function($r
...
+          self.satToUnit = 1 / self.unitToSatoshi;
+          var data = txs;
+          var satToBtc = 1 / 100000000;
+          var filename = 'Copay-' + (self.alias || self.walletName ) + '.csv';
+          var csvContent = "data:text/csv;charset=utf-8,";
+          csvContent += "Date,Destination,Note,Amount,Currency,Spot Value,Total Value,Tax Type,Category\n";
+
+          data.forEach(function(it, index) {
+            var _amount = (it.action == 'sent' ? '-' : '') + (it.amount * satToBtc).toFixed(8);
+            var dataString = formatDate(it.time * 1000) + ',' + formatString(it.addressTo) + ',' + formatString(it.message) + ',' + _amount + ',BTC,,,,';
```
isocolsky: If strings such as `message` contain commas, the whole file would be malformed.

src/js/controllers/index.js
cmgustavo: Oh, you are right. I forgot to fix that. Thanks!

src/js/controllers/index.js
cmgustavo: I thought the same situation, so I did some transactions with large comments with commas and another symbols, and it still works.

isocolsky: Oh I see the `formatString` function takes care of this. My bad.

isocolsky:
LGTM

PR #2968, opened 2015-07-10
Addon support

troggy:
Addons are simple Angular modules with views, controllers, services etc. Addons can register
themselves in Copay using pluginManagerProvider. It allows them to add extra items to the bottom
menu as well as extra tab-views:

````
  addonManagerProvider.registerAddon({
    menuItem: {
      'title': 'Assets',
      'icon': 'icon-pricetag',
      'link': 'assets'
    },
    view: {
      id: 'assets',
      'class': 'assets',
      template: 'colored-coins/views/assets.html'
    }
  });
````

Addons can consume core Copay services and listen for events to react on changes. For this very
first addon system inplementation Copay emits additional BalanceUpdated event so that interested
addons can react on new transactions (see plugin reference implementation below).

Bottom menu can accomodate only 6 items without sacrificing usability, so it was reworked to
have second layer of items. Now If menu has more than 6 items, toggle button will be added to
the menu allowing to reveal extra items in a sliding panel. Bottom menu in this case will show
only 5 items, the rest will be rendered on sliding panel.

This changes addresses issue #2949 and reference implementation of addon could be found here:
https://github.com/troggy/copay-colored-coins-plugin

### CODE REVIEW

**Comment by matiu** [comment]
Thanks a lot for the PR, we will review it shortly. In general, I like the idea and who it is implemented. How do you imagine that addons can be distributed?

**Comment by cmgustavo** [comment]
Hello @troggy , I am checking your PR with the colored-coin-plugin and I got this error in javascript console log: 
´´´
Cannot read property 'utxos' of undefined
```

I followed all steps in (https://github.com/troggy/copay-colored-coins-plugin). Could you say me what I 'm doing wrong?

Marksmargon:
Hello @matiu,
I was thinking about adding an add-on page to your project site and people can submit add-ons for your approval and only verified add-ons will be displayed there.
@troggy - will be happy to hear your thoughts on the matter

troggy:
Hi @cmgustavo . It looks like you haven't started or not properly configured API proxy (step 4). Please check you config. In my config I've added the following:

````
coloredCoins: {
  apiHost: 'localhost:8000'
},
````

then you have to run apiProxy.js on the server (it should work along with Copay server):

````
node bower_components/copay-colored-coins-plugin/server/apiProxy.js
````

**Comment by troggy** [comment]
I've just amended setup steps a bit, so it should be more straightforward now.

**Comment by troggy** [comment]
As for addon distribution, my idea is that people interested in running their own custom Copay instance may use any addons on their own risk. It would be good to have the list of addons in a single place though, albeit they don't need to be necessarily verified - just make clear that they should be used with caution.

File: src/js/controllers/index.js [hunk 1]
```diff
@@ -274,6 +278,7 @@ angular.module('copayApp.controllers').controller('indexController', function($r
         self.walletScanStatus = walletStatus.wallet.scanStatus;
         self.copayers = walletStatus.wallet.copayers;
         self.preferences = walletStatus.preferences;
+        $rootScope.$emit('Local/BalanceUpdated', walletStatus.balance);
```
cmgustavo: Why do you need this $emit here? It 's no longer used in this file.

troggy: As I outlined in PR description, colored addon subscribes to this event: https://github.com/troggy/copay-colored-coins-plugin/blob/master/js/controllers/assets.js#L9

cmgustavo: OK. But to do a code more clean I ‘d like to add a comment before like this:

```
// Used for external plugin or something
$rootScope.$emit('Local/BalanceUpdated', walletStatus.balance);

// Status Shortcuts
... 
```

That's all. Everything LGTM.

troggy: Fixed. Thanks for review!

matiu:
ACK.

it is yet to be defined how addons will be distributed, but with this PR it is possible for 3rd parties to create addons, share the source, and user can pull the sources into copay, and enable them manually.

cmgustavo:
ACK

PR #2994, opened 2015-07-15
handles invalid txs in history

matiu:
related to https://github.com/bitpay/bitcore-wallet-service/pull/282

handles better `invalid` txs in history.


cmgustavo:
LGTM

Issue #2971, opened 2015-07-10
Spec: Shared Wallet User Management


bitjson:
It would be valuable to make modifying the signature requirements in a shared wallet more intuitive.

If a quorum of users in a shared wallet want to add or remove one or more users from the wallet, it should be possible to sweep the "old" wallet to a "new" wallet in a way similar to signing a transaction. When this is done, it should be possible to maintain the wallet's transaction history.

bitjson:
[Requested on reddit](https://www.reddit.com/r/Bitcoin/comments/3cszc4/new_features_for_copay_users_in_version_11/csyt2i4)

dan-da:
This seems like progress, thx!   So here I'll provide some food for thought.  Sorry if its a bit rambly, but it was basically my thought process as I was first evaluating CoPay and hopefully you will get the gist.

Here's my present day situation that I'm wanting to replicate in a bitcoin environment.  So far, CoPay seems closest of all the wallets out there.

My secretary prepares bill payments and payroll payments using our bank's website.  She is able to prepare payments only.  When I login, I am notified if there are any pending payments awaiting authorization.  I can then review them and either authorize them invidually or en masse.

Important:  as the authorizer, I can also unilaterally send a payment on my own.

If the secretary ever leaves the company, the replacement will receive a new login account on the bank's website and will be given the same role/permissions.  I will remove permissions from the first account.  Everything then continues as normal.  easy peasy.

So this is basically a 1 preparer : 1 approver model.    There could theoretically also be n preparer : 1 approver or y preparer : m of n approver.

Now I do not know if an arbitrary number of preparer/signer is possible with bitcoin's multisig.  If one could create a transaction that says:   
    2 of 5 signers are needed and one of them must be X
then that would be getting pretty close.

Without the "one of them must be X" capability, that leaves us with something like:
   2 of 5 signers.   where preparer hold one key and authorizer holds 2 keys plus 2 backup keys.
( authorizer holds 2 keys to enable unilateral spending. )

So that all works fine until the preparer leaves employment or changes roles.

At that point, what do I do?
  The old key should be revoked, but bitcoin doesn't support that.  Can we at least remove or disable/blacklist that person from our CoPay wallet for future transactions?
   The replacement preparer needs a key, but where from?  Should I give them one of my 2 backup keys?  doesn't seem like a good idea.

[...]

PR #2998, opened 2015-07-16
export backup without signing capabilities

matiu:
- this add 'no private key' advanced option to backup.

It is then possible to export a backup without the wallet's private key.

A wallet with a imported backup without private key can: see wallet balance, see tx history, create spend proposal, reject spend proposals, BUT CANNOT SIGN (approve) proposals.

### DIFF
--- a/public/views/backup.html
+++ b/public/views/backup.html
@@ -8,7 +8,12 @@
 
 <div class="content p20v" ng-controller="backupController as backup">
   <div class="row m20t">
-    <div class="columns" ng-show="!backup.backupWalletPlainText && !backup.error">
+    <div class="columns" ng-show="!backup.backupWalletPlainText">
+      <div class="text-warning size-14 m10v" ng-show="backup.error">
+        <i class="fi-alert size-12"></i>
+        <span translate> Failed to create backup </span>
+      </div>
+
       <div class="text-warning size-14 m10v" ng-show="backup.isEncrypted">
         <i class="fi-alert size-12"></i>
         <span translate> The private key for this wallet is encrypted. Exporting a backup will keep the private key encrypted in the backup archive.</span>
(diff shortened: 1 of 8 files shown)


cmgustavo:
Everything is working awesome! 

Just one comment. When creating a proposal from a `no-private-key-backup` wallet and the other copayer with full capabilities sign it (approve), I see the message (on full wallet) "PAYMENT PROPOSAL CREATED" instead "PAYMENT ACCEPTED". This could be confusing.

matiu:
thanks, will check!

On Thu, Jul 16, 2015 at 10:34 AM, Gustavo Maximiliano Cortez <
notifications@github.com> wrote:

> Everything is working awesome!
>
> Just one comment. When creating a proposal from a no-private-key-backup
> wallet and the other copayer with full capabilities sign it (approve), I
> see the message (on full wallet) "PAYMENT PROPOSAL CREATED" instead
> "PAYMENT ACCEPTED". This could be confusing.
>
> —
> <https://github.com/bitpay/copay/pull/2998#issuecomment-121957146>.
>



-- 
Matías Alejo Garcia
@ematiu

matiu:
Updated.

cmgustavo:
LGTM

dan-da:
This is exciting guys.  If I'm not mistaken, it pretty much enables what I was asking for in issue #2971.  I will try it out at soonest opportunity.

Question: is it planned (or already possible) to add a non-signing user via the UI?   Exporting a backup, sending via external channel, and then importing on the other side is a pretty klunky workflow.

matiu:
@dan-da we will probably add QR show / QR scan soon to simplify the flow.

dan-da:
Is there already an issue created for the QR show/scan feature?

dan-da:
ok, I just tried this out with a 1 of 1 (personal) wallet.

The good:
  + backup and import worked flawlessly.
  + non signing wallet was able to create a spend proposal and signing wallet was able to approve or reject it.

The not so good:
 + backup/import is a non-intuitive way to add a non-signer ( aka preparer ) user.
 + Operations on either side do not cause a notification to appear in the other wallet.  eg after creating spend proposal, the other wallet must exit wallet then re-enter before it appears.   likewise vice-versa after accepting or rejecting.
 + When clicking on Payment Proposal, the created by field says "me".   This makes some technical sense, however it is not the desired workflow where there would be named user(s) in a preparer role and other named user(s) in a signer role.  anyway, it is odd to see it was "me" that created the tx when really it wasn't.

This is a great first step, but I think a bit more work is needed to make the non-signer role into a first class wallet participant.

I can turn these into separate issues if that would be helpful.

matiu:
Thanks a lot for the detailed review.  I agree on your "bad" points. We are currently working on 2 of those (1 and 3). And a couple more that will also help this feature: more detailed access control, qr for exporting/exporting, etc. 

Yes. It will help if you can spawn new tickets or if you can create a checkbox list for the bads. 

Thanks.

PR #3000, opened 2015-07-16
Addons can use OngoingProcess indicator

troggy:
Make Copay to call ``setOngoingProcess`` when ``Addon/OngoingProcess`` event received.
Addons may emit this event to display relevant process indication to user.
Example: https://github.com/troggy/copay-colored-coins-plugin/blob/41df7c7d90e203e39245e5a574204fd53a884112/js/controllers/assets.js#L11


cmgustavo:
ACK. Thanks!

PR #3011, opened 2015-07-17
Ledger hardware wallet support

EricLarch:
Add support for Ledger hardware wallet.

Requires to merge https://github.com/bitpay/bitcore-wallet-client/pull/92


src/js/controllers/walletHome.js
```diff
@@ -724,6 +751,9 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
       address = form.address.$modelValue;
       amount = parseInt((form.amount.$modelValue * unitToSat).toFixed(0));
 
+      console.log(fc);
```
matiu: please remote log or use $log.debug

src/js/controllers/create.js
```diff
@@ -35,6 +35,8 @@ angular.module('copayApp.controllers').controller('createController',
       $scope.requiredCopayers = Math.min(parseInt(n / 2 + 1), maxReq);
     };
 
+    this.externatIndexValues = lodash.range(0,20);
```
matiu: variable should be named 'externalIndexValues' ?

public/views/create.html
```diff
@@ -87,12 +100,22 @@
...
-        <div ng-hide="hideAdv" class="row">
+        <div ng-hide="hideAdv" class="row" ng-init="hideSlots=true">
           <div class="large-12 columns">
             <label for="network-name" class="line-b oh">
               <span translate>Testnet</span>
               <switch id="network-name" name="isTestnet" ng-model="isTestnet" class="green right m5t m10b"></switch>
             </label>
+            <label ng-show="create.isChromeApp()" for="hw-ledger" class="line-b oh">
+              <span translate>Use Ledger hardware wallet</span>
+              <switch id="hw-ledger" name="hwLedger" ng-model="hwLedger" class="green right m5t m10b" ng-change="hideSlots=false"></switch>
```
EricLarch: If I do `ng-change="hideSlots=!hideSlots"` instead of `ng-change="hideSlots=false"` I have a strange JS error at runtime. So right now if you switch off the Ledger HW it won't hide the select. Any idea?

src/js/controllers/walletHome.js
EricLarch: Fixed in c893aac5a4ee4377bd60d99f81f1e8061d7629e3

src/js/controllers/create.js
EricLarch: Fixed in c893aac5a4ee4377bd60d99f81f1e8061d7629e3

src/js/controllers/create.js
```diff
@@ -53,23 +59,50 @@ angular.module('copayApp.controllers').controller('createController',
...
+      }
+    };
+
+    this._create = function (opts) {
       $timeout(function() {
         profileService.createWallet(opts, function(err, secret) {
           self.loading = false;
           if (err) {
+            if (err == "Error creating wallet" && opts.extendedPublicKey) {
+              err = "This xpub index is already used by another wallet. Please select another index."
```
matiu: use gettext(str) for transation

public/views/create.html
matiu: probably you can use the model variable `hwLedger` in the ng-if/ng-hide of the select (no need to use hideSlots)

matiu:
I did a first quick review, looking good great work!

I will test it more and comment later.

src/js/controllers/create.js
EricLarch: Fixed in 837a7ae48be2a20eec4a8ce58cbf22e641deac45

public/views/create.html
EricLarch: Fixed in f6c34981c04323b4dea3e3568ed6813b1c0a6f5d
Thanks

ecv4btc:
When I click contact book or transaction details the screen goes white on all views. In order to see it, I have to resize the screen and it looks normal again.

src/js/controllers/create.js
```diff
@@ -53,23 +59,50 @@ angular.module('copayApp.controllers').controller('createController',
...
+          if (data.success) {
+            opts.extendedPublicKey = data.xpubkey;
+            opts.externalSource = 'ledger';
+            opts.externalIndex = $scope.externalIndex;
+            self._create(opts);
+          } else {
+            self.loading = false;
+            $log.debug(data.message);
+            self.error = data.message;
+            $scope.$apply();
```
matiu: it seemsthat $scope.$apply will be called 2 times if data.sucess is false. This could lead to an angular error.

matiu:
@EricLarch could you reproduce the problem mentioned by @ecv4btc ? Can you check the console log? Any error shown there?

Also I suggest to rebase to current master, just to see if it fixes the problem.

src/js/controllers/create.js
EricLarch: It's called twice indeed, but UI elements have been changed meanwhile. I'm not familiar with Angular, so what is the rule to be able to call again $scope.apply() ?

EricLarch:
I have merge uptream master, it looks like it fixes the refresh issue (I had it also)

src/js/controllers/create.js
matiu: I think you should call it once all changes are done.

EricLarch:
Regarding the white screen, I've spoken to quickly: the issue is still here. If you go on transaction details, the screen stays white until resize. 

I have built the app on master, and I also have the bug. So it doesn't look like it's related to our changes. Can you check from your side on master?

matiu:
I just checked on master / new builded chrome extension thes 2 cases:

1- click on 'send to other wallets' on `send` tab. The list of wallets show up.
2- click on a tx in history. Tx history show up.

I wonder if this is a problem with the build process (check you have done `npm i` ; `bower i`) and also please check the console log is any error appear.

EricLarch:
I don't have any errors on the console. If you don't have the issue when you build then it's a problem from my configuration, no need to go further.

XBTdev:
Glad to see this being worked on. Any idea when this will make it into a production build?

matiu:
Hi @EricLarch,

An update on this: We are given a lot of thought to the recovery process. It would be great if  the wallet could be recovered only with the ledger device (or, in general, with the xPrivKey alone), using a given BWS server. To support this we need to implement a way for the client to sign and submit a new reqPubKey to the server, so he can be granted access. We are working on that now, and we expect to have ir ready by next week. We will submit a PR to your PR to update on this. This will allow users to recover their funds only with the ledger device.

matiu:
closed in favor of https://github.com/bitpay/copay/pull/3143

PR #3012, opened 2015-07-17
Added scripts for Crowdin automation

dabura667:
## [15/7/21 This is ready for checking/merging]

A script such as this would allow you to manage translations on Crowdin.

Crowdin offers free enterprise translation service for open source projects and they support many languages.

There are many nice features of Crowdin which could be beneficial to Copay to help a lot of people translate without having to learn node, git, and all this stuff.

Free account > translate via simple UI = many many people will translate into many languages.

ZIP files gave me a headache and I gave up halfway (not to mention there is no Crowdin yet, so I couldn't really test heh.)

If you get a crowdin set up and figure out the zip file handling I couldn't, then the translation creation would just be like

```bash
> grunt translate
> cd i18n
> node crowdin_update.js
```
 And you would 1. update template.pot for crowdin. 2. generate the translations based on most recent (since last time you updated) and 3. generate the po files for each language (couldn't figure this one out)

I basically copied this script for electrum and rewrote it in nodejs:
https://github.com/spesmilo/electrum/blob/bb4d175978ac39f5e6c89dc827b9d1975d283dae/contrib/make_locale

If you do sign up for crowdin, I would appreciate it if you gave out the api key to some of us other translators. The more frequent the updates to crowdin, the better.

Any questions welcome.


dabura667:
https://crowdin.com/page/open-source-project-setup-request

cmgustavo:
It 's awesome! Thank you! I will check the code asap.

BTW, We already have a crowdin account approved for Copay: https://crowdin.com/project/copay

I will add you to translators team. Let me update the readme with more info about this.

dabura667:
I noticed that the Crowdin is closed on invite basis.

I find that more translations get done if you just open it to anyone with a Crowdin account. Since each language has its own activity history, each language "leader" (or whatever you want to call it) will be able to see if someone is trying to change translations and sock-puppet their votes to the top or something.

Also, I think opening the "Add Term" function in the terminology should be open to anyone as well. I have never seen abuse of it, and having terms set up will allow users to notify later translators of how certain words are translated.

So to recap my suggestion:
1. Open up translating to all
2. Open up "Add Term" to all

dabura667:
This is ready for merging.

To anyone with the crowdin api key, here is the update instructions: (I was wondering whether to write in the readme)

1. copy api key into a file called `crowdin_api_key.txt` just formatted in ANSI or ASCII. (utf8 was giving me a BOM I had to slice)
2. place this txt file into the copay/i18n folder.
3. `grunt translate` (while in copay folder)
4. `cd i18n`
5. `node crowdin_update.js` (any files that say "skipped" in the response meant that the files were identical, so no update was done.) (Also, Export call is also known as "Build translation" on crowdin, this will update the contents of the zip downloaded with the crowdin_download.js)

I had trouble getting https to work so for file update I used bhttp, but if this is not necessary, I am open to pull requests.

cmgustavo:
Great job! 
The code look good to me, but let me try it manually before merge.

cmgustavo:
Everything is OK, but I think we should try to solve another issue before merge this PR.

The issue is related with iso-code we are using in Copay right now and the impossibility to modify the header information inside *.po files. For example, in es.po we have:

```
"Project-Id-Version: Copay\n"
"POT-Creation-Date: \n"
"PO-Revision-Date: \n"
"Last-Translator: \n"
"Language-Team: \n"
"Language: es\n"
...
```

Crowdin generates a .po file with more headers and the most important with `Language` value different. It 's not wrong to use that format, for example es_ES (or xx_XX). The problem is that in Copay we are handling only the first two letters (making a global translation for each supported languages).

For example, if your device is configured to **es_AR**, you will see Copay in english, because we are matching with **es_ES**. For current Copay, the language `es_AR` doesn't exist.

So, we should do some changes in the code to support that denomination. I sent an email to Crowdin team asking about to modify the header of `.po` file, and they answered that it 's not possible for now.

What do you think? Do you have any idea how to solve it?

dabura667:
I'll add some parsing to crowdin_download.js to remove the _XX part, and then we must just make sure not to enable two different locales in the same language on crowdin.

I'll get to it within the next 24 hours.

dabura667:
This should fix.

I fixed on my mobile, so I can't test.

Let me kno if it doesn't work.

dabura667:
I just made it so that:

1. it only modifies files which are shown to have translations in them.
2. it also removes the e-mail address of the crowdin account that crowdin adds.in Last-Translator I figure not many people want to give that out. (Though you can find it on crowdin if you look hard enough.)

dabura667:
Now we just need to make sure not to enable two regions of a single language. (I went through crowdin and made sure of this.)

cmgustavo:
This is awesome!

dabura667:
I forgot to include - and _ in possible email, so I made the regex just be 1 space, 1 less than, at least one of any character, then @ then at least one of any character then . then one of any character then greater than.

This should catch any e-mail address at all.

dabura667:
Ok, I think I'm done here, but I keep wondering how much I should cover in this PR.
Currently merging will merge the most recent translations as well as add the new languages to the app.

If you think I should add something else to this PR or cleanup so that only the scripts are merged, or something like that, please let me know.

cmgustavo:
Change language isn't working, but it 's very simple to fix. Maybe you forgot to do that:

`Gruntfile.js` line 129:

```
files: {
          'src/js/translations.js': ['po/*.po']
        }
```

Add the right path:

```
files: {
          'src/js/translations.js': ['i18n/po/*.po']
        }
```

cmgustavo:
Everything looks great! Please, fix the issue that I mentioned before to then merge it

dabura667:
Fixed the compiler and also an issue where if appstore and update info English files don't exist there is nothing to compare to remove the untranslated files, so it would fail and all locale files (including po) would remain (including untranslated)

If there was a way to download the source files from crowdin, I would download and compare to local copy, just in case, but there is no such api.

Because of this, one thing must be remembered:

1. If you change appstore_en or updateinfo_en, please perform crowdin_update.js before doing another crowdin_download.js, otherwise all of that file will show up as "translated."

cmgustavo:
ACK. Great job!

PR #3016, opened 2015-07-18
Fix Japanese and English

dabura667:


cmgustavo:
Thanks! We have to update .pot file on Crowdin.

PR #3019, opened 2015-07-19
Add Greek appstore description

dabura667:
I got the appstore description and app strings translated by a Greek person I was introduced to.

App strings are on Crowdin, and download is disabled for normal users, otherwise I would add it here.

Here is the appstore description.


cmgustavo:
Thanks!

PR #3027, opened 2015-07-23
Add QR scanner directive

troggy:
QR scanner directive has been extracted from topbar controller, so it can be reused in other places ([example](https://github.com/troggy/copay-colored-coins-plugin/blob/2ae97b4a4f6a1cce02be9db6b130fe9634ea1bff/views/modals/send.html#L15)).

Attributes:
- ``on-scan`` - callback which will receive scan result
- ``before-scan`` - (optional) code to execute before scan.

Example (from topbar.html):

````
<qr-scanner ng-show="index.isComplete" on-scan="topbar.onQrCodeScanned(data)" before-scan="topbar.openSendScreen()" />
````


matiu:
Great job! We can use it also in `join` controller

cmgustavo:
Hi @troggy , I made some manual test and it works well on mobile (cordova), but it doesn't work on desktop (chromeApp and node-webkit). I get `Could not recognize a valid Bitcoin QR Code`.

Maybe it 's a problem with timeout's. I mean, when you scan a qr, it tries to recognize a qr before the scanner finished.

I tested it with many QR codes. Do you have this trouble on desktop?

troggy:
Just tried dozen of times and I've managed to get same error once (cannot reproduce though), so - yes - there is some problem in the code. I've look for it.

troggy:
I've added simple console.log to check which data scanner produces and surprisingly it sometimes reads code with mistakes.

E.g.
Valid value:
bitcoin:2N261uizM7DC8ibYoBhHQNaz6RbX2kEAmVj

Some examples of invalid values returned by scanner (off by one character - in bold):
**f**itcoin:2N261uizM7DC8ibYoBhHQNaz6RbX2kEAmVj
bitcoin:2N261uizM7DC8ibYoBhHQN**i**z6RbX2kEAmVj

I will do the same for original code (before my refactoring)

troggy:
Got similar error on the third try on copay master. 
Could you please check if it's reproduced for you on the codebase before this PR ?

cmgustavo:
I did the same test in `master` and your PR. In master it works always. In your PR the first time I open the desktop app (or chrome app) and scan a QR code I got not valid bitcoin address. But the console log show the right bitcoin address. If I scan the same QR again, it works. But if I scan another new QR, it doesn't. But I am not sure, because I did another test and it works for new QR.

The problem is in desktop app and I am pretty sure that there is an issue with timeout function or something related with the modal. Please, let me do more test and I will try to help you to discover the problem. Anyway, I think this issue is not so critical, but I 'd like to resolve it.

cmgustavo:
It ’s still not working at the first time when I am trying to scan (Desktop and Chrome). I also tested in web and not working the first time. Can you reproduce it in your local environment?

troggy:
@cmgustavo As I've already wrote, I was able to reproduce it and even [backtrace the source](https://github.com/bitpay/copay/pull/3027#issuecomment-125467480).
I don't have enough time to fix qrScanner library, so I've added quick workaround for this problem: now scanner returns result only if two consecutive scans produce the same data. This fix slows scanning process a bit, but fixed the issue for me. Also this fix won't work if the QR image is not moving, because obviously in this case two consecutive scans will also produce same result and both may be incorrect.

Please check if this fix resolves the issue for you.

cmgustavo:
LGTM. Thanks!

cmgustavo:
Manual test in Chrome, OSX and Android: OK

Issue #3022, opened 2015-07-21
We urgently need red!


samapplebee:
Every wallet colour under the sun exists for users to select. Except red.

Copay needs red! Please... :)

kaibakker:
I can give this a try, looks like https://github.com/bitpay/copay/blob/6cb4a4c34bc9d7c39d7f0a8ab5712ee5e6ef4a92/src/js/controllers/preferencesColor.js file is responsible for the possible color preferences. 
What kind of red would you like?

samapplebee:
Thanks Kai!

A nice bold one, maybe like #DD4B39 ? (just colour picked from Gmail
'compose' button haha)

On 23 July 2015 at 11:29, Kai Bakker <notifications@github.com> wrote:

> I can give this a try, looks like
> https://github.com/bitpay/copay/blob/6cb4a4c34bc9d7c39d7f0a8ab5712ee5e6ef4a92/src/js/controllers/preferencesColor.js
> file is responsible for the possible color preferences.
> What kind of red would you like?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitpay/copay/issues/3022#issuecomment-124050962>.
>



-- 
Sam Applebee
Product owner - Satalia <http://www.satalia.com/>

Mobile: 0789 111 0401
LinkedIn <http://uk.linkedin.com/in/samapplebee> - Twitter
<http://twitter.com/Sam_Applebee>

kaibakker:
To make things clear, I am not a bitpay developer, I just passed by and thought it would be a nice way to learn something. Here is the pull request:

https://github.com/bitpay/copay/pull/3028

samapplebee:
Yeah I realise it's because you're awesome and not because you're getting
paid haha

Nice one Kai

On 23 July 2015 at 13:50, Kai Bakker <notifications@github.com> wrote:

> To make things clear, I am not a bitpay developer, I just passed by and
> thought it would be a nice way to learn something. Here is the pull request:
>
> #3028 <https://github.com/bitpay/copay/pull/3028>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitpay/copay/issues/3022#issuecomment-124084647>.
>



-- 
Sam Applebee
Product owner - Satalia <http://www.satalia.com/>

Mobile: 0789 111 0401
LinkedIn <http://uk.linkedin.com/in/samapplebee> - Twitter
<http://twitter.com/Sam_Applebee>

[...]

PR #3028, opened 2015-07-23
Added red as a wallet color option, issue: #3022

kaibakker:
Feel free to merge or not. I am not a really a color person :)


matiu:
Nice! Good choice!

matiu:
Thanks for the contribution!

We may tune it a little bit later, so it does not get confused with red notifications, as @bechi suggested.

PR #3034, opened 2015-07-24
Added untranslated strings

dabura667:
btw is using gettext() with `$scope.ok` and `$scope.cancel` etc. ok to do?

### DIFF
--- a/i18n/po/template.pot
+++ b/i18n/po/template.pot
@@ -160,6 +160,8 @@ msgstr ""
 
 #: public/views/copayers.html
 #: public/views/walletHome.html
+#: src/js/controllers/copayers.js
+#: src/js/controllers/preferencesDelete.js
 msgid "Cancel"
 msgstr ""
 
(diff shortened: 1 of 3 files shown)


cmgustavo:
Need rebase.

I think `navigator.notification.*` doesn't work with gettext.

dabura667:
The problem I am trying to fix with this is that iOS app and Android app have the "Confirm" "Are you sure you want to delete wallet?" in English.

Kind of important that the user understands what this means. :-)

I rebased and restructured to make the localization its own string variable, then just use that.

cmgustavo:
I know what you mean, but it 's still not working with the cordova plugin `navigator.notification`. I tested it on iPhone:

Note that I have it in spanish, but the popup is in english.

screenshot_7_27_15__09_53

I think it happens with all others notification that uses the plugin.

dabura667:
Hmm... well, I can't figure it out... so I will close this PR and make it an issue instead.

cmgustavo:
Hi @dabura667 , 

Maybe you want to open it again, I have the solution:

To translate in-file you should use:
```
gettextCatalog.getString('string');
```
This replace the text before to do `navigation.notification` or call another cordova plugin. It works to me. Please, If you can, just replace `gettext` by `gettextCatalog.getString` and updated the PR. I could update all the other messages used by cordova plugins.

dabura667:
ok, I will do it in a few hours or less.

PR #3035, opened 2015-07-24
Feat/dynamic fees

cmgustavo:
* Adaptive fee policy: `priority`, `normal` and `economy`.
* Global settings.
* First steps for advanced send options (#3045).


matiu:
reviewing

matiu:
Using old BWC, I get:

```
Uncaught TypeError: Cannot read property '0' of undefined(anonymous function) @ copay.js:1631(anonymous function) 
```
which is:
```
          levelsLivenet[i]['feePerKBUnit'] = profileService.formatAmount(levelsLivenet[i].feePerKB) + ' ' + unitName;

```

cmgustavo:
Updated

matiu:
- Fee level: priority (xxx BTC/kB)  =>    'Fee Policy: Priority' 
 - In :
   -   Remove title
   -   Remove fee value in selector
   - Add fee value for selected option after `minutes`: (Current fee rate for this policy: xxx bits/kiB)

Add to:    `Actual fees are determined based on network load and the selected policy`.

cmgustavo:
Updated

matiu:
ACK, tested with and without adaptive fee support in BWS. Great work!

PR #3050, opened 2015-07-29
Translation/updates fee levels

cmgustavo:


cmgustavo:
@dabura667 , I have one more updating for `template.pot`.

src/js/services/feeService.js
```diff
@@ -1,8 +1,15 @@
 'use strict';
 
-angular.module('copayApp.services').factory('feeService', function($log, profileService, configService) {
+angular.module('copayApp.services').factory('feeService', function($log, profileService, configService, gettextCatalog) {
   var root = {};
 
+  // Constant fee options to translate
+  var feeOpts = [
+    gettextCatalog.getString('priority'),
```
dabura667: Shouldn't we capitalize?

```
'Priority'
'Normal'
'Economy'
```

?

cmgustavo:
It 's not necessary because the capitalization is made by the css class: 

```
.text-capitalize {text-transform: capitalize;}
```

So, the translation should be in lower case.

src/js/services/feeService.js
cmgustavo: The source returns theses words in lower case. So, to avoid modify in the html, the best way is to translate as it comes.

dabura667:
Ok, sounds good.

LGTM

PR #3058, opened 2015-08-03
Allow addons to customize pending proposals

troggy:
Addons now may register ``formatPendingTxp`` function which will be called on every pending tx proposal. This allows to add addon-specific information to proposal (e.g. custom amountStr)

Example which adds color information (if any):
https://github.com/troggy/copay-colored-coins-plugin/blob/b56c0f0d2b6f8e8186f712da980b3d33ed91423b/dist/copayColoredCoins.js#L17


matiu:
ACK

PR #3060, opened 2015-08-03
Advanced send options

### LINKED ISSUE #3045: Advanced Send Options
Set these options for current transaction:

* Fee policy
* Spend unconfirmed transactions

This action doesn't store the selected option in **preferences**.

### LINKED ISSUE #3075: ChromeApp can not recreate any wallet
Error with: `window.alert()` that is not allowed in ChromeApp.

```
Client ERROR: Copayer not found
extensions::platformApp:17 window.alert() is not available in packaged apps.
```

cmgustavo:
fixes #3045
fixes #3075

### DIFF
--- a/public/views/walletHome.html
+++ b/public/views/walletHome.html
@@ -418,6 +418,33 @@
               </div>
             </div>
 
+            <div ng-init="home.hideAdvSend=true">
+              <a class="button outline light-gray expand tiny" ng-click="home.hideAdvSend=!home.hideAdvSend">
+                <i class="fi-widget m3r"></i>
+                <span translate ng-hide="!home.hideAdvSend">Show Advanced options</span>
+                <span translate ng-hide="home.hideAdvSend">Hide Advanced options</span>
+                <i ng-if="home.hideAdvSend" class="icon-arrow-down4"></i>
+                <i ng-if="!home.hideAdvSend" class="icon-arrow-up4"></i>
+              </a>
+            </div>
+            <div class="m20b" ng-hide="home.hideAdvSend">
+              <h4 class="title m0" translate>Fee policy for this transaction</h4>
+              <ul class="no-bullet m0 size-14">
+                <li ng-repeat="fee in (index.network == 'livenet' ? index.feeLevels.livenet : index.feeLevels.testnet)"
+                  ng-click="home.currentSendFeeLevel = fee.level" class="line-b p20">
+                  {{index.feeOpts[fee.level]|translate}}
+                  <i class="fi-check size-16 right" 
+                    ng-show="(home.currentSendFeeLevel || index.currentFeeLevel) == fee.level"></i>
+                </li>
+              </ul>
+              <h4 class="title m0" translate>Spend unconfirmed transaction</h4>
+              <div class="p20 line-b">
+                <span class="size-14" translate>Use unconfirmed in this transaction</span>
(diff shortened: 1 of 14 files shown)


matiu:
code LGTM, will try it later today.

public/views/walletHome.html
```diff
@@ -412,6 +412,33 @@ <h4 class="title m0">
...
+                <h4 class="title m0" translate>Fee policy for this transaction</h4>
+                <ul class="no-bullet m0 size-14">
+                  <li ng-repeat="fee in (index.network == 'livenet' ? index.feeLevels.livenet : index.feeLevels.testnet)"
+                    ng-click="home.currentSendFeeLevel = fee.level" class="line-b p20">
+                    {{index.feeOpts[fee.level]|translate}}
+                    <i class="fi-check size-16 right" 
+                      ng-show="(home.currentSendFeeLevel || index.currentFeeLevel) == fee.level"></i>
+                  </li>
+                </ul>
+                <h4 class="title m0" translate>Unconfirmed transactions for this transaction</h4>
```
dabura667: This is confusing, maybe "Unconfirmed outputs for this transaction"

dabura667:
tbh "Spending a transaction" will probably confuse users... (not that "spending an output" is any less confusing...

However, this is "Advanced" so I think we should be specific and say "Ouputs" instead of "Transaction."

you spend outputs, not transactions.

public/views/walletHome.html
isocolsky: I think we should assess how intuitive the concept of "spending unconfirmed transactions" is to the end user. Maybe we should change it to "spend unconfirmed funds"?

dabura667: I agree. "Unconfirmed funds" sounds reasonable.

cmgustavo:
Updated. Thanks for your feedback!

public/views/walletHome.html
```diff
@@ -412,6 +412,33 @@ <h4 class="title m0">
...
+                  <li ng-repeat="fee in (index.network == 'livenet' ? index.feeLevels.livenet : index.feeLevels.testnet)"
+                    ng-click="home.currentSendFeeLevel = fee.level" class="line-b p20">
+                    {{index.feeOpts[fee.level]|translate}}
+                    <i class="fi-check size-16 right" 
+                      ng-show="(home.currentSendFeeLevel || index.currentFeeLevel) == fee.level"></i>
+                  </li>
+                </ul>
+                <h4 class="title m0" translate>Use unconfirmed funds for this transaction</h4>
+                <div class="p20 line-b">
+                  <span class="size-14" translate>Use Unconfirmed Funds</span>
```
isocolsky: One last comment, could you please not capitalize every word. I think "Use unconfirmed funds" is easier to read.

public/views/walletHome.html
```diff
@@ -412,6 +412,33 @@ <h4 class="title m0">
               </div>
             </div>
 
+            <div class="m20b" ng-init="home.hideAdvSend=true">
+              <a class="button outline light-gray expand tiny m0" ng-click="home.hideAdvSend=!home.hideAdvSend">
+                <i class="fi-widget m3r"></i>
+                <span translate ng-hide="!home.hideAdvSend">Show Advanced options</span>
+                <span translate ng-hide="home.hideAdvSend">Hide Advanced options</span>
```
isocolsky: Same here with "advanced" instead of "Advanced"

cmgustavo:
Updated. Thanks @matiu

PR #3064, opened 2015-08-05
txhistory of received payments show non-receiver outputs

greggzigler:
receiver seeing change address output of sender, and outputs to other recipients


cmgustavo:
ACK. Manual test: OK

PR #3069, opened 2015-08-08
Allow addon menu items to run arbitrary code on click

troggy:
This will allow viewless addons or addons having separate view, not only tab view in home contoller.

[Example](https://github.com/troggy/copay-colored-coins-plugin/commit/179cc7b4f0365743ad97a5f6e66c092d91a9e018#diff-ce1d47e10e38c9c3cede9d6c803b88d3R24)

### DIFF
--- a/src/js/controllers/index.js
+++ b/src/js/controllers/index.js
@@ -1,6 +1,6 @@
 'use strict';
 
-angular.module('copayApp.controllers').controller('indexController', function($rootScope, $scope, $log, $filter, $timeout, lodash, go, profileService, configService, isCordova, rateService, storageService, addressService, gettextCatalog, gettext, amMoment, nodeWebkit, addonManager, feeService, isChromeApp, bwsError, utilService) {
+angular.module('copayApp.controllers').controller('indexController', function($rootScope, $scope, $log, $filter, $timeout, lodash, go, profileService, configService, isCordova, rateService, storageService, addressService, gettextCatalog, gettext, amMoment, nodeWebkit, addonManager, feeService, isChromeApp, bwsError, utilService, $state) {
   var self = this;
   self.isCordova = isCordova;
   self.onGoingProcess = {};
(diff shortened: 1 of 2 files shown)


src/js/controllers/index.js
```diff
@@ -140,27 +153,30 @@ angular.module('copayApp.controllers').controller('indexController', function($r
...
-    if (!self.tab)
+    if (!self.tab || !$state.is('walletHome'))
       self.tab = 'walletHome';
 
-    if (document.getElementById(self.tab)) {
-      document.getElementById(self.tab).className = 'tab-out tab-view ' + self.tab;
-      var old = document.getElementById('menu-' + self.tab);
-      if (old) {
-        old.className = '';
+    go.path('walletHome', function() {
```
matiu: Why you removed go.path()? IIUC, go.path('walletHome', is needed for setting the state to `walletHome` when / if set path is called from outside a inner tab.

troggy: You probably wanted to ask why I've **added** go.path() ? Exactly for that purpose -- to set state to ``walletHome`` when ``setTab`` is called from other state.

troggy:
Rebased

troggy:
Rebased

cmgustavo:
LGTM

cmgustavo:
Need rebase

troggy:
Rebased

cmgustavo:
ACK

cmgustavo:
I found an issue in this PR using `bitcoin:` URI. 

* When Copay is opened by a bitcoin URI, and after choosing the folder, it finally goes to Home (instead Send).
* Also if you are testing in a browser, refresh the page in, for example, `Preferences`. After page was loaded, it returns to Home.

@troggy: let me know if you have the same trouble.

troggy:
@cmgustavo Yes, I have the same. The reason for this is that go.walletHome (which calls setTab) is being called even for Preferences - why it should?

troggy:
Without my code it just tries to setTab several times and eventually fails. For user everything is fine, but this kind of logic is very confusing for developer

troggy:
My reasoning is following: ``setTab`` is a function to set tab on the wallet home - it is the only view that has tabs. So in theory it should be perfectly safe to switch to Home in setTab - we shouldn't be calling setTab from other views anyway. But in reality it's flawed - ``go.walletHome()`` is being called for ANY view in ``Local/NewFocusedWallet`` event handler.

troggy:
@cmgustavo Please take a look at https://github.com/bitpay/copay/pull/3153 if it resolves the issue for you?

PR #3070, opened 2015-08-08
fix connection error messages to be grammatically correct

jlopp:
This was annoying me when I was using copay.

On a related note, I received these messages a lot and my connection was fine - I suspect that the problem may have actually been on the Bitcore Wallet Service's end.

### DIFF
--- a/i18n/po/de.po
+++ b/i18n/po/de.po
@@ -182,7 +182,7 @@ msgid "Changing wallet alias only affects the local wallet name."
 msgstr "Änderung der Aliases hat nur Auswirkungen auf den lokalen Namen des Wallets"
 
 #: src/js/controllers/walletHome.js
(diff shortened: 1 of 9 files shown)


cmgustavo:
Hello @jlopp, thanks a lot for your contribution! 

Respect your code, you shouldn't modify the .PO files. The only files you have to commit are `template.pot` and `walletHome.js`. After that, we have to upload the pot file to Crowdin and each translator will have to update their language.

jlopp:
Alright, should be only those relevant files now. FYI, it looks like the connection errors I was receiving were a result of using Copay through Tor - you might want to add this as a goal to provide better Tor support.

dabura667:
LGTM

matiu:
thanks for the contribution!

PR #3078, opened 2015-08-11
Feat/send max

matiu:
- fixes send-max with new fee policies
- shows unconfirmed balance, in case the setting is active.


cmgustavo:
ACK

PR #3081, opened 2015-08-12
Update appstore text for next release

dabura667:
If there are any other bullet point you would like, please let me know.

Once it is merged I will push to crowdin.

### DIFF
--- a/i18n/docs/updateinfo_en.txt
+++ b/i18n/docs/updateinfo_en.txt
@@ -1,5 +1,4 @@
 * Improved fee handling and fee notifications
 * Added advanced sending options
-* Create transaction proposals with multiple destinations
-* Added Italian and Greek language options
+* Added Italian, Russian, and Greek language options
 * Minor bug fixes
--- a/i18n/docs/updateinfo_en.txt
+++ b/i18n/docs/updateinfo_en.txt
@@ -1,7 +1,5 @@
-* Now is possible to make backups without signing capabilities (more info https://github.com/bitpay/copay/pull/2998)
-* Better refresh and sorting rules in transaction history
-* Better handling of `invalid` transactions in history
-* Better handling of wallets with 1K+ transactions in history
-* Exports .csv files with transaction ID
-* Better UTXOs selection polices
+* Improved fee handling and fee notifications
+* Added advanced sending options
+* Create transaction proposals with multiple destinations
+* Added Italian and Greek language options
 * Minor bug fixes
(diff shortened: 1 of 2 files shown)


dabura667:
Also, while thinking of translations:

Maybe we should place https://github.com/bitpay/bitcore-wallet-service/tree/master/lib/templates/en also on Copay's crowdin? Then we can just manually add the templates if and when they are done.

cmgustavo:
LGTM. But I think you should:

* Remove multiples transaction because it 's not supported by the UI yet.
* Add Russian :)

cmgustavo:
Another thing I see about Crowdin is that We should mark the options to export only translations approved. 

```
Skip untranslated strings

If checked, untranslated strings will be absent from the exported file.
This option is not applied to text documents: *.docx, *.pptx, *.xlsx, etc., since missing texts may cause the resulting files to be unreadable.

Export only approved translations

Includes only texts that are both translated and approved. 
This will require additional efforts from your translators to approve all suggestions.
```

What do you think? It 's possible using the API to download only approved files?

dabura667:
> Skip untranslated strings

With PO files, this will mean that if a language is only 50% translated, then the other 50% will not be included in the file. Having this disabled makes the non-translated strings insert the source language into the target language.

> Export only approved translations

This will also cause files to be incomplete.

-------------------

The only thing I was worried about is the situation of a PO file with missing strings. If template.pot has 100 strings, and xx.po only has 98 strings, will the app still build? Will the language still show up when you select it? What will happen when the user sees those two strings not yet translated.

If gettext will just revert to English if the string is missing, I think that is OK.

As far as the API is concerned, I currently have the script parse files to check if they are 100% untranslated. If they are, I delete the files.

If we set Export only approved translations, then it will only download any file with at least 1 approved translation, and any phrases not approved will be deleted from the file for PO files (no placeholders). (Not sure about txt files)

----------------

I don't feel strongly either way. But I think the way it works right now doesn't have any demerits. Most proofreaders are the translators themselves... so I don't see much difference quality-wise to approved and non-approved.

--------------

I will change the PR to match those changes. Hold on.

dabura667:
PS, once we merge this I am thinking of updating crowdin and then sending out a message to all translators.

A lot of the languages fell to 90-95% because of a lot of grammar fixes, so I hope we can have a few days for translators to get around to it.

cmgustavo:
Ok, I agree with maintain approved and non-approved as same thing and also remove untranslated files.

We have more changes in strings on another PR. I will got to update it and then we can to upload the template to crowdin.

Issue #2812, opened 2015-06-02
Access to shared wallet lost by some copayers if multiple copayers are in same application instance


tab00:
To reproduce the error, either in Android or Chrome extension:
Create a new 2-of-2 wallet with wallet name "Shared 2-of-2" nickname "user 1".
Copy the invitation code.
In the same application, join the shared wallet by pasting the invitation code and with nickname "user 2".
You will notice 2 wallets named "Shared 2-of-2".
Select the first "Shared 2-of-2" as the active wallet.
Have some bitcoin deposited into the wallet.
Send some bitcoin out of the wallet.
A message appears that says "Payment proposal created".
Select the second "Shared 2-of-2" wallet in order to add the other signature.
If you were able to do so, press on "Multisignature wallet" in the center of the Home screen. You'll notice "user 1 (Me)" is bolded. It should instead be "user 2 (Me)". The transaction cannot be signed because you are not "user 2".
If you were not able to select the second "Shared 2-of-2" wallet, it may have disappeared from the list (it happens in the Android app), and therefore cannot be selected as the active wallet. Hence transaction cannot be signed as "user 2".

The funds are thus stuck forever in the shared wallet and can no longer be spent.

The "user 2" wallet should not disappear, and it should be selectable as the active wallet so that the transaction can have all required signatures and then allowed to be broadcast.

Some people are going to test multisignature transactions in copay on a single device, just as I did, and may find their bitcoin stuck as described. This can cause some grief if this test is done using a significant amount of real bitcoins.

Please fix this problem as soon as possible.

matiu:
Thanks for the detailed report. I was able to reproduce it. This is purely an UI issue: we are indexing the available wallets be `walletId`, we should do it by `walletId:copayerId`.

chaosgrid:
It would be really useful to have this fixed. I just wanted to setup a multi sig wallet from within the same app instance and because of this bug I had to revert to using multiple devices / app instances.

matiu:
What would be the use case of having multiple keys of the same wallet on the same device?

chaosgrid:
[...]

PR #3083, opened 2015-08-12
Handling errors / multiple join / differenciated balance

matiu:
- fixes https://github.com/bitpay/copay/issues/2812
- Better error handling for error coming from BWC/BWS (matching new BWS's error codes)
- Handles and upgrades with the new 'newBlock' notification (when using the new `useUnconfirmedSpend` flag)
- Displays balance differentiated for confirmed/unconfirmed  (when using the new `useUnconfirmedSpend` flag)
- add a universal error popup (to replace alert() that does not work on ChromeApp)

### DIFF
--- a/public/views/walletHome.html
+++ b/public/views/walletHome.html
@@ -76,7 +76,7 @@
         <div class="amount" ng-style="{'background-color':index.backgroundColor}">
           <div ng-if="!index.anyOnGoingProcess && !index.notAuthorized">
             <div ng-show="index.updateError" ng-click='index.openWallet(); index.updateTxHistory()'>
-              <span class="size-12 db m10b">{{index.updateError}}</span>
+              <span class="size-12 db m10b">{{index.updateError|translate}}</span>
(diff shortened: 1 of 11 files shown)


public/views/walletHome.html
```diff
@@ -76,7 +76,7 @@
         <div class="amount" ng-style="{'background-color':index.backgroundColor}">
           <div ng-if="!index.anyOnGoingProcess && !index.notAuthorized">
             <div ng-show="index.updateError" ng-click='index.openWallet(); index.updateTxHistory()'>
-              <span class="size-12 db m10b" translate>Could not update Wallet</span>
+              <span class="size-12 db m10b" translate>{{index.updateError}}</span>
```
cmgustavo: Please, remove `translate` attribute here. If you want to `index.updateError` to be translated, add the follow: `{{index.updateError|translate}}`

matiu:
updated.

cmgustavo:
LGTM

PR #3095, opened 2015-08-14
Fixes txp modals for single outputs

### LINKED ISSUE #3084: Payment proposal details disappearing
When the popup showing a new spend proposal is open and the proposal gets accepted/rejected, details disappear:
image

cmgustavo:
fixes #3084

### DIFF
--- a/public/views/includes/output.html
+++ b/public/views/includes/output.html
@@ -1,24 +1,14 @@
-<div class="ng-animate-disabled row collapse">
-  <li ng-if="output.isSummary" class="line-b p10 oh" ng-click="toggleOutputDetails(output)">
-    <span class="text-gray" translate>Recipients</span>:
-    <span class="right enable_text_select">{{output.recipientCount}}
-        <i ng-show="output.showDetails" class="icon-arrow-up3 size-24"></i>
-        <i ng-show="!output.showDetails" class="icon-arrow-down3 size-24"></i>
-    </span>
-  </li>
-  <li ng-if="!output.isSummary" class="line-b p10 oh" ng-click="copyAddress(output.toAddress)">
-    <span class="text-gray" translate>To</span>:
-    <span class="right enable_text_select">{{output.toAddress || output.address}}</span>
-  </li>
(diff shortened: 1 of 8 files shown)


public/views/modals/tx-details.html
```diff
@@ -46,7 +37,27 @@ <h4 class="title m0" translate>Details</h4>
         <span class="enable_text_select"> {{btx.labelTo || btx.addressTo}}</span>
       </span>
     </span>
+    </li> 
+
+    <li ng-show="btx.hasMultiplesOutputs" class="line-b p10 oh"
```
greggzigler: when i click on the arrow, it does not expand the list of recipients. perhaps the list is not populated?

src/js/services/utilService.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
```
greggzigler: very nice. less redundancy, and the overall code is cleaner.  good design.

public/views/includes/output.html
```diff
@@ -1,24 +1,14 @@
...
-    <span class="right">{{output.amountStr}}
-      <span ng-show="output.alternativeAmountStr" class="label gray radius">{{output.alternativeAmountStr}}</span>
-    </span>
-  </li>
-  <li class="line-b p10 oh">
-    <span class="text-gray" translate>Note</span>:
-    <span class="right">{{output.message}}</span>
-  </li>
-</div>
+<li class="p10 oh" ng-click="copyAddress(output.toAddress)">
```
greggzigler: removing the lines between list items is a good UX improvement. much easier for the user to read the list of recipients.

src/js/services/utilService.js
cmgustavo: Yes, that 's the idea, because right now we haven't any common service of utilities. Maybe we need it later.

greggzigler:
all the functionality works for single and multiple outputs, the user experience is better, and the code is cleaner. nice job!

PR #3097, opened 2015-08-15
Changed crowdin download script to check for translation build.

dabura667:
Also includes newest translations that are not in master yet.


dabura667:
I noticed while fixing the download script that because Linux uses LF instead of CRLF, it was viewing the file as a different file when the contents were the same.

So that there is no difference in the files that are uploaded (similar to Github how they normalize to CRLF) I have added normalizing to the upload files.

cmgustavo:
ACK.

Thanks a lot!

Issue #3098, opened 2015-08-17
"Uncaught TypeError: Cannot read property 'utc' of undefined" on nodewebkit build for linux64


mpolci:
I builded the desktop executable of copay release 1.1.2 (on ubuntu 15.04 standard installation). When I Iaunch the linux64 executable I obtain the following error and it opens a blank window. Previusly I builded the release 1.1.0 without problems.

```
[3912:0817/175117:ERROR:nw_shell.cc(335)] TypeError: Cannot read property 'utc' of undefined
    at new <anonymous> (file:///tmp/.org.chromium.Chromium.3TGsLA/public/lib/angular.js:51028:17)
    at invoke (file:///tmp/.org.chromium.Chromium.3TGsLA/public/lib/angular.js:20510:17)
    at Object.instantiate (file:///tmp/.org.chromium.Chromium.3TGsLA/public/lib/angular.js:20518:27)
    at Object.<anonymous> (file:///tmp/.org.chromium.Chromium.3TGsLA/public/lib/angular.js:20379:24)
    at Object.invoke (file:///tmp/.org.chromium.Chromium.3TGsLA/public/lib/angular.js:20510:17)
    at Object.enforcedReturnValue [as $get] (file:///tmp/.org.chromium.Chromium.3TGsLA/public/lib/angular.js:20363:37)
    at Object.invoke (file:///tmp/.org.chromium.Chromium.3TGsLA/public/lib/angular.js:20510:17)
    at file:///tmp/.org.chromium.Chromium.3TGsLA/public/lib/angular.js:20328:37
    at getService (file:///tmp/.org.chromium.Chromium.3TGsLA/public/lib/angular.js:20469:39)
    at Object.invoke (file:///tmp/.org.chromium.Chromium.3TGsLA/public/lib/angular.js:20501:13)
[3912:0817/175117:INFO:CONSOLE(20474)] "Uncaught TypeError: Cannot read property 'utc' of undefined", source: file:///tmp/.org.chromium.Chromium.3TGsLA/public/lib/angular.js (20474)
```

mpolci:
The problem seems caused by angular-moment e moment libraries versions. The command `bower install` installed angular-moment v0.10.2 and moment v2.10.6.
I resolved changing bower.json file from:
```
    "angular-moment": "~0.10.1",
    "moment": "~2.10.3",
```
to:
```
    "angular-moment": "0.10.1",
    "moment": "2.10.3",
```

matiu:
[...]

PR #3099, opened 2015-08-18
Fixed a problem with angular-moment release (issue 3098).

mpolci:
Fixed the "Uncaught TypeError: Cannot read property 'utc' of undefined" exception on nodewebkit build. See issue https://github.com/bitpay/copay/issues/3098 .


matiu:
thanks!

PR #3118, opened 2015-08-24
WIP: Adds confirmation before send/sign a transaction

### LINKED ISSUE #3116: Add confirmation popup after send
and show the actual, total, transaction fee.

@Kirvx we are finally moving forward with this, it will be optional (it can be disabled in settings), and it will show the total transaction fee, before signing.

cmgustavo:
To merge this PR, we need to do some changes on `BWS` in order to support and works as expected. The idea is to delay sending notifications for 1 minute. 

fixes #3116


matiu:
outdated, this will be solved using a different approach in BWS.

PR #3131, opened 2015-09-03
Refactor txStatus to ease overriding in addons.

troggy:
Favour overriding instead of options introduced in #3065.

[Example](https://github.com/Colored-Coins/Colored-coins-add-on-for-Copay/commit/d052572ee72e6cfe211696eb1a68d4bf3bf542d1#diff-ce1d47e10e38c9c3cede9d6c803b88d3R147)


matiu:
thanks!

PR #3132, opened 2015-09-03
Create availableBalance directive for Send screen

troggy:
..so it can be decorated by addon.

[Example showing funds locked by colored utxos](https://github.com/Colored-Coins/Colored-coins-add-on-for-Copay/commit/4c28b113a5b0c1161d96622112d9d02096774f98#diff-ce1d47e10e38c9c3cede9d6c803b88d3R50)


matiu:
The advantage of using a directive instead of using `ng-include` is that it can be decorated by the add-on, is that correct?

troggy:
Correct

cmgustavo:
LGTM

Issue #3140, opened 2015-09-03
Copay logo disappears on the ChromeApp with the new directive


cmgustavo:
@troggy Do you have the same issue with coloredCoins? It works on node-webkit apps.

troggy:
Yes, it's weird. ColoredCoins logo is displayed using CSS background property and it works fine. But, yes, Copay logo disappears

troggy:
Alright, that's known issue: http://stackoverflow.com/questions/22754393/in-a-chrome-app-using-angularjs-can-i-use-the-ngsrc-directive-directly-for-inte

troggy:
Please take a look at possible fix: #3141

PR #3141, opened 2015-09-04
Whitelist chrome-extension URL schema for images

troggy:
Possible fix for #3140

By default Angular will prefix such image URLs with ``unsafe:`` thus stopping images to appear
Related reading: http://stackoverflow.com/questions/15606751/angular-changes-urls-to-unsafe-in-extension-page?lq=1


cmgustavo:
Thanks! Now it works.

PR #3145, opened 2015-09-07
Feat/glidera integration

cmgustavo:


matiu:
- preferences -> Glidera should be hidden if no auth token
- should be enabled for livenet
- clicking on the 'bank' icon, on a testnet wallet I get:
request_for_permission

matiu:
@bechi please ACK.

sironitomas:
Can anyone comment about this? How deep is the integration? What is Glidera exactly?

PR #3152, opened 2015-09-09
New release v1.2.0

cmgustavo:


matiu:
@dabura667 we are using the new release schedule now.

PR #3153, opened 2015-09-10
Allow addon menu items to run arbitrary code on click

troggy:
Another iteration of [this](https://github.com/bitpay/copay/pull/3069) PR
Now addresses this issue: https://github.com/bitpay/copay/pull/3069#issuecomment-138290326

Function ``setTab`` now accepts optional flag specifying whether to switch to ``walletHome`` state or not. This flag is set only for click handlers in bottom menu. This is not the proper fix, but rather a workaround. Proper fix would be to prevent ``setTab`` calling for states different from ``walletHome``


cmgustavo:
Ack. It works for me. Thanks!

Issue #3122, opened 2015-08-26
Show complete address list including current balance per address


isocolsky:
Related to #3119

Extra fields could include:
* creation date/time
* creator name
* address type (main, change)
* current balance
* used/unused flag

Addresses should be fully validated on the client side.

drewwells:
Did this feature get removed? I don't see a way to find a list of all addresses in the CoPay app

nycaleksey:
Has this feature been disabled? I no longer see an option to see all addresses associated with the wallet.

Issue #3119, opened 2015-08-25
Seeing every address of wallet


xanatas:
Is this possible without going through the history? Also unused addresses??

isocolsky:
There are several reasons why we decided against this:
* It encourages address reuse.
* It allows for a simpler and cleaner UI which improves usability.
* All addresses shown to the user have to be derived locally (to prove they actually belong to the wallet). This would have an impact in app performance.
Do you have an use case in mind that would require this feature?

dan-da:
This is pretty much the same as https://github.com/bitpay/copay/issues/3109.

There are many use cases (needs) for this:

1) User wishes to independently validate balance/transactions.
2) Importing addresses into libratax.
3) Importing addresses into my own auditing tool  ( bitprices ) or future auditing tools.
4) User wishes to lookup addresses on a blockchain explorer.
5) User is writing a script or using another utility/api/service that requires wallet addresses.
6) I'm probably not thinking of 1000 other cases.

Fundamentally though it comes down to this:  These are my addresses, representing my wealth.  Hiding them from me, the owner, is not acceptable.  It also requires that I, the user put 100% trust in the CoPay software, which I do not.   

In short, in your efforts to "simplify" and "protect" the user, you have effectively said "Don't worry your pretty little head about your keys or addresses, we'll just manage them them for you.  You want to see them?  too bad!".  This disempowers the user/owner, and to my mind, is only one step away from a hosted service like coinbase that actually holds the keys.

Presently extracting addresses out of CoPay is amongst the most laborious of any wallet I have seen.

One must go to history, then click on individual transaction, then click "See it on the blockchain" then manually determine which of the inputs/outputs correspond to our own address(es), then store those somewhere, then repeat all this for the next transaction, and so on.

I know this because I've done it, because I've had a NEED.  It is a serious pain point.

I am fine with it being buried under "Advanced" somewhere, or even as a separate utility or manually documented procedure.  But the capability needs to exist.

This could be easily solved with a new CSV export.

xanatas:
Actually my use case now:
I gave someone an address from my co-pay wallet two months ago. he used this address now to pay me and the funds are not showing up on my wallet but in the blockcain transactions they go to the address i gave him. I also checked It IS the address i gave to him. I actually want to reconstruct if something went wrong and what actually did happen.
Somehow the ability to prevent me from checking on all addresses keeps me from doing so.

[...]

Issue #3109, opened 2015-08-20
Need a way to export addresses used in transactions.


dan-da:
The CSV history does not presently include addresses, and the issue I opened for that functionality was closed without adding them.

As such, I am opening this issue to request an alternate method of obtaining the list of addresses that:

1) have ever been used in a transaction.
2) presently have unspent outputs.

I view (1) as most important and (2) as nice-to-have.

At present, I would be satisfied if clear documentation is provided for how the user may manually obtain (1) and hopefully also (2).

This is important for accounting (cost-basis) purposes.

Before closing, please have a look at these tools I've developed that make use of (1)
- https://github.com/dan-da/bitprices
- http://mybitprices.info

I would like to add documentation for how CoPay users may obtain (1) and (2).

thanks.

dan-da:
just a note that this is also useful for libratax, which has a "Bulk Add Addresses" feature in addition to CSV transaction import.

JDonadio:
Already done in https://github.com/bitpay/copay/pull/5070

Maybe we can add the ability to export. Thoughts?

JDonadio:
In the current version we have this feature complete.

PR #3162, opened 2015-09-12
add "advanced > wallet information"

matiu:
Adds Advanced -> Wallet Information

shows general Wallet information, xPubKeys, Balance by Address, and wallet main addresses.

fixes https://github.com/bitpay/copay/issues/3122
fixes https://github.com/bitpay/copay/issues/3119
related to: https://github.com/bitpay/copay/issues/3109

### DIFF
--- a/public/views/preferencesAdvanced.html
+++ b/public/views/preferencesAdvanced.html
@@ -7,6 +7,12 @@
 <div class="content preferences" ng-controller="preferencesAdvancedController as preferences">
 
   <ul class="no-bullet m0 size-14">
+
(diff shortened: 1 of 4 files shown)


cmgustavo:
ack

dan-da:
What have y'all got against (optionally) displaying change addresses?

1) If I'm checking out a block explorer such as blockchain.info I want to easily be able to see/verify that the address matches an address in my wallet.

2) For auditing purposes, change addresses are needed to verify funds stayed "here" and didn't go "elsewhere".

So why not add a checkbox to display them also?   The default could still be only "main" addresses.

PR #3168, opened 2015-09-15
Bug/various minor fixes

matiu:
- fixes WalletInfo when there is no wallets
- fixes WalletInfo cache issue (index.byAddress)
- better display of wallet seed.


cmgustavo:
ACK

PR #3186, opened 2015-09-17
fix Japanese phrase spacing for non ja language

dabura667:
This fixes the following bug:

1. Generate wallet in Japanese setting.
2. Switch device language to English.
3. view phrase for wallet
4. notice the ideographic space is missing. (words look very close together)

the [\u3041-\u308f] block was taken from here:
https://en.wikipedia.org/wiki/Hiragana_(Unicode_block)

All words in the list contain at least one of the characters in the block I set, and also, other languages should not have Japanese in them I think :-P


matiu:
ACK, good catch.

PR #3187, opened 2015-09-17
Disable Copying BIP39 phrase + facilitate copying on testnet

dabura667:
This is not secure. I believe we should encourage writing the phrase down by hand onto a physical medium.

Therefore I think we should leave the phrase as not copyable.

### DIFF
--- a/public/views/backup.html
+++ b/public/views/backup.html
@@ -74,8 +74,8 @@
 
     <div class="row" ng-show="show">
         <div class="columns">
-        <div class="p10 enable_text_select" style="background:#eee">
-          <span ng-repeat="word in wordsC.mnemonicWords"><span style="white-space:nowrap">{{word}}</span><span ng-show="wordsC.useIdeograms">&#x3000;</span> </span>
+        <div class="p10" style="background:#eee">
+          <span ng-repeat="word in wordsC.mnemonicWords"><span style="white-space:nowrap">{{word}}</span><span ng-show="word.match(/[\u3041-\u308f]/)!=null">&#x3000;</span> </span>  
         </div>
         </div>
     </div>
--- a/public/views/backup.html
+++ b/public/views/backup.html
@@ -74,7 +74,7 @@
 
     <div class="row" ng-show="show">
         <div class="columns">
-        <div class="p10" style="background:#eee">
+        <div class="p10" style="background:#eee" ng-class="{{index.network == 'testnet': 'enable_text_select'}}">
           <span ng-repeat="word in wordsC.mnemonicWords"><span style="white-space:nowrap">{{word}}</span><span ng-show="word.match(/[\u3041-\u308f]/)!=null">&#x3000;</span> </span>  
         </div>
         </div>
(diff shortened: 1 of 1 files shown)


dabura667:
In the event that copying is a feature we definitely want, I would like to modify this PR to auto-copy the phrase when tapped (so I may remove the extra spaces from the text when copying Japanese)

isocolsky:
I agree with @dabura667, we should encourage people to write down the seed. I still think copying the seed is no big deal if a passphrase has been set but for simplicity I would disallow copy altoghether.

sironitomas:
I agree. Just about any other application can read the clipboard. Having your master seed in that place is just not worth the risk.

public/views/backup.html
```diff
@@ -74,7 +74,7 @@
 
     <div class="row" ng-show="show">
         <div class="columns">
-        <div class="p10 enable_text_select" style="background:#eee">
+        <div class="p10" style="background:#eee">
```
matiu: Could you add
```
ng-class={{index.network == 'testnet': 'enable_text_select'}}
```
so in testnet is possible to copy the mnemonic?

dabura667: ok
 holdon

dabura667: added. Is this ok?

public/views/backup.html
```diff
@@ -74,7 +74,7 @@
 
     <div class="row" ng-show="show">
         <div class="columns">
-        <div class="p10 enable_text_select" style="background:#eee">
+        <div class="p10" style="background:#eee" ng-class="{{index.network == 'testnet': 'enable_text_select'}}">
```
cmgustavo: It should be:

```
ng-class=“{‘enable_text_select’: index.network == ’testnet’}”
```

matiu:
thanks!

ACK

public/views/backup.html
dabura667: Fixed

dabura667:
Manually tested.

Can't select with mainnet. Can select with testnet.

dabura667:
Since showing the \u3000 is only so that users don't get confused when looking at it, and if the person is copy-pasting anyways, they don't need to worry about looks or being able to read, they just need to copy.

So this way, now livenet people will see it shown as nice large spaces between so it is obvious where each word ends, they will not break in middle of words, AND testnet users can copy, and only 1 ASCII space will be copied per word.

dabura667:
rebased due to merge conflicts

dabura667:
@cmgustavo do you think we could get this into the release?

Allowing copying of phrase to clipboard is very unsafe.

Also, if we will allow copying, the current behavior will have Japanese people copying a phrase with two spaces between each word, and some other wallets won't be able to handle that (causing problems for users who copy/paste and try to restore with another wallet)

dabura667:
Separated the space handling commits into #3206 

This PR might be a UX decision, but #3206 is a bug fix. (using word.match() with the [\u...] character class was throwing errors in the browser and sometimes English phrases were showing the ideographic space.)

Please merge #3206 before release. Thanks.

PR #3190, opened 2015-09-17
Fix disclaimer

dabura667:
This should make it so that the TM will save, and we won't have to try to dig up old po files to copy.

### DIFF
--- a/i18n/po/template.pot
+++ b/i18n/po/template.pot
@@ -12,6 +12,10 @@ msgstr ""
 msgid "* A payment proposal can be deleted if 1) you are the creator, and no other copayer has signed, or 2) 24 hours have passed since the proposal was created."
 msgstr ""
 
+#: public/views/disclaimer.html
+msgid "<b>IF YOU LOSE ACCESS TO YOUR COPAY WALLET OR YOUR ENCRYPTED PRIVATE KEYS AND YOU HAVE NOT SEPARATELY STORED A BACKUP OF YOUR WALLET AND CORRESPONDING PASSWORD, YOU ACKNOWLEDGE AND AGREE THAT ANY BITCOIN YOU HAVE ASSOCIATED WITH THAT COPAY WALLET WILL BECOME INACCESSIBLE.</b>"
+msgstr ""
+
 #: public/views/backup.html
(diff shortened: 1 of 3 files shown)


cmgustavo:
I think the best way to do that is to link to the original document: https://copay.io/disclaimer

Something like this: 
```
<div ng-show="index.defaultLanguageIsoCode != 'en'>
  <a ng-click="$root.openExternalLink('https://copay.io/disclaimer')">Disclaimer</a>
</div>
```

What do you think?

dabura667:
Sounds good, will add.

Kirvx:
It's like you want.

dabura667:
```
<p ng-show="index.defaultLanguageIsoCode != 'en'">
  <a class="center" ng-click="$root.openExternalLink('https://copay.io/disclaimer')">Original Disclaimer</a>
</p>
```
Even with this, I have my device set to 'en' and the link still shows up... even if I use div instead of p it still shows when language is en

dabura667:
`index.defaultLanguageIsoCode` is `null` at the disclaimer screen.

cmgustavo:
Yes, it 's not so simple.

This is my solution (already tested locally):

src/js/controllers/disclaimer.js
```
 angular.module('copayApp.controllers').controller('disclaimerController',
-  function($scope, $timeout, storageService, applicationService, gettextCatalog, isCordova) {
+  function($scope, $timeout, storageService, applicationService, gettextCatalog, isCordova, uxLanguage) {
...
$scope.init = function() {
       storageService.getCopayDisclaimerFlag(function(err, val) {
+        $scope.lang = uxLanguage.currentLanguage;
         $scope.agreed = val;
         $timeout(function() {
           $scope.$digest();
...
```

public/views/disclaimer.html
```
<p class="text-center" ng-show="lang != 'en'">
  <a ng-click="$root.openExternalLink('https://copay.io/disclaimer')">Original Disclaimer</a>
</p>
```

dabura667:
All done. Have tested locally.

cmgustavo:
Great! LGTM

dabura667:
Can someone merge so I can upload to Crowdin and go to sleep? ( :-) )

dabura667:
Thanks!

Kirvx:
Nice it perfectly works, finally :)

PR #3198, opened 2015-09-18
better history refresh logic

matiu:
refresh tx history on new blocks only if there are tx with less that 6 confirmations


cmgustavo:
ACK

PR #3206, opened 2015-09-19
Fix spacing / validation of ja phrases

dabura667:
This contains a bug fix, so I separated it from #3187 

@cmgustavo If disabling copying is controversial, then just merge this please.

### DIFF
--- a/public/views/backup.html
+++ b/public/views/backup.html
@@ -74,7 +74,7 @@
 
     <div class="row" ng-show="show">
         <div class="columns">
-        <div class="p10 enable_text_select" style="background:#eee">
+        <div class="p10 enable_text_select" style="background:#eee;text-align: center;">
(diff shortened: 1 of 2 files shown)


matiu:
Closed on favor of https://github.com/bitpay/copay/pull/3187

Selecting is only enabled on testnet, and it will work on the roundtrip import-export on Copay. That seems to be all we need for now, so probably is better not to add more logic to the view.

dabura667:
This contains a bug fix.

The current master throws errors in the browser, and occasionally I get Japanese phrases that don't show the `\u3000` character.

By adding the logic to the js controller, this error no longer appears and it is consistent.

This might cause problems when compiled for Android / iOS... I don't have the environment to compile those.

I could make a separate PR just fixing the regex and relegating it to the controller.

PR #3210, opened 2015-09-19
fix seed for repeated words

matiu:


Kirvx:
Maybe my question is stupid but:

What would happen during the recovery of the seed if the user has a seed which is compatible for the french and for the english wordlist (given that the french wordlist has more than 100 words in common with the english wordlist)?

dabura667:
@Kirvx It would not matter, as the seed is generated from HASH( NORMALIZE( phrase ) ) so if the phrase is the same, then it doesn't matter if it is English or French.

However, if the 100 common words are in different positions in the wordlist, some apps would have trouble validating the phrase (checking the checksum contained within the phrase) and might throw an error.

@matiu Does Copay check the checksum of the phrases at any point? If it does, how does it decide which wordlist to check against?

dabura667:
```
English check OK:
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon angle
French check OK:
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon brave
both check OK:
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon amateur double
```
Here are 3 phrases, all using only the common words of English and French.

One only checks OK on the English wordlist, one only checks OK on the French wordlist, and the last one checks OK on both lists.

matiu:
Sent from my iPhone

> On Sep 19, 2015, at 10:29 PM, dabura667 <notifications@github.com> wrote:
> 
> @Kirvx It would not matter, as the seed is generated from HASH( NORMALIZE( phrase ) ) so if the phrase is the same, then it doesn't matter if it is English or French.
> 
> However, if the 100 common words are in different positions in the wordlist, some apps would have trouble validating the phrase (checking the checksum contained within the phrase) and might throw an error.
> 
> @matiu Does Copay check the checksum of the phrases at any point? If it does, how does it decide which wordlist to check against?
> 

Yes, we are checking the checksum. Check bitcore-mnemonic. The word lost is determined before the check. See function _belongsToWordlist and _getDictionary.

matiu:
Thanks for checking!

Kirvx:
Thanks guys :)

PR #3233, opened 2015-09-25
fix mnemonics with priv key is encrypted

matiu:
ask for password when private key is encrypted when showing mnemonics.

### DIFF
--- a/public/views/backup.html
+++ b/public/views/backup.html
@@ -5,8 +5,16 @@
 </div>
 
 
-
 <div class="content p20v" ng-controller="wordsController as wordsC">
+
+
+  <div class="box-notification" ng-show="wordsC.error">
+    <span class="text-warning">
+    {{wordsC.error|translate}}
+    </span>
+  </div>
+
+
 
   <div ng-show="wordsC.mnemonicWords">
     <div class="row"  ng-show="index.n==1">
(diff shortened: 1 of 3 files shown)


dabura667:
Just checking:

This update does not actually encrypt the phrase, it only refuses to show the phrase when password lock is present on private keys?

Why couldn't we just store the phrase encrypted and make sure to encrypt the phrase as well as the private key?

matiu:
On Sat, Sep 26, 2015 at 12:25 AM, dabura667 <notifications@github.com>
wrote:

> Just checking:
>
> This update does not actually encrypt the phrase, it only refuses to show
> the phrase when password lock is present on private keys?
>

Currently, when the priv key is encrypted, the phrase is also encrypted. So
if you go to backup it will show an empty mnemonic.

This PR will trigger a popup to ask the password to decrypt the privkey
(and the mnemonic) to show it to the user.



> Why couldn't we just store the phrase encrypted and make sure to encrypt
> the phrase as well as the private key?
>
Yes, IIUC that your are saying, that is what we are doing now. When
encryption is activated, both xprivkey and mnemonics and `.credentials` are
encrypted.


> —
> <https://github.com/bitpay/copay/pull/3233#issuecomment-143394278>.
>



-- 
Matías Alejo Garcia
@ematiu

dabura667:
Oh ok, so this PR will just allow you to decrypt the phrase when you want to look at it?

Right now if you have a password, you can not view the phrase without first disabling the password? Is that correct?

matiu:
On Sat, Sep 26, 2015 at 12:50 AM, dabura667 <notifications@github.com>
wrote:

> Oh ok, so this PR will just allow you to decrypt the phrase when you want
> to look at it?
>
Yes


> Right now if you have a password, you can not view the phrase without
> first disabling the password? Is that correct?
>

Exactly.



> —
> <https://github.com/bitpay/copay/pull/3233#issuecomment-143395129>.
>



-- 
Matías Alejo Garcia
@ematiu

Issue #3111, opened 2015-08-21
Advanced User Fee Control (formerly "Manual Override Dynamic Fee")


KeyJockey:
Hi guys... love the Copay wallet, great job overall _but_ really don't like the new "dynamic fee" feature as I prefer to have exact control over fee amounts (to keep my bitcoin balance precise).

Please add a fourth choice for the user to manually set preferred _exact_ miner fee (say .00005 to .0005 or maybe .0010 maximum range, maybe with a scroll wheel limiting so to protect newbies from sending whole bitcoins as "fee")

KeyJockey:
{Edit} Update, here's a simulated screenshot image for what I have in mind:
   
![poh12zv](https://cloud.githubusercontent.com/assets/13903392/9412354/3e5fe626-47f9-11e5-86f1-7be632a0f320.png)

The change from .2 version old settings to *only* offering this new dynamic fee system in .3 version is kind of a step backwards, i.e. forcing users to accept only the new indeterminate fee on the fly system now.  Less choice and *removing* feature functionality is not progress!

Putting in something like this would at least allow more control and *choice* to advanced users, even if default is to keep the dynamic system in place for most people... please consider it!

KeyJockey:
{Edit} One more update... I know it's usually a BAD idea but I suppose there might be some situations in which a user may deliberately want to send a transaction with NO miner fee at all.

Adding this manual override feature with ability to set the value down to 'zero' would allow Advanced Users to do that with Copay too... which is currently not possible at all.  

Other wallets (blockchain.info etc) for whatever reasons DO allow this manual setting of the fee level: would like to see Copay match and exceed all other wallets!  So you guys need to add this IMHO for that reason alone ;)

dabura667:
> Other wallets (blockchain.info etc) for whatever reasons DO allow this manual setting of the fee level: would like to see Copay match and exceed all other wallets!

Currently blockchain.info is the only wallet that does this, and almost every wallet developer unanimously thinks it's a dumb idea. (Not to mention the fact that blockchain.info is a bad wallet to begin with)

KeyJockey:
Hi Dabura... yeah, okay maybe it's not a good reason to add this after all... LOL

But, still.  Other points still stand I think.  More control over setting fee *exactly* for advanced users in certain circumstances doesn't really have any huge downside, does it?  

[...]

PR #3235, opened 2015-09-26
[WIP] Custom Flat Fees.

dabura667:
I don't think this will get merged, nor do I have a ton of time to work on it, nor do I have much skill with angular apps in general, but just making this PR to appease #3111.

Current Ideas:

- Are there any ways to make some sort of hurdle to using this feature? (Ideas: remove 'custom' from the "preferences" and only allow selecting it on a per transaction basis)
- Perhaps there should be a button to set it, and every time you click the button it pops up with a warning explaining why custom fees are a bad, horrible, insane idea.
- Perhaps there should be a high minimum fee for custom. (ie. The use case of "xyz service requires insane fees in order to do something" kinda makes sense, but imo that xyz service is stupid.) Setting a custom fee to 1000 satoshis just because you're a cheapskate should be discouraged. I propose minimum fee of `15000 satoshis x kB size of tx` for custom flat fee.
- Maybe have something like a special action to enable. (ex. you have to tap "Fee policy for this transaction" text 20 times in a row to enable the custom fee option etc.)

Current Questions:

- I can't seem to figure out where the feeLevels is actually being set. I have added an attribute to feeOpts with the string, but changing `feeService.getFeeLevels` for loop from i < 3 to i < 4 says that levelsLivenet[3] is undefined.

### DIFF
--- a/public/views/walletHome.html
+++ b/public/views/walletHome.html
@@ -403,7 +403,10 @@
                       ng-click="home.setFee(fee.level)" class="line-b p20">
                       {{index.feeOpts[fee.level]|translate}}
                       <i class="fi-check size-16 right"
-                        ng-show="(home.currentSendFeeLevel || index.currentFeeLevel) == fee.level"></i>
+                        ng-show="(home.currentSendFeeLevel || index.currentFeeLevel) == fee.level && fee.level != index.feeOpts['custom']"></i>
+                      <div class="input right" ng-show="(home.currentSendFeeLevel || index.currentFeeLevel) == fee.level && fee.level == index.feeOpts['custom']">
+                        <input type="number" id="customFee" ng-disabled="home.blockUx" name="customFee" ng-attr-placeholder="{{'Satoshis'|translate}}" ng-minlength="10000" ng-maxlength="10000000" ng-model="_customFee" valid-amount required autocomplete="off" ng-focus="home.formFocus('customFee')" ng-blur="home.formFocus(false)">
+                      </div>
                     </li>
                   </ul>
                 </div>
--- a/src/js/controllers/walletHome.js
+++ b/src/js/controllers/walletHome.js
@@ -680,6 +680,19 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
         configurable: true
       });
 
+    Object.defineProperty($scope,
+      "_customFee", {
+        get: function() {
+          return $scope.__customFee;
+        },
+        set: function(newValue) {
+          $scope.__customFee = newValue;
+          self.resetError();
+        },
+        enumerable: true,
+        configurable: true
+      });
+
     var fc = profileService.focusedClient;
     // ToDo: use a credential's (or fc's) function for this
     this.hideNote = !fc.credentials.sharedEncryptingKey;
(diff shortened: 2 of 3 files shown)


KeyJockey:
Hi there... I'm the 3111 guy, thanks for this but just wondering, really what's the so-big deal about allowing manual fee entry?  Maybe I'm not technical enough to get it but why is this idea seemingly so bad for some of you guys?

I mean, yeah I realize why the dynamic fee feature is *better* for saving money and ensuring that *enough* of a fee is allocated to a transaction in the event the network is very busy (stress test, etc) but so far in bitcoin's existence basically ALL wallets don't work this way, i.e. they all have basically fixed fee or at least just "simple" fee values set at 0.0001 or 0.0005 or whatever.

That's always worked fine, so far in all other wallets.  So *really* what's the big deal to just allow it as a manual user-selected "downgrade" feature for advanced-user specific purposes in Copay?

When I first suggested this I never thought it'd be so controversial... I figured it was simple, easy, no big deal really, just something that'd be nice to have for added user CONTROL and *choice*... so why do some of you guys really think it's so "bad" ...?

dabura667:
Throughout Bitcoin's history, the number of Bitcoin wallet projects you have maintained is 0, so I can understand your ignorance on the matter, but the *number 1* by far support request is:

"Hey uhhhh I sent a 0 fee transaction (read: I have no understanding how bitcoin works and am being a cheapskate) and it's been stuck for a whole day omg do something about it smh"

And a close second place:

"My transaction didn't confirm fast enough! So I lost money somehow, reimburse me omg smh!"

Dynamic fees are a step towards fixing these problems.

Just because something says "Advanced" does not mean new people ignore it, so there needs to be extra hurdles if this is going into the codebase.

Maybe a quiz asking them to ballpark the current average fee of the network, or something.

dan-da:
maybe a config setting they have to set manually with a text editor.

Or an if( false ) they have to manually set to true in the code with a big warning comment above it.   :-)   That would for sure weed out the noobs.

KeyJockey:
Hi guys... LOL okay yeah, I get it.

I am not a "programmer" (as I said) BUT YES actually I have had to deal with "stoopid noob user error" in a past career -- I was the primary UI designer on a fairly prominent niche Win95 utility software product (and the main documentation & help file author) 'way back in the day around 1992 thru early 2001 or so... so YEAH I've had to deal with such things, had to answer more than my fair share of bizarre tech support calls, etc too... so, yah okay sure, I get where you're comin' from now :+1:  

So okay then....so, how about THIS:

How's about we just make the advanced-user manual-fee scroller wheel value ONLY able to increment UPWARDS, not downwards at all, starting from whatever Copay has currently determined to be the correct MINIMUM fee value, and with some "maximum cap" sanity-level value control.

For example, say Copay shows current "Economy" dynamic fee is 0.00007 (60 minutes confirmation) and "Priority" dynamic fee is 0.000268 (10 minutes confirmation)...?  So in this case, the manual control fee scroll-wheel shouldn't be able to go lower than 0.00007 (i.e. no ZERO fee allowed, fine with me!) or any higher than, say, maybe like 0.0010 maximum (or maybe 0.0100 if we wanna be really crazy LOL)

Basically, as long as I can set something like this annoying 0.000268 to just "0.0003" simple... I'd be happy.

This work for you guys....?  Limits like this would mitigate the need for big angry scary warning messages or jump-thru-hoops UI elements?  Whattya think?  ;)

matiu:
Closing this one. To implement flat fees it would be needed to first implement them on BWS.

PR #3245, opened 2015-09-28
hide advanced options from send form

### LINKED ISSUE #3222: Remove advanced options from send tab

javierbitpay:
fixes #3222

### DIFF
--- a/public/views/walletHome.html
+++ b/public/views/walletHome.html
@@ -387,35 +387,6 @@
               </div>
             </div>
 
-            <div class="m20b" ng-init="home.hideAdvSend=true">
-              <a class="button outline light-gray expand tiny m0" ng-click="home.hideAdvSend=!home.hideAdvSend">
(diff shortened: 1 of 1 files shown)


dabura667:
So this will mean all transactions follow the wallet preferences?

matiu:
NACK.

On Mon, Sep 28, 2015 at 11:05 AM, dabura667 <notifications@github.com>
wrote:

> So this will mean all transactions follow the wallet preferences?
>
> —
> <https://github.com/bitpay/copay/pull/3245#issuecomment-143752294>.
>



-- 
BitPay.com

dabura667:
NACK as well. Until I hear a good reason.

isocolsky:
I am ok with this change. I don't think the advanced options are useful. It just adds complexity to the send tab with no real gain: it is not needed by casual users and it does not add enough control for power users. Have you ever needed to change these settings for a single transaction?

PR #3270, opened 2015-10-02
Feat/paper wallet

javierbitpay:
There is a problem when a private key that is not hexa is enter. An error appear in console but not on screen. Also, the spinner is loaded and don't stop.


matiu:
LGTM

Regarding the error metioned on the decription, it should be fixed on BWC:

https://github.com/bitpay/bitcore-wallet-client/issues/153

cmgustavo:
ACK. Great job!

PR #3283, opened 2015-10-07
disables TREZOR

matiu:
Disables TREZOR integration until a workaround to make TREZOR Connect API with Chrome Apps is found.


cmgustavo:
ACK

PR #3289, opened 2015-10-07
Set touch id to spend funds (only iOS)

cmgustavo:


matiu:
ACK!

dabura667:
Very nice! Exciting!

PR #3291, opened 2015-10-07
Self-signed certs from server

kleetus:
- Right now, the behavior as it pertains to accepting self-signed certs in Copay is that non-node web kit version accept and node web kits versions do not
- Added chromium args to make this behavior consistent

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "java": false,
     "plugin": false
   },
+  "chromium-args": "--allow-running-insecure-content --ignore-certificate-errors",
   "id": "jid1-x7bV5evAaI1P9Q",
   "homepage": "https://github.com/bitpay/copay",
   "license": "MIT",


koalalorenzo:
I am not an expert probably, but I think this is dangerous. Are we using HTTPS or encrypted communication between the clients and a BWS? Right?

dabura667:
Would be better to allow the user to manually add trusted certs to the Copay client.

> The fatal errors were preventing the script from running, so I just disabled all errors.

Never a good idea.

NACK

kleetus:
Copay currently supports self-signed certs. Unintentionally, self-signed certs aren't working for node web kit clients such as Android/iOS/Mac App whereas they do work for Copay clients such the browser, Chrome app. Another alternative to this PR is to remove self-signed cert support from all Copay clients. The drawback to this plan is that you either go get yourself a signed cert (which costs money) and a domain, or you run BWS in the clear (not a good idea). Yet another alternative would be to build a ui/ux experience into Copay similar to the browser, which would allow you to check cert to ensure the self-signed cert actually comes from the correct place. I am for the last one, but this PR does not address the ui/ux changes needed for Copay. Other options?

kleetus:
@matiu mentioned to me that this PR will not address Android/iOS/WP clients like I thought. I am closing this PR in favor of a new one. Possibly one that can be safer for people to use these certs.

PR #3298, opened 2015-10-08
fix menu-item

matiu:
@troggy there was an error with the touchdown event binding (`menu-walletHome` ID was repeated) so we had to change a `ng-show` with a `ng-if`. Please check you code still works.


cmgustavo:
ACK

troggy:
@matiu Thanks for letting me know. Quick check shows that ``menu-toggle`` thingie now works only once and then stops working.

How can I reproduce "touchdown event binding" error so I can check what's wrong with my code?

matiu:
Sure.

In a mobile device, when you tap (pushing down, without releasing) an icon,
the action must be triggered. Normally the action is triggered with
releasing the icon, not when pressing down.  The `bindTouchDown` in better
for responsiveness.



On Mon, Oct 12, 2015 at 3:48 PM, Kosta Korenkov <notifications@github.com>
wrote:

> @matiu <https://github.com/matiu> Thanks for letting me know. Quick check
> shows that menu-toggle thingie now works only once and then stops working.
>
> How can I reproduce "touchdown event binding" error so I can check what's
> wrong with my code?
>
> —
> <https://github.com/bitpay/copay/pull/3298#issuecomment-147489348>.
>



-- 
Matías Alejo Garcia
@ematiu

PR #3319, opened 2015-10-21
Fix strings to translate

cmgustavo:


public/views/create.html
```diff
@@ -132,9 +132,8 @@
             </label>
 
             <label  for="bws" class="oh">
-              <span translate>Specify Bitcore Wallet Service URL
-                <input type="text" id="bwsurl" name="bwsurl" ng-model="bwsurl">
-              </span>
+              <span translate>Specify Bitcore Wallet Service URL</span>
```
matiu: It should be "Wallet Service URL" only

public/views/import.html
```diff
@@ -85,9 +85,8 @@
               </label>
 
               <label  for="bws" class="oh">
-                <span translate>Specify Bitcore Wallet Service URL
-                  <input type="text" id="bwsurl" name="bwsurl" ng-model="bwsurl">
-                </span>
+                <span translate>Specify Bitcore Wallet Service URL</span>
```
matiu: see above

public/views/import.html
```diff
@@ -213,9 +211,8 @@
               <div class="large-12 columns">
 
                 <label  for="bws" class="oh">
-                  <span translate>Specify Bitcore Wallet Service URL
-                    <input type="text" id="bwsurl" name="bwsurl" ng-model="bwsurl">
-                  </span>
+                  <span translate>Specify Bitcore Wallet Service URL</span>
```
matiu: see above

public/views/preferencesAdvanced.html
```diff
@@ -27,7 +27,7 @@
  
     <li  ng-click="$root.go('preferencesBwsUrl')">
       <i class="icon-arrow-right3 size-24 right text-gray"></i>  
-      <span translate>Bitcore Wallet Service</span>
+      <span>Bitcore Wallet Service</span>
```
matiu: Wallet Service URL

cmgustavo:
Updated

PR #3321, opened 2015-10-21
Bug/csv

matiu:
add uniq for TXID history.


cmgustavo:
LGTM. Tested with my personal wallet. Total balance is correct.

PR #3345, opened 2015-10-26
fix signed

matiu:


afrikanhut:
There is something I have been looking for in all the wallet and payment systems ..I think developers need to put that into consideration, that is the possibilities of using ones call credits to buy and send.eg.if I don't have credit card ,no Bank account ,but I have smartphone and I recharge more two hundred dollars per month..it can possible to convert my credit to visible money, and I can use it to buy and pay and transfer.. What do you think about?

Issue #3173, opened 2015-09-15
Android blank white screen


XBTdev:
I recommended copay to a friend who installed the app and used it to store his bitcoins. That was a few months ago. Today he went to open the app to use his bitcoins and it shows a blank white screen after the splash screen (found a similar bug mentioned, #1519). He was running an old version, so we updated Copay to the latest version and tried again, same bug.

So running the latest version of copay we just see a white blank screen after the splash screen.

He has a backup, but we are unable to restore the backup because the password he remembers isn't accepted when we try to import the backup.

Any help?

matiu:
If this a 1-1 wallet?

Could he try to decrypt the backup using:
https://bitwiseshiftleft.github.io/sjcl/demo/ ?
(he should enter the backup text at Cyphertext and the password)




On Tue, Sep 15, 2015 at 4:33 PM, XBTdev <notifications@github.com> wrote:

> I recommended copay to a friend who installed the app and used it to store
> his bitcoins. That was 6 months ago. Today he went to open the app to use
> his bitcoins and it shows a blank white screen after the splash screen
> (found a similar bug mentioned, #1519
> <https://github.com/bitpay/copay/issues/1519>). He was running an old
> version, so we updated Copay to the latest version and tried again, same
> bug.
>
> He has a backup, but we are unable to restore the backup because the
> password he remembers isn't accepted when we try to import the backup.
>
> Any help?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitpay/copay/issues/3173>.
>



-- 
BitPay.com

XBTdev:
yes

matiu:
The decryption worked?

On Tue, Sep 15, 2015 at 4:43 PM, XBTdev <notifications@github.com> wrote:

> yes
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitpay/copay/issues/3173#issuecomment-140516006>.
>



-- 
Matías Alejo Garcia
@ematiu
Roads? Where we're going, we don't need roads!

XBTdev:
No. It says the password he remembers for his backup is incorrect.

matiu:
You mean than using https://bitwiseshiftleft.github.io/sjcl/demo/ is says the password is incorrect, right?

No, i wont work on the Copay. 6 months ago Copay was Beta and the storage format has changed in version 0.10 (https://github.com/bitpay/copay/releases/tag/v0.10.0,  ~ 5 month ago)

[...]

PR #3354, opened 2015-10-28
Handle error when the device id mismatch

cmgustavo:


cmgustavo:
Updated.

matiu:
LTGM.

This will give more info in this case: https://github.com/bitpay/copay/issues/3173
if use 'move' the storage from one device to other.

PR #3377, opened 2015-10-30
Improves the addressbook for mobile

cmgustavo:


matiu:
Nice!

I like `addressbook[]` much better that a directive.

In the current implementation of history, it is possible to have a listing for thousands of TX. It seems that `addressbook[]` will support it better.

matiu:
PS: Could you change that in txHistory also?

PR #3392, opened 2015-11-04
Adds limit to create addresses

cmgustavo:
Note: Require latest version of BWS

### DIFF
--- a/src/js/services/bwsError.js
+++ b/src/js/services/bwsError.js
@@ -94,7 +94,7 @@ angular.module('copayApp.services')
             body = gettextCatalog.getString('Wallet seed is invalid');
             break;
           case 'MAIN_ADDRESS_GAP_REACHED':
-            body = gettextCatalog.getString('You cannot generate more until you register any activity');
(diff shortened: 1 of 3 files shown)


src/js/services/bwsError.js
```diff
@@ -93,6 +93,9 @@ angular.module('copayApp.services')
           case 'INVALID_BACKUP': 
             body = gettextCatalog.getString('Wallet seed is invalid');
             break;
+          case 'MAIN_ADDRESS_GAP_REACHED':
+            body = gettextCatalog.getString('You cannot generate more until you register any activity');
```
matiu: Empty addresses limit reached. New addresses cannot be generated.

matiu:
note that this breaks copay with the current installed version of BWS.

cmgustavo:
Yes. Please, don't merge yet.

src/js/services/bwsError.js
cmgustavo: Thanks.

cmgustavo:
Updated with latest version of `angular-bitcore-wallet-client`

cmgustavo:
Updated

PR #3401, opened 2015-11-05
Fix resume event on mobile

### LINKED ISSUE #3273: App freeze/crash when using modal views
Observed on iOS9 (others not tested).  Two different scenarios:

A.
1. Launch Copay
2. Tap wallet settings icon
3. Minimize app
4. Restore app
5. Wait a few seconds; app minimizes by itself - when brough back to foreground it restarts at splash screen

B.
1. Launch Copay
2. Open Add wallet view
3. Minimize app
4. Restore app
5. App is unresponsive; must kill and restart

cmgustavo:
fixes #3273

### DIFF
--- a/src/js/init.js
+++ b/src/js/init.js
@@ -28,13 +28,20 @@ angular.element(document).ready(function() {
 
       document.addEventListener('pause', function() {
         if (!window.ignoreMobilePause) {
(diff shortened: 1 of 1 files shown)


matiu:
nice!

PR #3407, opened 2015-11-06
Incluiding address book and transaction history cache in export file

gabrielbazan7:
feat : #3395

### DIFF
--- a/src/js/services/profileService.js
+++ b/src/js/services/profileService.js
@@ -408,9 +408,12 @@ angular.module('copayApp.services')
       var addressBook = str.addressBook || {};
       var historyCache = str.historyCache ||  [];
 
-      root.setMetaData(walletClient, addressBook, historyCache, function(err) {
-        if (err) console.log(err);
-        root._addWalletClient(walletClient, opts, cb);
+      root._addWalletClient(walletClient, opts, function(err, walletId) {
+        if (err) return cb(err);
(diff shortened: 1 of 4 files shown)


src/js/services/profileService.js
```diff
@@ -324,6 +324,17 @@ angular.module('copayApp.services')
       });
     };
 
+    root.setMetaData = function(walletClient, addressBook, historyCache, cb) {
+
+      storageService.setAddressbook(walletClient.credentials.network, JSON.stringify(addressBook), function(err) {
```
matiu: Since address book is global (per network type) It seems it would be better to 'merge' the backup's addressbook with the current one, (and removing duplicates).

matiu:
Very nice work! LGTM. I left a comment to be discussed before merging.

src/js/services/profileService.js
cmgustavo: Agree.

cmgustavo:
Reviewing

gabrielbazan7:
updated

gabrielbazan7:
updated

gabrielbazan7:
updated

matiu:
LGTM

gabrielbazan7:
updated

cmgustavo:
LGTM

PR #3408, opened 2015-11-06
Ref delete wallet

cmgustavo:
* Also removes the lastAddress and backupFlag.


matiu:
ACK

dabura667:
Curious: If you delete a wallet on BWS, and then create a new wallet using an old phrase (that has been used by Copay, but deleted) will BWS scan the addresses and find the correct lastAddress upon initialization?

matiu:
Yes!

PR #3411, opened 2015-11-07
do not encrypt profiles in mobile

matiu:
Disables UUID encryption on mobile devices, due to random problems detected on some Android devices, getting the UUID at startup.  

Encryption on mobile devices using UUID does not add significant protection anyways, and the user can still encrypt the xprivkey using it own password.

Manually tested on iOS simulators and iPhone 6.


cmgustavo:
ACK

PR #3427, opened 2015-11-10
Auto-close modals after send/sign/reject a tx

### LINKED ISSUE #3422: Idea: autoclose popup after send/sign/reject a transaction
* The last screen after send/sign/reject a transaction without errors could auto close after a few seconds (instead to press `Okay` button to close).

cmgustavo:
* Fixes #3422
* Also ref `txStatus` service


cmgustavo:
Updated with @tonygallippi suggestion

PR #3435, opened 2015-11-11
fix join message

matiu:
- fixes refresh of status bar on mobile devices
- better 'connecting to trezor' message in create/join
- use text input in join (as in create)


cmgustavo:
LGTM

Issue #3461, opened 2015-11-14
Weird behaviour of send proposal in personal wallet


sipak:
In personal wallet, after interrupting singing process with Trezor, it is not possible to try again by clicking on Send again (coins are already locked) but the proposal in not shown automatically. It is not also shown in the wallet home until the app is restarted.

There are two way how to way how to discard it: Reject + Delete proposal. For personal wallet, there should be just one.

matiu:
Fixed by https://waffle.io/bitpay/copay/cards/5647a56dd7d883030026c28f

PR #3465, opened 2015-11-14
hide delete is n=1, show err if sign failed

matiu:
Fixes: #3461

(hide 'delete' if n=1); show error, when error is not an object (ie: from a hardware wallet app)


cmgustavo:
ACK

PR #3483, opened 2015-11-18
Bug/wallet list

matiu:
Fix empty wallets at the sidebar

This happen when incomplete wallet are recovered from seed (fixed in BWC https://github.com/bitpay/bitcore-wallet-client/pull/192), will be added in BWC in Copay 1.6.

For the moment, if the wallet completes, it will appear (with its name) at the list


matiu:
Closed in favor of: https://github.com/bitpay/copay/pull/3484

PR #3484, opened 2015-11-18
sanitize wallet list

matiu:
* Fix empty wallets at the sidebar
This happen when incomplete wallet are recovered from seed (fixed in BWC bitpay/bitcore-wallet-client#192), will be added in BWC in Copay 1.6.

If the wallet completes, it will appear (with its name) at the list

* Updates BWC with the mentioned fix
* Removes `angular-bitcore-wallet-client` dependency, and adds `bitcore-wallet-client`. Adds browserify to grunt.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,8 @@ webkitbuilds/*
 !webkitbuilds/build-osx.sh
 !webkitbuilds/Background.png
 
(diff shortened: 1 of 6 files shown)


cmgustavo:
LGTM

PR #3516, opened 2015-11-23
Fix disclaimer bug

gabrielbazan7:


gabrielbazan7:
update. i'm still not convinced of the way to solve the last issue (fix blank page) @cmgustavo

src/js/controllers/disclaimer.js
```diff
@@ -32,6 +36,8 @@ angular.module('copayApp.controllers').controller('disclaimerController',
       });
     };
 
-    create();
-
+    if (!profileService.profile)
+      create();
+    else
+      applicationService.restart();
```
cmgustavo: This action will restart Copay while I just want to view the disclaimer from Settings->About->Disclaimer.

gabrielbazan7: Yes. That problem not solved yet. I will make a new file that includes only the Terms of Use and not the icon and "welcome to copay" title.

gabrielbazan7:
updated

gabrielbazan7:
Conflicts resolved.

cmgustavo:
I found a specific issue: when I delete manually the flag `agreeDisclaimer` from localStorage, Copay delete all my current wallets and creates a new one.

cmgustavo:
For slow connections (or devices), I would like a spinner or something that indicating that Copay is loading/creating a wallet

cmgustavo:
iOS: Preferences -> About Copay -> Terms of Use

I can not scrolling the page down.

gabrielbazan7:
updated

gabrielbazan7:
updated

gabrielbazan7:
updated

cmgustavo:
ACK

PR #3525, opened 2015-11-25
Clarify language

jzbz:
"Word" is a bit confusing and redundant here. "Phrase" or "Mnemonic" might be better choices.


dabura667:
ACK

matiu:
Thanks!

PR #3565, opened 2015-12-02
Global settings terms of use

gabrielbazan7:
changes in global settings terms view and taking out auto capitalize in join and create view when writting wallet seeds and passphrase


cmgustavo:
ACK

PR #3573, opened 2015-12-02
Transaction history options in advanced

gabrielbazan7:


public/views/walletHome.html
```diff
@@ -506,7 +506,7 @@ <h4 class="title m0">
             <input id="export_file" type="file" nwsaveas="Copay-{{index.alias || index.walletName}}.csv" accept=".csv" style="display:none">
             <a class="button outline dark-gray round" ng-click="index.csvHistory();">
               <i class="fi-page-export-csv"></i>
-              <span translate>Download CSV file</span>
+              <span translate>CSV</span>
```
cmgustavo: I think that the attribute `translate` is not longer necessary.

gabrielbazan7:
updated

cmgustavo:
Missing add this to `animationService`

Issue #3629, opened 2015-12-09
Avoiding the accidental exit - second back button press to exit


gabrielbazan7:
Also, show a message that a second back button press is required in order to exit the app.

PR #3631, opened 2015-12-09
Second back button press to exit

gabrielbazan7:
#3629


gabrielbazan7:
Updated

cmgustavo:
Nice!

LGTM. Tested both Android and iOS

PR #3687, opened 2015-12-16
fix button on small device

### LINKED ISSUE #3685: iOS version stuck at step 1.
Most people won't understand that they have to scroll down to be able to click "agree with T&C"

Especially since scrolling scroll into the T&C instead of the whole page.

bechi:
Fixes #3685


cmgustavo:
ACK

PR #3692, opened 2015-12-17
Fix: configService.set() pollutes default config

troggy:
Call to ``configService.getDefaults()`` after ``configService.set(...)`` returned current config from storage, not the default config.


cmgustavo:
ACK

Issue #3701, opened 2015-12-18
Mixed languages on activity screen


jzbz:
After sending some bits to a fresh Copay install I noticed under "Activity" it showed the transaction and some Chinese characters appeared in lieu of the words "days ago" so instead of saying "2 days ago" there appeared the number 2 followed by two Chinese characters. This was rectified by going into Copay's settings (where no default language had been selected) and selecting "English" in my case.

Copay version: 1.6.3
Android version: 5.1

PR #3707, opened 2015-12-20
Fix/lang bug android

gabrielbazan7:
Fix : #3701

### DIFF
--- a/cordova/build.sh
+++ b/cordova/build.sh
@@ -94,6 +94,9 @@ if [ ! -d $PROJECT ]; then
     checkOK
   fi
 
+  cordova plugin add cordova-plugin-globalization
+  checkOK
+
   cordova plugin add cordova-plugin-splashscreen
   checkOK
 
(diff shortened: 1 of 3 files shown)


tanojaja:
LGTM

gabrielbazan7:
Updated

cmgustavo:
I think we shouldn't use `async` function to set the language, because it 's necessary at bootstrapping. Perhaps some Android devices have different versions of the browser and return a false positive (in this line: https://github.com/bitpay/copay/pull/3707/files#diff-85bd5aff1692eb058e100ad2aaf83cfaL45).

matiu:
Is this still valid?

gabrielbazan7:
Still valid, because copay doesn't recognize the language that is set by the user in some android phones. Now, with the latest update of Copay ( https://github.com/bitpay/copay/pull/3752 ), set english as default if this happen, so there will be not mixed languages on activity screen (this issue https://github.com/bitpay/copay/issues/3701 is solved). But, the problem still there.

cmgustavo:
Ok. It works for me (Android and NW).
Please rebase is needed.

gabrielbazan7:
Updated

cmgustavo:
LGTM. Tested on Android/NW

PR #3720, opened 2015-12-28
Fix ignoreMobilePause value on multiple pause/resume events

mpolci:
Hi, studying the source code I noticed a small bug in the pause/resume events handling. 
The bug is easily reproducible on mobile devices (I tried android):
* open copay
* go to wallet import 
* hide the app (go to device home with the home button) then reopen it, copay should remain in the wallet import state
* hide the app a second time and reopen it, now copay goes to the wallet home

Some controller like the importController sets the global variable `window.ignoreMobilePause = true` to avoid returning to walletHome state on resume event. In the init.js file on the resume event handler, the value of ignoreMobilePause is restored to false, the controller don't restore the "true" value so from the second resume event the skip doesn't work.

I opened this pull request with a possible solution, but **it is not tested**, I opened it for discussion.
In the proposed code the controllers that set to true ignoreMobilePause are responsible to restore the value to false. 
The importController and the qrScanner directive already do that, but not the other controllers.
Finally I removed the `window.ignoreMobilePause = false` in the resume event handler in init.js file.

### DIFF
--- a/src/js/controllers/copayers.js
+++ b/src/js/controllers/copayers.js
@@ -9,6 +9,14 @@ angular.module('copayApp.controllers').controller('copayersController',
     var cancel_msg = gettextCatalog.getString('Cancel');
     var confirm_msg = gettextCatalog.getString('Confirm');
 
+    if (isMobile.Android() || isMobile.Windows()) {
+      $scope.$on('$destroy', function () {
+        $timeout(function () {
+          window.ignoreMobilePause = false;
+        }, 100);
+      });
+    }
+
     self.init = function() {
       var fc = profileService.focusedClient;
       if (fc.isComplete()) {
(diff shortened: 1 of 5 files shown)


cmgustavo:
Hi @mpolci , 

I was able to reproduce the issue on iOS. But, I 'm not sure that it 's the behaviour we are waiting for. Some devices (Android) "emit" a call to `Pause` when open the camera and then a `Resume`. This last event does that Copay refresh the balance and back home automatically. So, that 's the reason we are using `ignoreMobilePause`. I 'm not sure if it works correctly removing `window.ignoreMobilePause = false;` from init.js. We should do some tests.
Regarding `window.ignoreMobilePause = true` on each controller, I think it 's not a good solution, because we need that Copay updates the balance and other things. So, IMO the bug is in the `import.js` controller (we should remove this block of code about `ignoreMobilePause`).

Anyway, as you said, This is open to discussion. Thanks for your PR.

matiu:
Closing this one. (outdated, the bug is fixed already). Thanks anyways for the contribution.

PR #3751, opened 2016-01-08
Fix automatic lock of focused client on timeout

mpolci:
The profileService.unlockFC() method automatically locks the fc after 2 seconds but it doesn't work.
Before calling lockFC() it checks fc.isPrivKeyEncrypted(), but it is false after unlocked the fc, and if it is true it is not necessary to call lockFC(). In my opinion the ritght test should be on fc.hasPrivKeyEncrypted().


matiu:
great catch, thanks a lot!

PR #3758, opened 2016-01-11
Speed up QR code scanner for iOS

cmgustavo:


cmgustavo:
Android doesn't work with two camera related plugins: ```Error:
	Element uses-feature#android.hardware.camera at AndroidManifest.xml:44:5 duplicated with element declared at AndroidManifest.xml:43:5```

PR is still in progress.

cmgustavo:
Updated

matiu:
reviewing

cordova/build.sh
```diff
@@ -84,8 +84,14 @@ if [ ! -d $PROJECT ]; then
   cordova plugin add https://github.com/florentvaldelievre/virtualartifacts-webIntent.git
   checkOK
 
-  cordova plugin add https://github.com/phonegap/phonegap-plugin-barcodescanner.git
-  checkOK
+  if [ $CURRENT_OS != "WP8" ]
```
matiu: Maybe is better too use $OSTYPE
(like `if [[ "$OSTYPE" == "win32" ]]; then`)

cmgustavo: I think that some platforms don't have $OSTYPE, so using $CURRENT_OS is more secure because it is passed from the command line

matiu: ok, could you please add a 'echo "Using plugin XXXX for WINDOWS PHONE" so if windows phone release version 9 we can at least track the error?

cmgustavo:
Updated.

PR #3760, opened 2016-01-11
Bug/glidera resume event

cmgustavo:
* Fix Glidera 2FA step on resume event
* Fix padding and close modals with the back button or swipe-right

### DIFF
--- a/src/js/init.js
+++ b/src/js/init.js
@@ -45,7 +45,9 @@ angular.element(document).ready(function() {
           }, 100);
         }
         setTimeout(function() {
-          window.ignoreMobilePause = false;
+          var loc = window.location;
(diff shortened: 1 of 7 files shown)


src/js/init.js
```diff
@@ -45,7 +45,9 @@ angular.element(document).ready(function() {
           }, 100);
         }
         setTimeout(function() {
-          window.ignoreMobilePause = false;
+          var loc = window.location;
+          var ignoreMobilePause = loc.toString().match(/(glidera|buy|sell)/) ? 'true' : 'false';
```
matiu: true / false should not be in quotes (')

cmgustavo: Right. Thanks!

cmgustavo:
Updated. Thanks for reviewing

PR #3788, opened 2016-01-20
WIP: Touchid before Unlock on iOS

cmgustavo:


src/js/controllers/walletHome.js
```diff
@@ -335,69 +335,70 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
...
-              $scope.$digest();
+              $timeout(function() {
+                $scope.$digest();
+              });
               return;
             }
 
-            profileService.signTxProposal(txp, function(err, txpsi) {
-              self.setOngoingProcess();
+            profileService.unlockFC(function(err) {
```
matiu: we should not call unlock if the wallet is not locked. that will ask a password.

matiu:
Looks good, but we need to test it in details.

I think we should remove the timeout in locking and add:
```
  // for all clients
   if (client.hasPrivateKeyEncrypted()) {
     client.lock();
   }
```
on app startup, so we can lock open wallet in case of a crash.

cmgustavo:
Agree. Thanks for your comment.

I 'll set it as a `WIP` PR until we decide a better solution

cmgustavo:
Updated.

PR #3789, opened 2016-01-20
adding control for non ios and android device in pushnotifications registration

gabrielbazan7:


gabrielbazan7:
Updated

gabrielbazan7:
Updated

PR #3794, opened 2016-01-21
Fix remove unsent tx. Extend time to auto locking a wallet

cmgustavo:


cmgustavo:
Closing this in favor of #3788

PR #3795, opened 2016-01-21
Send confirmation popup

### LINKED ISSUE #3116: Add confirmation popup after send

cmgustavo:
Fixes #3116

### DIFF
--- a/src/js/controllers/index.js
+++ b/src/js/controllers/index.js
@@ -391,7 +391,6 @@ angular.module('copayApp.controllers').controller('indexController', function($r
 
   self.setFeeAndSendMax = function(cb) {
 
-    self.feeToSendMaxStr = null;
     self.availableMaxBalance = null;
     self.currentFeePerKb = null;
 
@@ -1451,6 +1450,19 @@ angular.module('copayApp.controllers').controller('indexController', function($r
 
   $rootScope.$on('Local/SetTab', function(event, tab, reset) {
     self.setTab(tab, reset);
+  });
+
+  $rootScope.$on('Local/NeedConfirmation', function(event, txp, cb) {
+    self.confirmTx = {
+      txp : txFormatService.processTx(txp),
+      callback: function(accept) {
+        self.confirmTx = null;
+        return cb(accept);
+      }
+    };
+    $timeout(function() {
+      $rootScope.$apply();
+    });
   });
 
   $rootScope.$on('Local/NeedsPassword', function(event, isSetup, cb) {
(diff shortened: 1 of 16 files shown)


cmgustavo:
Updated and ready for review

src/js/services/txFormatService.js
```diff
@@ -26,7 +26,7 @@ angular.module('copayApp.services').factory('txFormatService', function(profileS
 
     var outputs = lodash.isArray(tx.outputs) ? tx.outputs.length : 0;
```
matiu: shouldn't be outputs = 1 if no tx.outsputs defined?

src/js/services/txFormatService.js
```diff
@@ -37,6 +37,7 @@ angular.module('copayApp.services').factory('txFormatService', function(profileS
       }, 0);
     }
 
+    tx.toAddress = tx.outputs[0].toAddress; // Get first address for single transactions
```
matiu: what about all txs? is tx.outputs[0] will always be defined?

src/js/services/txSignService.js
```diff
@@ -85,10 +85,47 @@ angular.module('copayApp.services').factory('txSignService', function($rootScope
         };
 
         return cb();
+
+      });
+    });
+  }; 
+
+  root.createTx = function(opts, cb) {
```
matiu: why txSignService includes createTx? and publishTx?

matiu: Maybe rename it to txService?

src/js/controllers/index.js
```diff
@@ -1453,6 +1451,19 @@ angular.module('copayApp.controllers').controller('indexController', function($r
     self.setTab(tab, reset);
   });
 
+  $rootScope.$on('Local/NeedConfirmation', function(event, txp, cb) {
```
matiu: should be 'needsConfirmation'

src/js/services/txFormatService.js
cmgustavo: It 's not necessary. The code outside the condition will complete the required info (toAddress, amountStr, AlternativeAmountStr, feeStr).

src/js/services/txFormatService.js
cmgustavo: You are right. It 's not always defined for old tx proposals. Fixing it right now.

cmgustavo:
Updated. Thanks for reviewing

matiu:
Code ACK, tested on browser.

PR #3824, opened 2016-01-27
agree button is not disable before creating profile for a moment

gabrielbazan7:


src/js/controllers/disclaimer.js
```diff
@@ -4,6 +4,7 @@ angular.module('copayApp.controllers').controller('disclaimerController',
   function($scope, $timeout, $log, profileService, isCordova, storageService, applicationService, gettextCatalog, uxLanguage, go) {
     var self = this;
     self.tries = 0;
+    $scope.creatingProfile = true;
 
     var create = function(noWallet) {
       $scope.creatingProfile = true;
```
cmgustavo: Duplicated?

cmgustavo:
ACK. Good catch!

Issue #3816, opened 2016-01-23
The note field on send window should be visible


bitcoinuser:
Hi, see the video uploaded to understand the problem.

Thanks
[2016_01_23_19_01_35.mp4.zip](https://github.com/bitpay/copay/files/102029/2016_01_23_19_01_35.mp4.zip)

bitcoinuser:
Suggestion: when android keyboard be active, hide the buttons request/activity/send

matiu:
fixed in copay 2.0

PR #3845, opened 2016-01-29
scroll up

### LINKED ISSUE #3535: If an error occurs when sending a payment, scroll up to the top so that the error is actually visible (which is not always the case, e.g. if the advanced options area is opened)

tanojaja:
Fixes #3535 #3816

### DIFF
--- a/public/views/create.html
+++ b/public/views/create.html
@@ -49,13 +49,13 @@
   </div>
 
   <form name="setupForm" ng-submit="create.create(setupForm)" novalidate>
-    <div class="box-notification m20b" ng-show="create.error">
+    <div class="box-notification m20b" id="notification" ng-show="create.error">
       <span class="text-warning">
         {{create.error|translate}}
       </span>
     </div>
     <div class="row">
-      <div class="large-12 columns"> 
+      <div class="large-12 columns">
 
         <div ng-hide="create.hideWalletName">
           <label><span translate>Wallet name</span>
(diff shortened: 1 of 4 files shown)


gabrielbazan7:
I think #3816 is not solved.

gabrielbazan7:
ACK #3535

cmgustavo:
ACK, it works on iOS. 

Agree with @gabrielbazan7 , this PR doesn't solve #3826 .

tanojaja:
Updated.

cmgustavo:
I ‘m reviewing a little more, and I figured it out that the `master` branch works as expected without this PR. Please, could you review that and confirm if this PR is necessary? The issue is old, and I can check that after pressing the send button, the page is scrolled up automatically

gabrielbazan7:
Discussing, we decided that is not necessary the scroll up in the send tab, considering that all the options fit correctly on the screen in a standard device and also the page is scrolled up automatically on ios.

gabrielbazan7:
ACK

tanojaja:
Updated

gabrielbazan7:
ACK

public/views/includes/menu.html
```diff
@@ -9,7 +9,7 @@
     <menu-toggle ng-show="index.menu.length > 6"></menu-toggle>
 </div>
 
-<div class="bottom-bar row collapse" ng-show="!index.notAuthorized">
+<div ng-hide="hideMenuBar" class="bottom-bar row collapse" ng-show="!index.notAuthorized">
```
cmgustavo: `hideMenuBar` should be in `walletHome.html` where this file is included.

tanojaja:
Updated

tanojaja:
Updated

tanojaja:
Updated

matiu:
Code ACK.

PR #3852, opened 2016-02-02
reduces the response time when close preferences

tanojaja:


src/js/controllers/topbar.js
```diff
@@ -1,13 +1,13 @@
...
-angular.module('copayApp.controllers').controller('topbarController', function(go) { 
+angular.module('copayApp.controllers').controller('topbarController', function(go, $rootScope) {
 
   this.goHome = function() {
     go.walletHome();
   };
 
   this.goPreferences = function() {
-    go.preferences();
+    $rootScope.$emit('Local/GoPreferences');
```
matiu: events should be events and not functions.

src/js/controllers/index.js
```diff
@@ -935,8 +936,10 @@ angular.module('copayApp.controllers').controller('indexController', function($r
       });
     });
   }
+
   self.showAllHistory = function() {
     self.historyShowShowAll = false;
+    self.shouldHideShowAll = true;
```
matiu: why there are two variables:

 - historyShowShowAll
and
 - shouldHideShowAll 

isn't that the same?

src/js/controllers/index.js
```diff
@@ -926,7 +926,7 @@ angular.module('copayApp.controllers').controller('indexController', function($r
         if (walletId == profileService.focusedClient.credentials.walletId) {
           self.completeHistory = newHistory;
           self.txHistory = newHistory.slice(0, self.historyShowLimit);
-          self.historyShowShowAll = newHistory.length >= self.historyShowLimit;
+          self.historyShowShowAll = newHistory.length > self.historyShowLimit;
```
tanojaja: "Show all" appears only if strictly greater than the limit

tanojaja:
Updated

tanojaja:
Updated. Thanks for comment!

cmgustavo:
ACK

PR #3853, opened 2016-02-02
Use cordova-plugin to paste from clipboard on mobile

### LINKED ISSUE #3832: App responsiveness review
Tested in IOS

Not OK:
starting from the not running
- [ ] opening / closing the wallet settings takes a while
- [ ] tapping on a tx to see its details take a while (closing it is faster)
- [ ] tapping on add  wallet / settings / buy & sell -> first close the side panel, and then shows the popup, it should be otherwise
- [x] the wallet icon on send to popup the address book has an action area too small
- [ ] Delay after pressing `Set/Ok` button in the popup if request password is required
- [ ] changing the focused wallet should not refresh the balance (the balance should be refreshed on background for wallet wallets).

OK:
 - qr reader popups and closes really fast!
 - switching between send/activity/recv
 - opening the side menu

cmgustavo:
Also fixes #3832 , #3859

### DIFF
--- a/public/views/modals/destination-address.html
+++ b/public/views/modals/destination-address.html
@@ -1,4 +1,4 @@
-<div ng-init="wallets[0] ? selectedWalletsOpt = true : selectedWalletsOpts = false">
+<div ng-init="wallets[0] ? selectedWalletsOpt = true : selectedWalletsOpts = false; checkClipboard()">
   <nav class="tab-bar" ng-style="{'background-color':color}">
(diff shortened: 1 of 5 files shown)


matiu:
reviewing

matiu:
What about copying the addr from the clipboard when focusing the send

cmgustavo:
Great idea!

cmgustavo:
Updated

PR #3858, opened 2016-02-04
Feat/clean home screen

### LINKED ISSUE #3793: Wallet name on two places (unnecessary)
I think on activity window is unnecessary show the wallet name on top. The wallet name is already showed.

screenshot_2016-01-20-21-50-13

cmgustavo:
Fixes #3793

copay_-_multisignature_wallet


matiu:
ACK

Issue #3029, opened 2015-07-24
Search transaction history


XBTdev:
It would be great if I could search transaction history. Current workaround is exporting to csv and searching.

matiu:
needs #3220

matiu:
It should also be possible to search by date. Maybe using a predefined syntax, like:
from:xx/xx/xx to:xx/xx/xx
(not practical for mobile users, thought)

PR #3862, opened 2016-02-05
Search box for transactions history

gabrielbazan7:
Feat : #3029

### DIFF
--- a/src/js/services/configService.js
+++ b/src/js/services/configService.js
@@ -105,7 +105,7 @@ angular.module('copayApp.services').factory('configService', function(storageSer
   };
 
   root.set = function(newOpts, cb) {
-    var config = lodash.clone(defaultConfig);
+    var config = lodash.cloneDeep(defaultConfig);
     storageService.getConfig(function(err, oldOpts) {
       if (lodash.isString(oldOpts)) {
         oldOpts = JSON.parse(oldOpts);


public/views/walletHome.html
```diff
@@ -219,7 +216,16 @@ <h4 class="title m0">
...
         </div>
 
-        <div ng-repeat="btx in index.txHistory"
+        <div id="searchLabel" class="left-inner-addon ">
+          <i class="fi-magnifying-glass"></i>
+          <input ng-click="home.searchInput()"
+                ng-focus="home.formFocus(true)"
+                ng-blur="home.formFocus(false)"
+                type="text"
+                placeholder="Amount, address, mm/dd/yyyy"
```
cmgustavo: Add filter `translate`, please.

src/css/main.css
```diff
@@ -609,6 +609,18 @@ ul.manage li {
   visibility: hidden;
 }
 
+.left-inner-addon {
+    position: relative;
+}
+.left-inner-addon input {
+    padding-left: 30px !important;
+    margin-bottom: 0rem !important;
```
cmgustavo: * `0` shouldn't have unit.
* Is `important` required?

gabrielbazan7:
Updated

public/views/walletHome.html
```diff
@@ -219,7 +216,16 @@ <h4 class="title m0">
          </div>
         </div>
 
-        <div ng-repeat="btx in index.txHistory"
+        <div id="searchLabel" class="left-inner-addon ">
+          <i class="fi-magnifying-glass"></i>
+          <input ng-click="home.searchInput()"
+                ng-focus="home.formFocus(true)"
```
cmgustavo: The value that receives `formFocus` isn't boolean. It 's an string with the names of the input.

src/js/controllers/index.js
```diff
@@ -948,6 +948,25 @@ angular.module('copayApp.controllers').controller('indexController', function($r
...
+      return lodash.includes(tx.amountStr, search) ||
+        lodash.includes(tx.message, search) ||
+        lodash.includes(self.addressbook[tx.addressTo], search) ||
+        lodash.includes(tx.addressTo, search) ||
+        lodash.isEqual(formatDate(new Date(tx.time * 1000)), search);
+    });
+    return result;
+  }
+
+  function formatDate(date) {
```
cmgustavo: Is this function used outside of `self.filter`?. If not, I think it should be inside `self.filter`.

gabrielbazan7:
Updated

public/views/walletHome.html
```diff
@@ -219,7 +216,17 @@ <h4 class="title m0">
          </div>
         </div>
 
-        <div ng-repeat="btx in index.txHistory"
+        <div id="searchLabel" class="left-inner-addon ">
+          <i class="fi-magnifying-glass"></i>
+          <input name="search"
+                ng-click="home.searchInput()"
+                ng-focus="home.formFocus('search')"
```
cmgustavo: I 'm not sure that it 's useful, because `search` doesn't exist in the controller. What 's the idea to do `ng-focus`?

gabrielbazan7: When I focus the input the menu bar hides. And I use the ng-blur to show the menu bar when the input has lost focus.

PR #3866, opened 2016-02-08
Fix error history.state in chrome app where window.history is not wri…

cmgustavo:
…teable for chrome 48.

```
history.state is not available in packaged apps.
```

Also fixes karma test.


cmgustavo:
Updated

cmgustavo:
Closed. It doesn't work on mobile.

PR #3872, opened 2016-02-11
Hide balance when is pressed for more than half a second

gabrielbazan7:


src/js/controllers/walletHome.js
```diff
@@ -28,6 +28,27 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
   this.addr = {};
   this.lockedCurrentFeePerKb = null;
 
+  self.hideBalance = function() {
+    var fc = profileService.focusedClient;
+    storageService.getHideBalanceFlag(fc.credentials.walletId, function(err, shouldHideBalance) {
+      if (err) self.shouldHideBalance = false;
+      else self.shouldHideBalance = (shouldHideBalance == 'true') ? true : false;
```
matiu: should be like https://github.com/bitpay/copay/pull/3872/files#diff-313c93846dfce71c19438c4889833d7fR43

no need to check against  string 'true'

JDonadio:
Would be nice hide all balances and not just the current.

src/js/controllers/walletHome.js
```diff
@@ -28,6 +28,27 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
...
+      if (err) self.shouldHideBalance = false;
+      else self.shouldHideBalance = shouldHideBalance ? true : false;
+    });
+  }
+  self.hideBalance();
+
+  self.onMouseDown = function() {
+    var fc = profileService.focusedClient;
+    self.mouseDown = $timeout(function() {
+      self.shouldHideBalance = self.shouldHideBalance ? false : true;
```
matiu: why this line?

src/js/services/storageService.js
```diff
@@ -198,6 +198,14 @@ angular.module('copayApp.services')
       storage.remove('config', cb);
     };
 
+    root.setHideBalanceFlag = function(val, walletId, cb) {
```
matiu: walletId should be the first parameter.

src/js/controllers/walletHome.js
```diff
@@ -28,6 +28,27 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
   this.addr = {};
   this.lockedCurrentFeePerKb = null;
 
+  self.hideBalance = function() {
+    var fc = profileService.focusedClient;
+    storageService.getHideBalanceFlag(fc.credentials.walletId, function(err, shouldHideBalance) {
+      if (err) self.shouldHideBalance = false;
+      else self.shouldHideBalance = shouldHideBalance ? true : false;
```
matiu: should be:
else self.shouldHideBalance = shouldHideBalance ;

cmgustavo:
I think we shouldn't add attributes that don’t work on mobile (like `onmousedown`). Could you use `ng-click` instead?

src/js/controllers/walletHome.js
gabrielbazan7: Self.shouldHideBalance should change from false to true when the balance is pressed for half a second, and also from true to false when the lock image is pressed.

src/js/controllers/walletHome.js
gabrielbazan7: This should be `self.shouldHideBalance = (shouldHideBalance == 'true') ? true : false;` because I receive a string value from local storage.

cmgustavo:
It doesn't work on iOS. Closed until found a new idea to implement

PR #3885, opened 2016-02-16
Correction for #3692, cloneDeep() must be used otherwise defaultConfi…

ajp8164:
…g gets overwritten.


cmgustavo:
Good catch! ACK

PR #3905, opened 2016-02-22
refactor disclaimer handling

matiu:
- refactor disclaimer handling
- fix CRITICAL timing issue when creating a new wallet, that makes that the new wallet is not inmediately save in certain devices (this only happends in MASTER not in the deployed versions)


cmgustavo:
LGTM. Tested on iOS and OSX (desktop)

PR #3908, opened 2016-02-23
Bug/scope

matiu:
fix delete wallet goHome
fix locking amount in send
fix changing wallet when a tab is selected


matiu:
updated and ready for review.

cmgustavo:
ACK

PR #3914, opened 2016-02-24
Fix build transaction method in ledger service

JDonadio:
Referred to this error:

Requesting Ledger Chrome app to sign the transaction
bwcService.buildTx is not a function


matiu:
Good catch.

Issue #3973, opened 2016-03-08
Crazy bug (Shared wallet)


bitcoinuser:
Hi,

1- As I can see, if I start a shared wallet, I can't delete it until the other copayer join it. Right? if Yes, Why?
2- To test, I created a 2of2 shared wallet, but the copayer didn't join it.
3- I closed the copay app and reopened it. 
4- Now, when I open the teste wallet, I can't delete it. 
5- If I switch to any other walltet, appear the message (Share this walllet...)

See the video to understand the problem. I think it's a serious bug.

[2016_03_08_16_26_08.zip](https://github.com/bitpay/copay/files/163828/2016_03_08_16_26_08.zip)

cmgustavo:
Hi @bitcoinuser , 

Yes, it 's a bug. We are working on a fix.
BTW, the remove wallet button while you are waiting for copayer is at the bottom.

Thanks.

![copay_-_multisignature_wallet](https://cloud.githubusercontent.com/assets/237435/13614482/783d930e-e54f-11e5-8516-f1fb8b5b1844.png)

bitcoinuser:
Hi,

I didn't see the buttons at first time. But now I can't even scroll down the window to see if the buttons appear. 

![screenshot_2016-03-08-17-31-22](https://cloud.githubusercontent.com/assets/1178856/13615413/0622f084-e554-11e5-9e23-e344d403a2ea.png)

matiu:
Thanks for the report, this is an important bug, we will fix it ASAP.

matiu:
fixed in 1.9.3

PR #3979, opened 2016-03-09
fixes SHARED Wallet status, and refresh until it changes

matiu:
Fixes https://github.com/bitpay/copay/issues/3973

and fixe refresh until it change for payment proposals.


cmgustavo:
LGTM

PR #3995, opened 2016-03-14
Adds more fee data to the confirmation popup

cmgustavo:
copay_-_multisignature_wallet

copay_-_multisignature_wallet


cmgustavo:
Updated

isocolsky:
Very nice! ACK

PR #4014, opened 2016-03-22
fix remaining time paypro

JDonadio:


src/js/controllers/walletHome.js
```diff
@@ -1107,26 +1107,24 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
...
-  function _paymentTimeControl(timeToExpire) {
-    var now = Math.floor(Date.now() / 1000);
-
-    if (timeToExpire <= now) {
-      setExpiredPaymentValues();
-      return;
-    }
+  function _paymentTimeControl(expirationTime) {
+    var countDown;
+    setEpirationTime();
```
isocolsky: you dropped an `x`:)

Issue #4013, opened 2016-03-21
Bug freezes UI with waiting ms wallet and "About Copay" view


ajp8164:
Recreate this issue doing the following:

1. Create a new shared wallet with entering only the wallet name and nickname (that's all I entered)
2. Ensure the "waiting wallet" is selected
3. Open sidebar and click "Settings"
4. Click "About Copay"

The view presented is "Global preferences" (which is incorrect) and the UI is frozen.  You have to close and reopen the app.  This bug is repeatable.

gabrielbazan7:
#4068

matiu:
fixes in Copay 2.0

PR #4025, opened 2016-03-28
Preferences global could not be open if focused wallet is not complete

gabrielbazan7:
fix : #4013


src/js/controllers/index.js
```diff
@@ -1084,6 +1084,7 @@ angular.module('copayApp.controllers').controller('indexController', function($r
   self.setCompactTxHistory = function() {
     self.isSearching = false;
     self.nextTxHistory = self.historyShowMoreLimit;
+    self.completeHistory = self.completeHistory ? self.completeHistory : [];
```
gabrielbazan7: fix **Cannot read property 'slice' of undefined**

gabrielbazan7:
moved to v1.9 #4036

PR #4035, opened 2016-03-29
If pn was not enabled by user, do not show pn switch in IOS

gabrielbazan7:
fix: #4032

### DIFF
--- a/src/js/controllers/preferencesGlobal.js
+++ b/src/js/controllers/preferencesGlobal.js
@@ -14,14 +14,24 @@ angular.module('copayApp.controllers').controller('preferencesGlobalController',
       this.feeOpts = feeService.feeOpts;
       this.currentFeeLevel = feeService.getCurrentFeeLevel();
       this.usePushNotifications = isCordova && !isMobile.Windows();
+      if (!typeof cordova.plugins.diagnostic != undefined && isMobile.iOS()) {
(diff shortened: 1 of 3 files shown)


bityogi:
I would suggest (for the sake of consistency) to use: `index.PNEnabledByUser`

`index.usePushNotifications && index.PNEnabledByUser`

I did notice that all methods and properties are on $scope itself, but that also could also be changed to have those methods directly on the controller rather than on $scope.

src/js/controllers/preferencesGlobal.js
```diff
@@ -17,6 +17,9 @@ angular.module('copayApp.controllers').controller('preferencesGlobalController',
       $scope.glideraEnabled = config.glidera.enabled;
       $scope.glideraTestnet = config.glidera.testnet;
       $scope.pushNotifications = config.pushNotifications.enabled;
+      cordova.plugins.diagnostic.isRemoteNotificationsEnabled(function(isEnabled) {
```
cmgustavo: What about WP? If the plugin doesn't support WP, we should check it before

cmgustavo:
I agree with @bityogi about the global variable name, but this solution in general is not Ok. I mean, you are only hiding the option (in config it 's enabled). What about if I want to enable it again?

cmgustavo:
Needs rebase.

gabrielbazan7:
Updated

cmgustavo:
ACK. Tested on iOS.

PR #4048, opened 2016-04-01
notify when a new version of copay is available

JDonadio:
Works with:

https://github.com/bitpay/bitcore-wallet-service/pull/489
https://github.com/bitpay/bitcore-wallet-client/pull/229


matiu:
NACK. We should query github, not BWs.

PR #4054, opened 2016-04-05
more storage for big wallet's history

matiu:
fixes storage for big wallets.

Windows PHone is still not support it for this (we use localStorage there, and the limitation is still 5MB)


cmgustavo:
ACK

PR #4068, opened 2016-04-11
Fix fee levels when there is no wallet and hide Glidera if wallet is not complete

gabrielbazan7:
fix : #4013

### DIFF
--- a/src/js/controllers/index.js
+++ b/src/js/controllers/index.js
@@ -295,6 +295,8 @@ angular.module('copayApp.controllers').controller('indexController', function($r
   self.updateAll = function(opts, initStatusHash, tries) {
     tries = tries || 0;
     opts = opts || {};
+    var fc = profileService.focusedClient;
+    if (!fc) return;
 
     var walletId = profileService.focusedClient.credentials.walletId
 
@@ -320,10 +322,6 @@ angular.module('copayApp.controllers').controller('indexController', function($r
         });
       }
     };
-
-    var fc = profileService.focusedClient;
-    if (!fc) return;
-
 
     // If not untilItChanges...trigger history update now
     if (opts.triggerTxUpdate && !opts.untilItChanges) {
--- a/src/js/controllers/index.js
+++ b/src/js/controllers/index.js
@@ -298,7 +298,7 @@ angular.module('copayApp.controllers').controller('indexController', function($r
     var fc = profileService.focusedClient;
     if (!fc) return;
 
-    var walletId = profileService.focusedClient.credentials.walletId
+    var walletId = fc.credentials.walletId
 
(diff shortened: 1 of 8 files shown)


isocolsky:
I believe this needs https://github.com/bitpay/bitcore-wallet-client/pull/235 to work. You should add 'WIP' until that gets merged.

src/js/controllers/preferencesGlobal.js
```diff
@@ -16,7 +16,7 @@ angular.module('copayApp.controllers').controller('preferencesGlobalController',
       this.usePushNotifications = isCordova && !isMobile.Windows();
       $scope.PNEnabledByUser = true;
       $scope.isIOS = isMobile.iOS();
-      if (!typeof cordova.plugins.diagnostic != undefined && $scope.isIOS) {
+      if ($scope.isIOS) {
```
gabrielbazan7: cordova is not define fixed

src/js/services/feeService.js
```diff
@@ -19,25 +19,32 @@ angular.module('copayApp.services').factory('feeService', function($log, profile
...
-    var fc = profileService.focusedClient;
     var unitName = configService.getSync().wallet.settings.unitName;
 
-    fc.getFeeLevels('livenet', function(errLivenet, levelsLivenet) {
-      fc.getFeeLevels('testnet', function(errTestnet, levelsTestnet) {
+    walletClient.getFeeLevels('livenet', function(errLivenet, levelsLivenet) {
+      walletClient.getFeeLevels('testnet', function(errTestnet, levelsTestnet) {
         if (errLivenet || errTestnet) $log.debug('Could not get dynamic fee');
         else {
           for (var i = 0; i < 3; i++) {
```
isocolsky: What if the server returns more than 3 levels?

gabrielbazan7:
Updated

src/js/controllers/preferencesGlobal.js
cmgustavo: `isMobile.iOS()` is for web-browsers. Use `isCordova` instead.

src/js/controllers/index.js
```diff
@@ -295,6 +295,8 @@ angular.module('copayApp.controllers').controller('indexController', function($r
   self.updateAll = function(opts, initStatusHash, tries) {
     tries = tries || 0;
     opts = opts || {};
+    var fc = profileService.focusedClient;
+    if (!fc) return;
 
     var walletId = profileService.focusedClient.credentials.walletId
```
cmgustavo: Use `fc` defined above.

gabrielbazan7:
Updated

gabrielbazan7:
Updated

cmgustavo:
ACK

PR #4088, opened 2016-04-16
translate global preferences

JDonadio:


dabura667:
I think this was done intentionally as those services are only available in the US or something.

matiu:
ACK

matiu:
I agree that the 'inner' pages of Glidera should not be translated, but the selected (one/off) are shown on all languages.

PR #4089, opened 2016-04-16
fix scan fingerprint value

### LINKED ISSUE #4086: ios - imported a wallet, Scan fingerprint option won't stay on
Imported a wallet onto ios device, find that I cannot turn on the "scan fingerprint" option for that wallet. Tried turning it off for the default wallet created, then on for my imported one. still doesn't work. the slider stays on but when I click "Back" it turns off again.

JDonadio:
Fixes #4086


matiu:
ACK.

PR #4092, opened 2016-04-18
Add error callback  to getFile() call

mpolci:
There is a missing error callback in dir.getFile() call and a duplicated parameter in the fileEntry.remove() call


matiu:
ACK

good catch. thanks.

Issue #4096, opened 2016-04-20
Invalid "send max" amount


braydonf:
Selecting "send max" produces an invalid amount:

![copay-invalid-amount](https://cloud.githubusercontent.com/assets/544777/14689497/8e34f52a-0715-11e6-9f4c-3bf324731728.png)

Version: 1.10.2

braydonf:
Looks like the amount is satoshis until this point where precision is lost:
https://github.com/bitpay/copay/blob/master/src/js/controllers/walletHome.js#L1300
```js
confirmDialog.show(msg, function(confirmed) {
  if (confirmed) {
    self._doSendMax(resp.amount * self.satToUnit);
  } else {
    self.resetForm();
  }
});
```

```js
resp.amount = 21287873
self.satToUnit = * 1e-8
```

```
> 21287873 * 1e-8
0.21287873000000002
```

bitcoinuser:
I have the same bug here.

![screenshot_2016-04-20-21-38-18](https://cloud.githubusercontent.com/assets/1178856/14694653/f917cdce-0740-11e6-878d-83e7a73c6989.png)

JDonadio:
There was a problem with the float point representation. I already sent a fix to solve this bug.
Thanks @braydonf @bitcoinuser for report.

braydonf:
Thanks for taking a look @JDonadio

braydonf:
This, from https://github.com/bitpay/copay/pull/4100, works in this particular case:
```js
> parseFloat((21287873 * 1e-8).toFixed(8)); 
0.21287873
```

However, since the amount is already in satoshis, and will ultimately be  satoshis, why not keep it in satoshis and then only display in the unit of choice?

PR #4100, opened 2016-04-21
Fix float point

JDonadio:
Fixes https://github.com/bitpay/copay/issues/4096

### DIFF
--- a/src/js/controllers/walletHome.js
+++ b/src/js/controllers/walletHome.js
@@ -1297,7 +1297,7 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
 
         confirmDialog.show(msg, function(confirmed) {
           if (confirmed) {
-            self._doSendMax(parseFloat((resp.amount * self.satToUnit).toFixed(8)));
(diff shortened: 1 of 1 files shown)


src/js/controllers/walletHome.js
```diff
@@ -1297,7 +1297,7 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
 
         confirmDialog.show(msg, function(confirmed) {
           if (confirmed) {
-            self._doSendMax(resp.amount * self.satToUnit);
+            self._doSendMax(parseFloat((resp.amount * self.satToUnit).toFixed(8)));
```
isocolsky: This is not right. Take a look at https://github.com/bitpay/copay/blob/master/src/js/controllers/walletHome.js#L611
The nb of decimal places is dependent on the selected unit.

JDonadio: Excelent @isocolsky , thanks!

gabrielbazan7:
ACK

PR #4115, opened 2016-04-22
Fix osx signing in script

matiu:
Fix OSX Code signing

(following https://github.com/nwjs/nw.js/wiki/MAS:-Signing-the-app )


cmgustavo:
ACK

PR #4135, opened 2016-04-26
Ref: Updates node modules. Split CSS in Sass files

cmgustavo:
Install Sass:

```
sudo gem install sass
```

And

```
~copay/$ npm install
```


matiu:
Great!!

Issue #4154, opened 2016-04-29
Clicking on the Push Notification doesn't open the notified wallet


cmgustavo:

matiu:
done

PR #4160, opened 2016-04-29
Tap on notification

gabrielbazan7:
fix : #4154

### DIFF
--- a/cordova/build.sh
+++ b/cordova/build.sh
@@ -93,6 +93,16 @@ if [ ! -d $PROJECT ]; then
     checkOK
   fi
 
+  if [ $CURRENT_OS == "IOS" ]; then
+    cordova plugin add phonegap-plugin-push --variable SENDER_ID="XXXXXXX"
+    checkOK
+  fi
+
+  if [ $CURRENT_OS == "ANDROID" ]; then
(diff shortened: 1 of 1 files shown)


gabrielbazan7:
There is a plugin error in version 1.2.3 for IOS devices in this feature.

gabrielbazan7:
Updated

cordova/build.sh
```diff
@@ -93,6 +93,16 @@ if [ ! -d $PROJECT ]; then
     checkOK
   fi
 
+  if [ $CURRENT_OS == "IOS" ]; then
+    cordova plugin add phonegap-plugin-push --variable SENDER_ID="XXXXXXX"
```
matiu: is XXXXXX ok??

gabrielbazan7: Sender id maps to the project number in the Google Developer Console. So, as we are using for only IOS application we can put in anything for this value.

cmgustavo: For iOS you set the latest version of the plugin, but for Android your are using an specific version (1.2.3). Is that ok too?

gabrielbazan7: Yes

gabrielbazan7:
Moved to v1.12 #4185

PR #4164, opened 2016-04-30
fix logs in txservice, add logs to bwc

matiu:
- Bump to version 1.12.3
- show source error after signature fail.


cmgustavo:
ACK

PR #4192, opened 2016-05-06
fix import from xPriv

matiu:
Fixes import a xPriv directly

(typing xprivxxxxxx ) at mnemonic input

Only supports livenet.


matiu:
depends con bwc 2.4.0 (https://github.com/bitpay/bitcore-wallet-client/pull/259)

cmgustavo:
ACK

PR #4270, opened 2016-05-31
Ref/platform info

matiu:
Refactor 

 - isMobile
 - isChromeApp
 - isDevel
 - nodeWebKit
 - isCordova

into platformInfo service.


matiu:
udpated.

Issue #3888, opened 2016-02-18
New Bitcoin network fee policy


franamati:
At this moment "Economy" policy is for an average confirmation time of 60 minutes. Sometimes I'm getting a USD 3 fee when using it, so if I'm not in a hurry, I would prefer to have another option with an average confirmation time of a few hours more to pay less.

For example, Bitcoin Core has an option for 25 blocks in average (4 hours). Copay could have one for 4 hours too. Thanks.

KeyJockey:
Linked this article in my #3111 report just now, but posting here too in case it's interesting for you...

https://blog.bitaccess.co/block-confirmation-myth/

miguelmorales85:
The same happened to me, to move just 0.7 BTC Copay calculated a "economic" 0.002269 BTC
I want the option to choose a manual fee, It could be an advanced feature but I want to choose a fixed fee.

KeyJockey:
Perhaps the devs here are getting "hung up" on us calling it a "fixed fee".

That is NOT exactly what I am asking for, i.e. it's perfectly FINE (and yes a very good starting point!) to have the software work as it currently does, to find the optimum current 'dynamic fee' FOR US based on network traffic etc.  That's great, and very cool to keep and no no no we do not want to throw that away!

All we want, really, is to be able to TWEAK that number for the "ideal fee" AS WE CHOOSE to either round it UP a bit and pay a little more sometimes, if we choose to pay a higher fee just to get priority service from the miners, or maybe bump it down a bit and pay less IF we are not in a huge hurry to get a confirmation right away and we want to save a few cents.

Honestly MOST users will probably just leave the default alone and accept whatever fee the Copay algorithm says they should use.  But for 'advanced users' who know what they want, we need CONTROL.

How on earth could anyone really be against this idea???  Srsly I am just blown away that it's taking *so long* to get this done, and that I don't have more support here in this thread... (???)

PR #4275, opened 2016-06-01
adding super economy fee option

gabrielbazan7:
fix:  #3888


miguelmorales85:
Can someone care to explain?

matiu:
Adds a new fee level of an estimated confirmation in 24 blocks (so now we have fee levels for target  confirmation in estimated 1,2, 6, and 24 blocks).

PR #4276, opened 2016-06-01
Feat/derivation test cache

matiu:
* Cache key derivation test, using client's user-agent.
* refactor angular-bwc's config to make it stateless.

### DIFF
--- a/src/js/services/profileService.js
+++ b/src/js/services/profileService.js
@@ -67,7 +67,6 @@ angular.module('copayApp.services')
       var defaults = configService.getDefaults();
 
       bwcService.setBaseUrl((config.bwsFor && config.bwsFor[walletId]) || defaults.bws.url);
-      bwcService.setTransports(['polling']);
     }
 
     root.setWalletClient = function(credentials) {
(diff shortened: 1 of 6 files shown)


matiu:
updated. Test fixed.

src/js/services/profileService.js
```diff
@@ -62,12 +62,11 @@ angular.module('copayApp.services')
       });
     };
 
-    root.setBaseURL = function(walletId) {
+    root._getBaseURL = function(walletId) {
```
JDonadio: Should not be "root" if it's an internal function

src/js/services/storageService.js
```diff
@@ -113,11 +113,14 @@ angular.module('copayApp.services')
     };
 
     root.storeProfile = function(profile, cb) {
+console.log('[storageService.js.115:storeProfile:]',profile); //TODO
```
JDonadio: Remove console.log

matiu:
updated. Thanks for the comments

JDonadio:
ACK

PR #4283, opened 2016-06-05
if profile exist do not try to re-create

gabrielbazan7:


src/js/controllers/disclaimer.js
```diff
@@ -7,32 +7,35 @@ angular.module('copayApp.controllers').controller('disclaimerController',
...
-              self.tries += 1;
-              create(false);
-            }
-          }, 3000);
-        } else {
-          $scope.error = "";
-          $scope.creatingProfile = false;
-        }
-      });
+          if (err === "EEXISTS") return self.init();
```
matiu: init() calls create(). I dont think this fix is OK.

PR #4294, opened 2016-06-07
better error when importing a wrong mnemonic

matiu:


src/js/controllers/import.js
```diff
@@ -52,7 +54,7 @@ angular.module('copayApp.controllers').controller('importController',
       try {
         str2 = sjcl.decrypt(self.password, str);
       } catch (e) {
-        err = gettext('Could not decrypt file, check your spending password');
+        err = gettext('Could not decrypt file, check your password');
```
isocolsky: Good catch.

PR #4306, opened 2016-06-08
Fix wording Chinese (Simplified)

cmgustavo:


src/js/services/uxLanguage.js
```diff
@@ -28,7 +28,7 @@ angular.module('copayApp.services')
       isoCode: 'ja',
       useIdeograms: true,
     }, {
-      name: '繁体中文',
```
dabura667: good thing you changed this. It says traditional chinese.

Also, 中国 is China the country and 中文 is the chinese language... so take the last two characters from the old one, and the parenthesis from the new one and you should be good I think.

cmgustavo: Updated #4312

Thanks!!

Issue #4304, opened 2016-06-08
[Ionic] UI problems


Kirvx:
### "Enregistrer" is truncated
![d6685118-2cfc-11e6-8779-9d36331ad004](https://cloud.githubusercontent.com/assets/6627786/15894851/7afe03d6-2d88-11e6-93e4-9eeff309ab3a.png)
### The switches hide the phrase
![sans titre](https://cloud.githubusercontent.com/assets/6627786/15894852/7afe5ec6-2d88-11e6-9502-6915874d600f.png)
### Untranslated strings
![sans titre4](https://cloud.githubusercontent.com/assets/6627786/15894911/d4bf3142-2d88-11e6-96c5-6c69e3582cfe.png)
### Should be like Copay 1.15 (deleting the `row` tags for each div fixes the problem)
![sans titre2](https://cloud.githubusercontent.com/assets/6627786/15894853/7b0268e0-2d88-11e6-886d-4cbdfcc31f04.png)

JDonadio:
@Kirvx thanks for this report. Regarding the 3rd screenshot, I cant see the problem. I'll send a fix for this bugs. Thanks again

Kirvx:
@JDonadio The three choices should be translatable :)

PR #4307, opened 2016-06-08
Fix ionic UI bugs

JDonadio:
Fixes https://github.com/bitpay/copay/issues/4304

No translations files updated


Kirvx:
Apparently it only solves the switch problem and there is a bug with the translations of the 3 choices (they are present in the .pot file but don't display translated).

JDonadio:
We'll only display "Ledger" or "Trezor" to simplify it. https://github.com/bitpay/copay/pull/4308

PR #4308, opened 2016-06-08
Fix translate - label width

JDonadio:


public/views/create.html
```diff
@@ -161,7 +161,12 @@
             </div>
             <div class="oh" ng-show="create.seedSourceId == 'new'">
               <label for="single-address" class="dbi">
-                <span translate>Single Address</span> <small translate>For audit purposes</small>
+                <div>
+                  <span translate>Single Address Wallet</span>
```
cmgustavo: Tip: add class `class="db"` to convert *span* in block

JDonadio: Awesome ! It's updated now. Thanks

cmgustavo:
ACK

Issue #3978, opened 2016-03-09
Image is mirrored when using the QR code scanner on some android phones


cjr-roland:
Follow up of https://github.com/bitpay/copay/issues/3663

This is a known issues on some recent android phones (tested on Nexus 5X, Copay version 1.9.2) which can easily be fixed http://www.xda-developers.com/why-images-in-some-camera-apps-are-upside-down-on-the-nexus-5x/

It has been fixed on some other bitcoin wallet (mycelium, Blockchain)

freeAgent85:
This has been bugging me since I got a Nexus 5X and it still hasn't been fixed as of v1.12.14.  It's very disorienting when attempting to scan a QR code as things are.

matiu:
We should consider migrating to https://github.com/phonegap/phonegap-plugin-barcodescanner/commits/master

instead of 
https://github.com/JrontEnd/phonegap-plugin-barcodescanner/commits/master

and check on a Nexus 5x device.

PR #4322, opened 2016-06-09
WIP: update phonegap-plugin-barcodescanner

matiu:
Fixes https://github.com/bitpay/copay/issues/3978

(according to https://github.com/phonegap/phonegap-plugin-barcodescanner/issues/175)


matiu:
NACK, build fails. See https://github.com/phonegap/phonegap-plugin-barcodescanner/issues/255

Issue #4318, opened 2016-06-09
[Bug] Fee selection


Kirvx:
The fee selection works visually with a double click to be set up, but if you click once on a choice (no visual change), go back, and return to the fee selection, the choice you have cliked will be selected.

Copay 1.12.15 and 2.0

Kirvx:
Also, note that if you switch to a different language, the language of the strategies will not change.

JDonadio:
Just to clarify, **Unit and Alt. Currency** should not be translated. Also, Super Economy (opt) is not translated yet.

Kirvx:
@JDonadio Not really fixed if you alternatively click a lot on "Normal" and "Economy", you will see that the confirm time and the rate is the same for both :)

JDonadio:
I was missing a refresh. Thanks a lot @Kirvx for this report!

Issue #4317, opened 2016-06-09
[Ionic] Spinner doesn't load for the fee preferences.


Kirvx:
Deleting the `class="spinner-stable"` in the `<ion-spinner>` fixes the problem.

JDonadio:
Done. Thanks @Kirvx !

PR #4323, opened 2016-06-09
Bug fee selection on mobile

JDonadio:
Fixes https://github.com/bitpay/copay/issues/4317 
Fixes https://github.com/bitpay/copay/issues/4318


cmgustavo:
ACK

PR #4337, opened 2016-06-09
Fix language selection - IOs

JDonadio:


src/js/controllers/preferencesLanguage.js
```diff
@@ -3,11 +3,14 @@
 angular.module('copayApp.controllers').controller('preferencesLanguageController',
   function($scope, $log, $timeout, configService, profileService, uxLanguage, walletService, go) {
 
-    this.availableLanguages = uxLanguage.getLanguages();
-    this.currentLanguage = uxLanguage.getCurrentLanguage();
+    this.init = function() {
+      this.availableLanguages = uxLanguage.getLanguages();
+      $scope.data = {
```
matiu: Why $scope.data.currentLanguage ? Lets use $scope.currentLanguage

public/views/preferencesLanguage.html
```diff
@@ -1,20 +1,12 @@
...
-  <ul class="no-bullet m0 ">
-  <li
-    ng-repeat="lang in prefLang.availableLanguages" 
-    ng-click="prefLang.save(lang.isoCode)" class="line-b p20 size-14">
-    <span>{{lang.name}}</span>
-    <i class="fi-check size-16 right" ng-show="prefLang.currentLanguage == lang.isoCode"></i>
-  </li>
-  </ul>
+
+  <ion-radio class="line-b size-12" ng-repeat="lang in prefLang.availableLanguages" ng-value="lang.isoCode" ng-model="data.currentLanguage"
```
matiu: I would prefer we dont user `$scope` on ng-model, but `prefLang`

if that is not possible, lets use $scope everywhere, and not `prefLang`

JDonadio:
Updated. Thanks @matiu !

matiu:
Code ACK.

I am in favor of migrating to ionic based components... what you think @cmgustavo ?

cmgustavo:
ACK. +1 to migrate all components to ionic. I just have a doubt with use of `this` or `$scope`. For example, this line in modal controller *txDetails.js*, I think it 's not good: 

```var self = $scope.self```

Anyway, the code LGTM to merge.

PR #4370, opened 2016-06-13
Replace $root.go - Use scope on preferences

JDonadio:
Related https://github.com/bitpay/copay/issues/4358

* Use scope only on preferences and preferences global
* Pending to replace $root.go on topbar.html. The state is used as a variable
* Fix test


src/js/controllers/preferencesEmail.js
```diff
@@ -1,21 +1,19 @@
 'use strict';
 
-angular.module('copayApp.controllers').controller('preferencesEmailController',
-  function($rootScope, go, profileService, gettext, $log, walletService) {
```
gabrielbazan7: $scope is not defined

gabrielbazan7:
ACK!

Issue #4336, opened 2016-06-09
Blank page if changes number of copayer/signatures when creating wallet


cmgustavo:

matiu:
IOS

matiu:
and at master

Issue #4328, opened 2016-06-09
On IOS, Using master, Create wallet leads to a blank screen


matiu:
Go to create wallet
 -> type a name
 -> click on 'advanced'

then the screen will turn white.

Could not reproduce this on Android nor in the browser, only IOS.

PR #4376, opened 2016-06-14
Fix IOs white page

JDonadio:
Fixes https://github.com/bitpay/copay/issues/4336
Fixes https://github.com/bitpay/copay/issues/4328


gabrielbazan7:
ACK !

PR #4382, opened 2016-06-14
Ref/validation

matiu:
- scanning wallet -> updating wallet
- uses the new BWC 2.7.0
- validate the wallet in background
   ^  works great in IOS.


cmgustavo:
LGTM

PR #4409, opened 2016-06-15
Remove handle imported client method

JDonadio:
After import a wallet -> start scan, backup flag and address flag events are triggered more than once.


matiu:
well done!

PR #4428, opened 2016-06-16
Adds editorconfig

cmgustavo:


dabura667:
This is good, but if this is something you want to enforce, maybe creating a style guide (or making it more apparent if there is one) would be better?

matiu:
@dabura667 We should describe on the README that we are using editor config and link it.

Issue #4427, opened 2016-06-16
Error at deleting recovery phrase


JDonadio:
Cannot delete the recovery phrase > "ERROR: credentials must have .walletId"

PR #4429, opened 2016-06-16
Fingerprint request in export and backup

JDonadio:
Fixes https://github.com/bitpay/copay/issues/4423
Fixes https://github.com/bitpay/copay/issues/4427

### DIFF
--- a/src/js/controllers/backup.js
+++ b/src/js/controllers/backup.js
@@ -1,9 +1,10 @@
 'use strict';
 
 angular.module('copayApp.controllers').controller('backupController',
-  function($rootScope, $scope, $timeout, $log, go, lodash, profileService, gettext, bwcService, bwsError, walletService, ongoingProcess) {
+  function($rootScope, $scope, $timeout, $log, go, lodash, fingerprintService, platformInfo, configService, profileService, gettext, bwcService, bwsError, walletService, ongoingProcess) {
 
     var fc = profileService.focusedClient;
+    var prevState;
     $scope.customWords = [];
     $scope.walletName = fc.credentials.walletName;
 
@@ -23,7 +24,8 @@ angular.module('copayApp.controllers').controller('backupController',
         initWords();
     }
 
-    $scope.init = function() {
+    $scope.init = function(state) {
+      prevState = state;
       $scope.passphrase = '';
       $scope.shuffledMnemonicWords = shuffledWords($scope.mnemonicWords);
       $scope.customWords = [];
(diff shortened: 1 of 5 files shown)


src/js/controllers/backup.js
```diff
@@ -40,11 +56,8 @@ angular.module('copayApp.controllers').controller('backupController',
       return false;
     };
 
-    $scope.backTo = function(state) {
-      if (state == 'walletHome')
-        go.walletHome();
-      else
-        go.preferences();
+    $scope.backTo = function() {
```
matiu: The function should not be called `backTo...`  and receive not params.

Maybe just `goBack()`

src/js/controllers/backup.js
```diff
@@ -40,11 +56,8 @@ angular.module('copayApp.controllers').controller('backupController',
...
+      go.path(prevState);
```
matiu: add `prevState || 'walletHome' ` in case it is not defined.

src/js/controllers/export.js
```diff
@@ -1,15 +1,29 @@
...
+
+      var config = configService.getSync();
+      var touchidAvailable = fingerprintService.isAvailable();
+      var touchidEnabled = config.touchIdFor ? config.touchIdFor[fc.credentials.walletId] : null;
+
+      if (!touchidAvailable || !touchidEnabled) return;
+
+      fingerprintService.check(fc, function(err) {
+        if (err)
+          go.path($scope.goBackToState);
```
matiu: Where is defined `$scope.goBackToState` ?

matiu:
Good work!

I left some minor comments.

src/js/controllers/export.js
JDonadio: topbar.html is used by _ng-include_ and set the $scope variable 

````
<div
  class="topbar-container"
  ng-include="'views/includes/topbar.html'"
  ng-init="titleSection='Export Wallet'; goBackToState = 'preferencesAdvanced'">
</div>
````

JDonadio:
Updated. Thanks for comment @matiu!

matiu:
Code ACK.

I will test it manually tomorrow for the final ACK.

cmgustavo:
I found some issues:

* Export: asks for touchID and tries to open the system keyboard. Check if there is an autofocus or something in the input.

img_1029

* Backup: TouchID and spending password request open at the same time. 

img_1030

* Also backup: if asks only TouchID, hide background content (as when request spending password)

img_1032

Issue #4395, opened 2016-06-15
Remove Fundation dependency


matiu:
given that we are using Ionic now.

PR #4430, opened 2016-06-16
remove angular foundation and use only ionic modals

gabrielbazan7:
fixes : #4395

### DIFF
--- a/src/js/controllers/modals/txStatus.js
+++ b/src/js/controllers/modals/txStatus.js
@@ -1,0 +1,11 @@
+'use strict';
+
+angular.module('copayApp.controllers').controller('txStatusController', function($scope, $timeout) {
+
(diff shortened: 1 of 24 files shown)


src/js/controllers/modals/txStatus.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+angular.module('copayApp.controllers').controller('txStatusController', function($scope, $timeout, profileService) {
```
JDonadio: profileService is not used

cmgustavo:
Scann QR only works the first time. If you open it again, it doesn't work.

cmgustavo:
Except the QR scan, LGTM.

gabrielbazan7:
Updated

gabrielbazan7:
Updated

cmgustavo:
ACK

PR #4433, opened 2016-06-16
Fix: modal choose wallet. Sell in Coinbase and Glidera. Spinner

cmgustavo:
* Fixes selection of wallet client correctly after choose one to send funds (sell Coinbase or Glidera)
* If `sellPrice` API call fail, disable Sell button.
* Disable spinner for Glidera and Coinbase from `index.js`
* Enable remove botton if transaction has error (Coinbase)
* On confirm transaction popup, if user cancel it, return to Coinbase/Glidera (selling process)

### DIFF
--- a/src/js/controllers/modals/wallets.js
+++ b/src/js/controllers/modals/wallets.js
@@ -2,25 +2,24 @@
 
 angular.module('copayApp.controllers').controller('walletsController', function($scope, bwsError, profileService) {
 
-  var self = $scope.self;
-
   $scope.selectWallet = function(walletId, walletName) {
-    if (!profileService.getClient(walletId).isComplete()) {
-      self.error = bwsError.msg({
-        'code': 'WALLET_NOT_COMPLETE'
-      }, 'Could not choose the wallet');
-      self.error = {
-        errors: [{
-          message: 'The Wallet could not be selected'
-        }]
-      };
-      $scope.cancel();
-      return;
-    }
-    self.selectedWalletId = walletId;
-    self.selectedWalletName = walletName;
-    self.fc = profileService.getClient(self.selectedWalletId);
-    $scope.cancel();
+
+    var client = profileService.getClient(walletId);
+    $scope.errorSelectedWallet = {};
+
+    profileService.isReady(client, function(err) {
+      if (err) { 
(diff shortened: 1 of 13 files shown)


src/js/controllers/modals/wallets.js
```diff
@@ -2,25 +2,24 @@
...
+    profileService.isReady(client, function(err) {
+      if (err) { 
+        $scope.errorSelectedWallet[walletId] = bwsError.msg(err);
+        return;
+      }
+
+      var obj = {
+        'walletId': walletId,
+        'walletName': walletName,
+        'client': profileService.getClient(walletId)
```
matiu: no need to send the client or the wallet name.  The called can call profileService and get that info. Better only return the walletId.

src/js/services/onGoingProcess.js
```diff
@@ -30,8 +30,16 @@ angular.module('copayApp.services').factory('ongoingProcess', function($log, $ti
     'deletingWallet': gettext('Deleting Wallet...'),
   };
 
+  root.isProcessing = false;
```
matiu: Why we need `isProcessing`?

src/js/services/onGoingProcess.js
```diff
@@ -30,8 +30,16 @@ angular.module('copayApp.services').factory('ongoingProcess', function($log, $ti
...
+
   root.clear = function() {
     ongoingProcess = {};
+    root.isProcessing = false;
+    if (isCordova) {
```
matiu: good catch

src/js/controllers/modals/wallets.js
cmgustavo: +1

src/js/services/onGoingProcess.js
cmgustavo: I 'm not sure if it 's the best way to block some function in the controller walletHome, like a `loading` action or something. I just wanted to prevent this line: 

```
-    if (form.address.$invalid && !ongoingProcess.get('fetchingPayPro')) {
 +    if (form.address.$invalid && !ongoingProcess.isProcessing) {
```

If this don't like it, I agree to leave it as it was until we find another solution.

cmgustavo:
Updated. Thanks for your comments.

matiu:
testing.

PR #4434, opened 2016-06-16
Fix UI delete wallets

cmgustavo:


src/js/controllers/index.js
```diff
@@ -1656,12 +1656,6 @@ angular.module('copayApp.controllers').controller('indexController', function($r
       self.isComplete = null;
       self.walletName = null;
       uxLanguage.update();
-
```
matiu: nice! Better this way.

Issue #4435, opened 2016-06-16
Scroll not working at wallet selection (iOS)


franamati:
On my iPhone 5s, after upgrading to v2.01, I can´t scroll to reach after the 6th wallet. This also blocks "add wallet" and "settings" since these options are below the wallet selection.

I later noticed that I can scroll using only the lower part, but it should work using the finger in any part of the wallet selection, like before. Also, once I'm able to scroll-down I can no longer scroll-up to reach the first wallets, so I need to shutdown the app to reach them (that resets the scroll).

Example:
![img_6651b](https://cloud.githubusercontent.com/assets/8605599/16136655/dbf98428-3402-11e6-87c5-61269af95f72.png)

matiu:
Thanks for the report.

I was able to reproduce the error on iphone 5s. but it is not happening on iphone 6/6s or in other platforms.

Are a work around:
 - you can scroll down from the botton of the side menu (as you mentioned)
 - it seems that closing and opening the side menu twice (from the "hamburger" button) reset the scroll to top.

franamati:
The "hamburger" workaround is not working for me, but I can survive shutting it down and reopening, no problem.

At first I was afraid of not being able of accessing my money, but scrolling from the bottom saved the situation. Thanks.

matiu:
@franamati Once the scroll is down, can you try selecting the top wallet you see? Does that make the scroll to be on top after you open the side menu again? thanks.

franamati:
Correct, that workaround worked and its far better than shutting down and reopening. Thanks!

PR #4437, opened 2016-06-17
fix side bar scroll - iphone5s issue

gabrielbazan7:
fixes: #4435


matiu:
reviewing

matiu:
ACK, tested on browser + 5s simulator.

PR #4442, opened 2016-06-17
skip device validation in ios

matiu:


cmgustavo:
ACK. Tested on iPhone6+

Issue #4189, opened 2016-05-05
Error Exporting wallet including history cache


JDonadio:
If addressbook and history cache are included, the export file generated is too long to copy on mobile clipboard, so it's not possible to import the wallet with this option. The size of this json file depends exclusively of the number of txs in the history, so if there is more than ~40/50 the import is not possible. Tested on android

matiu:
Lets make 2 switches: one for address book and one for tx history, and lets hide the tx history switch in mobile.

JDonadio:
I think the addressbook should be included always and only ask for history cache.

PR #4447, opened 2016-06-17
Export refactor

gabrielbazan7:
- Remove including history cache option from export (long transactions histories are imposible to copy) fixes:  #4189

- Remove "backup done" flag from export (the user should always get through 12 words flow)

- Include always address book when exporting file.


cmgustavo:
LGTM.

Issue #4462, opened 2016-06-20
Can't access any wallet or option


franamati:
After updating to v2.0.2 on iOS v9.3.2 (iPhone 5s) when opening Copay it shows "Could not access storage" and I can't access any of my wallets:
![error](https://cloud.githubusercontent.com/assets/8605599/16213275/90b840ce-3724-11e6-8d82-83b7c48a577d.png)

Pressing "OK" shows a blank screen with no options. Shutting down Copay or iOS and reopening doesn't help.

matiu:
Hi Franco,

We are sorry you  are having problems with v2.0.2. The error you are getting is when Copay just cannot access the physical device memory or the profile is corrupted.

1. Did you upgrade from v2.0.1? 
2. Anything we should know about the storage? (what it full? did you clean it? Is your device jailbroken?)
3. Do you have backups of your wallets?

thanks.

franamati:
1. I had v2.0.1 and I updated. v2.0.2 worked once, and I tested the scrolled issue that was fixed. The 2nd time I opened v2.0.2 I had this error.
2. I have 3.4 gb free on the phone and is not jailbroken. "Documents & Data" for Copay is 1.7mb at this moment.
3. I have backup on every wallet but one. And the one that not it's a 1/4 multisig, so money is not at risk.

I will now delete the app and reinstall, since no option is even shown in the screen, I guess there is no other way.

dabura667:
I had the same issue.

Reinstall might fix it, but I might be able to wait for a fix and hope my storage wasn't corrupted.

matiu:
@dabura667 thanks for the report, we are working to reproduce it. will keep this ticket updated.

matiu:
We were able to reproduce it on an iphone 5S. We are working in this.

Cyberfunks:
We are experiencing an almost identical problem in Android. ("code":"JSON error"). Cannot access wallet. Uninstalled and updated to v2.0.2 release and the problem persists.

matiu:
@Cyberfunks we dont that is related to this bug. Could you send a screenshot?

matiu:
@dabura667 The fix has been merged, we are releasing v2.1now.

matiu:
This is fixed in 2.1, which is already on app stores.

thanks for you patience, and feedback.

PR #4467, opened 2016-06-21
Bug/ios profile

matiu:
add locks to fileStorage
add lock to validation
read duplicate profiles
Fixes: https://github.com/bitpay/copay/issues/4462


isocolsky:
ACK!

PR #4475, opened 2016-06-21
internal notification should use default wallet color

### LINKED ISSUE #4474: Internal notification: after scan finished, it has incorrect color of the default wallet.
copay_-_multisignature_wallet

gabrielbazan7:
fixes #4474


cmgustavo:
ACK

PR #4486, opened 2016-06-22
remove pull to refresh when side bar is open

gabrielbazan7:


gabrielbazan7:
moved to master branch https://github.com/bitpay/copay/pull/4488

PR #4499, opened 2016-06-23
Show amount & fees with full precision on tx details modal

### LINKED ISSUE #3125: Show FULL Bitcoin Value/Balance, Some Places...
I realize Copay is truncating or "rounding" the bitcoin values in order to help make the user interface more simple and clean, and generally this is a good idea for anywhere it's just a "displayed value" such as Main screen in the big colored box that shows current wallet balance.

Keeping this to .0000 four or five decimal places given BTC's current exchange rate of the 4th digit equal to just a couple cents is fine... no problem with that :smile: 

However, in other places it *really* should be showing the WHOLE, exact bitcoin value... for example on the SEND page at the top, where Available Balance is displayed.

donthidebalance

This is a small font and there's plenty of room for 8 decimal places here: THIS page is where the user will be entering his *actual* next transaction amount, so THIS is where we need to know *exactly* how much bitcoin is available, precise down to the last single satoshi!

I think these values probably ought to be exact in the transaction history page too... and maybe in other places in the wallet as well: anywhere the number isn't just "displayed for a quick look" at it, but rather where it really ought to be *precise*.

P.S. I know we can click See It On The Blockchain to get actual full value of current address balance but that's really annoying to have to do: Copay should have this precise *full* value number at least in ONE place in the wallet itself.

isocolsky:
Fixes #3125 
Requires bitpay/bitcore-wallet-client#293

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "url": "https://github.com/bitpay/copay/issues"
   },
   "dependencies": {
-    "bitcore-wallet-client": "2.7.1",
+    "bitcore-wallet-client": "2.8.0",
     "coveralls": "^2.11.9",
     "express": "^4.11.2",
     "fs": "0.0.2",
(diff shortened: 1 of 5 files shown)


matiu:
Code ACK

cmgustavo:
LGTM

PR #4514, opened 2016-06-29
Fix spinner while create/import/join a wallet (iOS)

cmgustavo:


gabrielbazan7:
ACK - (iOS and Android)

PR #4517, opened 2016-06-29
black color easter egg

gabrielbazan7:


isocolsky:
ACK 😜

dabura667:
NACK!!!

Should be

Swipe up up down down left right left right menu button back button sleep button

PR #4521, opened 2016-06-30
allow creating transaction proposal without a private key

### LINKED ISSUE #4503: When creating a transaction proposal without a private key, copay hangs on "signing transaction"
While I was testing ledger support on the desktop version, it occurred to me that it'd be handy to use the mobile version to create transaction proposals. I figured I'll see if I can export the configuration to a mobile device and do that. Well, transaction proposal gets created and sent to other parties. However, copay hangs right after. UI is completely unusable and it gets stuck showing the info text "signing transaction".

Also another slight issue is that while the transaction proposal gets created, it doesn't get automatically relayed to the desktop copy of the same subwallet unless you restart it or switch back&forth between wallets, so it resyncs. Other parties in the multisig wallet do get them as normal though.

I didn't test this, but I suspect you'll see the same issues if you export a regular copay wallet without the private key and then import to another device.

gabrielbazan7:
fixes #4503


cmgustavo:
ACK!

PR #4551, opened 2016-07-06
enable button when all words are selected

### LINKED ISSUE #4550: Backup fails with 24 word wallets
Summary:
Externally creating a 24 word phrase breaks the backup logic so that it will always show the wallet as not backed up. Attempting the backup process always ends in a failed state.

To reproduce:

1. Create a valid 24 word BIP39 compatible phrase (outside of the CoPay). Example: https://github.com/dcpos/bip39
2. Create a new single signer wallet in CoPay.
3. Unhide advanced options
4. Select 'Specify Recovery Phrase' and input the 24 words
5. Inclusion of a password doesn't matter.
6. Create the wallet
7. When you view the wallet properties, it will show that the wallet has not yet been backed up. If you go through the backup procedure, the 'continue' button turns blue on the 12th correct word you select, then goes gray if you continue to select past the 13th word.
8. If a password was used in the creation of the wallet, you are prompted for the password. Typing in the correct password will result in 'Backup failed'.
9. Press 'continue' with only the first 12 correct words since that's the only way to proceed. This will result in "Backup Failed".

gabrielbazan7:
fixes #4550

### DIFF
--- a/src/js/controllers/backup.js
+++ b/src/js/controllers/backup.js
@@ -123,7 +123,7 @@ angular.module('copayApp.controllers').controller('backupController',
     };
 
     $scope.shouldContinue = function() {
-      if ($scope.customWords.length == 12)
+      if ($scope.customWords.length == $scope.shuffledMnemonicWords.length)
         $scope.selectComplete = true;
       else
         $scope.selectComplete = false;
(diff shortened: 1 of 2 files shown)


isocolsky:
Well done but I don't think this fixes #4550. A complete solution would assume an external backup phrase is already backed up.

cmgustavo:
ACK

PR #4565, opened 2016-07-09
Hide advanced options on export QR tab

### LINKED ISSUE #4561: [BUG] No private key option and warning text remain on QR export screen
Reproduce:

1. Open export wallet
2. Open advanced
3. check the no private key button
4. tap QR option on the upper right
5. notice that the no private key button and warning text remain, BUT the QR still contains the xprv private key.

-----------------------

Recommended method of fixing:

1. Allow for exporting via QR without private key (include xpub instead of xprv)
2. remove button and warning text tags when switich over to QR screen.

I vote for # 1 personally :-D

JDonadio:
Fixes #4561

### DIFF
--- a/public/views/export.html
+++ b/public/views/export.html
@@ -68,7 +68,7 @@
       </div>
     </div>
 
-    <div ng-show="showAdvanced">
+    <div ng-show="showAdvanced && !exportQR">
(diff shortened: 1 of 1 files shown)


cmgustavo:
ACK

PR #4573, opened 2016-07-11
prevent scroll when side bar is open

gabrielbazan7:


cmgustavo:
ACK! Tested on iPhone 6+

PR #4574, opened 2016-07-11
enable keyboard scroll when is show

### LINKED ISSUE #4569: No way to scroll on iPhone 5s when sending
Scroll is not working on iPhone 5s when sending BTC, so pressing the button to send is not possible.

A workaround is to press the "Go" key on the keyboard, but if one wants to set a description the "Go" key disappears (changes to "return"). So after setting a description one needs to go back to "amount" so the "Go" key reappears.

Example:
copay scroll bug

gabrielbazan7:
fixes #4569


JDonadio:
ACK

PR #4576, opened 2016-07-11
fix hide single address toggle

### LINKED ISSUE #4575: Cannot create Single Address Wallet when using a custom recovery phrase
To reproduce:

1. Create a new wallet
2. Enable 'Single Address Wallet' switch.
3. Select 'Specify Recovery Phrase'.
4. Single Address Wallet switch disappears.
5. Completing the rest of the fields and creating the wallet results in a multiple address wallet.

Is there something special about specifying a custom recovery phrase that is incompatible with single address wallets?

gabrielbazan7:
fixes #4575

### DIFF
--- a/public/views/create.html
+++ b/public/views/create.html
@@ -138,7 +138,7 @@
                 <span class="toggle-label">Testnet</span>
               </ion-toggle>
             </div>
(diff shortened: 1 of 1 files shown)


cmgustavo:
ACK

PR #4577, opened 2016-07-11
Wallet information: enables copy to clipboard on mobile

### LINKED ISSUE #4562: [REQUEST] (Single|Multiple) Tap(s?) to copy to clipboard for wallet xpub
For desktop, you can double click the xpub and even though it only shows a part of the xpub on the screen, right clicking and clicking Copy will copy the full xpub to the clipboard.

However, for iPhone, long tap, double tap, swipe, any action will not select the text.

To be honest, here are my thoughts:

1. It should not be immediately apparent that it is copyable, so the desktop method is good... but doesn't translate well into mobile.
2. If someone is set on copying the xpub, we should make it easy for them to do... otherwise they might have to resort to entering their recovery phrase to an online BIP39 tool to check their xpub... I'd rather recommend to such a person "oh, you just tap the xpub 3 times within 2 seconds, and it will copy to the clipboard" or something.
3. If we were to add something like tap twice to copy, it should also be available for desktop (click twice to copy etc)

Any thoughts?

cmgustavo:
Fixes #4562

### DIFF
--- a/public/views/preferencesInformation.html
+++ b/public/views/preferencesInformation.html
@@ -22,7 +22,7 @@
         </span>
       </li>
 
-      <li class="line-b p20 oh">
+      <li class="line-b p20 oh" ng-click="copyToClipboard(walletId)">
         <span translate>Wallet Id</span>
         <span class="right text-gray enable_text_select">
           {{walletId}}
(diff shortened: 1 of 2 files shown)


gabrielbazan7:
ACK

Issue #4567, opened 2016-07-10
Disable scrolling for amount input


CredibleFriends:
Accidentally happened to me today.  Not a big deal because there is a confirmation screen, but sometimes users click the confirmation hastily... perhaps it should be disabled. Bumping the mouse just a tad made it jump quickly from 0.002 btc to 5 btc

http://stackoverflow.com/questions/9712295/disable-scrolling-on-input-type-number

matiu:
good point,. thanks for reporting.

PR #4583, opened 2016-07-12
Disable mousewheel on numeric input fields

JDonadio:
Fixes https://github.com/bitpay/copay/issues/4567


cmgustavo:
ACK. Tested on iOS

PR #4590, opened 2016-07-14
WIP: Ref/build scripts

matiu:
refactor build script


matiu:
updated and ready for review.

Makefile was removed.

All targets are on `grunt` now.

matiu:
This need to be updated with https://github.com/bitpay/copay/pull/4656

PR #4629, opened 2016-07-28
Bug/nw quota

matiu:


cmgustavo:
ACK! Tested on Windows, Linux and OSX with existent wallets

braydonf:
Run this on Ubuntu and it looks like it works. I did notice that once migrated, older clients will think that there isn't any data, and will start with the normal welcome screen.

matiu:
updated. Test passing!

matiu:
Thanks @cmgustavo and @braydonf for testing!

Issue #4617, opened 2016-07-21
Implement PIN / Fingerprint App lock


matiu:
- Remove per-wallet fingerprint option
 - Keep per-wallet 'spending password' option (becuase that one actually encrypt the keys, for ultimate security)
 - Add a global fingerprint / pin option, something like:
 - Lock the app:
   - [ ] inmediatly
   - [ ] after 2 min
   - [ ] after 5 min
   - [ ] disabled
 -  [x] Use fingerprint (instead of PIN)
     
Fingerprint will be asked every time the app opens. On resume, it will be asked if x minutes passed since the last time the user enter the pin/fingerprint

Related:

https://github.com/bitpay/copay/issues/4508

https://github.com/bitpay/copay/issues/4555

https://github.com/bitpay/copay/issues/4589

https://github.com/bitpay/copay/issues/3061

JDonadio:
Lock the app after 1 minute of inactivity

bitjson:
https://github.com/bitpay/copay/issues/4555

Issue #4555, opened 2016-07-07
"Lock" app with PIN, fingerprint, or password


ivorpad:
I would like to have an option to hide the main balance or maybe lock Copay and open it with a PIN or Touch ID like Circle.app.

<img src="https://cloud.githubusercontent.com/assets/52452/16659212/27453f82-4437-11e6-9a86-239a5575765d.jpg" height="200">

JDonadio:
Try to tap and hold balance for 1-2 seconds.

ivorpad:
@JDonadio Yes I know but the balance should be hidden once I close the app.

JDonadio:
The balance remains hidden if you close/open or pause/resume the app.

gabegattis:
https://github.com/bitpay/copay/issues/3924 is a very similar issue. It was closed when https://github.com/bitpay/copay/pull/3872 was merged.

When I restart copay, the balance remains hidden for me. If it does not for you, that would be a bug. Do you have multiple wallets in Copay? The hidden balance feature is wallet specific, not a global setting.

JDonadio:
> The hidden balance feature is wallet specific, not a global setting.

That's very important ;)

ivorpad:
>When I restart copay, the balance remains hidden for me.

Agree, but anyone could grab your phone (let's say their checking in your photos with your permission) then out of curiosity they open Copay, 2-sec hold into Copay's balance and see it. It does not ask for a PIN, and for me this is a _security_ concern. 

I would like my balance to remain **private**. In BreadWallet if you want to check your balance you must enter a 4-PIN or Touch ID.

gabegattis:
Ok. I see what you are saying. I think the proper solution would be to have a way to lock the entire app. That way someone couldn't even open the app at all, much less see your balance, without the password, touchid, etc.

What do you think of that?

ivorpad:
@gabegattis That would pretty nice to have. 

I was googling around and found this: https://github.com/AlexDisler/ionic-lock-screen

dabura667:
tbh, if you leave your device unlocked, no amount of software PIN screen will protect your data.

If you don't have a strong PIN set on iPhone or a strong password or pattern set with Android disk encryption, you're vulnerable.

That said. Unless it brings up some sort of bug, adding the feature shouldn't hurt. Electrum's Thomas V said this kind of stuff lowers security way back and is why Electrum doesn't do it.

I like the way breadwallet does it.

[...]

PR #4630, opened 2016-07-28
WIP: Lock Copay App - Fingerprint

JDonadio:
* Moved: fingerprint option to global setting
* Prevents to show any wallet and balance if it's enabled
* Remove fingerprint control in backup, export, and send flows
* Add lock/unlock control on sidebar
* Error page

Closes https://github.com/bitpay/copay/issues/4555
Closes https://github.com/bitpay/copay/issues/4617


dabura667:
utACK

I guess I'll have to set a password now... I liked the way it was with only fingerprinting for sensitive actions.

Since we have no way to show the receive address without unlocking the app, someone could grab my phone from me while I am showing them my QR in the app and send all to their address.

I don't see why there is a need to remove it from all sensitive actions as well.

dabura667:
I like the way breadwallet does it:

1. Allow the user into the app with no pin or touchID.
2. Have a button for unlocking balance and history display. This asks you for auth.
3. ALSO ask for auth before every sensitive action.

dabura667:
This way you can access receive address

cmgustavo:
Needs rebase

JDonadio:
Updated

dmo580:
@dabura667 i don't see the point of only protecting sensitive actions. The app already had a send password option previously. One of the major complaints has been no ability to lock the app (zero user access until one enters PIN/fingerprint).

Also I don't see the app being locked in Android with fingerprint.

PR #4639, opened 2016-08-01
Prevent double tap (on mobile)

cmgustavo:
Problems (mobile): 

1.- Pressing the sign button, it isn't blocked by a few seconds. 
2.- Doing double tap on `Send Max` makes two call to the function.

Solution: 

Only adds a `$timeout` to block the button and the screen as soon as it was pressed in these cases:

* Sign a transaction
* Send Max


JDonadio:
LGTM. Tested on IOs and Android

Issue #4638, opened 2016-08-01
Scroll and recovery phrase problem when importing wallet


franamati:
Two problems when importing wallet in Copay v2.3.0 using an iPhone 5s:

* Scroll does not work: I can't find the button to import because I can't scroll.
* Imported recovery phrase is not seen after using the camera to complete the field.

Example:
![copay import](https://cloud.githubusercontent.com/assets/8605599/17297689/1dc3e436-57dd-11e6-846d-6d7e77c361ca.png)

JDonadio:
We could reproduce it. We are working on that and will be fixed in the next release. Thanks alot for the detailed info!

PR #4644, opened 2016-08-03
Shows scanned data - Import from QR

JDonadio:
Related in https://github.com/bitpay/copay/issues/4638

### DIFF
--- a/public/views/import.html
+++ b/public/views/import.html
@@ -50,19 +50,11 @@
             <span translate>Type the Recovery Phrase (usually 12 words)</span>:
           </label>
 
-          <div class="qr-scanner-input-import" ng-show="!dataFromQR">
+          <div class="qr-scanner-input-import">
             <qr-scanner on-scan="processWalletInfo(data)"></qr-scanner>
           </div>
 
-          <div class="lock-fromQR" ng-show="dataFromQR">
-            <div class="left m10l">
-              <i class="fi-lock color-greeni"></i>
-            </div>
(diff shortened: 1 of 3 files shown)


public/views/import.html
```diff
@@ -50,19 +50,11 @@
...
-          <div class="lock-fromQR" ng-show="dataFromQR">
-            <div class="left m10l">
-              <i class="fi-lock color-greeni"></i>
-            </div>
-            <div class="right icon-close-import" ng-click="dataFromQR = null">
-              <i class="icon-close-circle size-14"></i>
-            </div>
-          </div>
-          <textarea class="form-control m10t" ng-disabled="dataFromQR" name="words" ng-model="words" rows="2" autocapitalize="off" spellcheck="false"></textarea>
+          <textarea class="form-control m10t" name="words" ng-model="words" rows="3" autocapitalize="off" spellcheck="false"></textarea>
```
cmgustavo: I think that it should be disabled (read-only mode)

src/js/controllers/import.js
```diff
@@ -62,8 +62,7 @@ angular.module('copayApp.controllers').controller('importController',
       $scope.testnetEnabled = info.network == 'testnet' ? true : false;
 
       $timeout(function() {
-        $scope.words = null;
-        $scope.dataFromQR = info.data;
+        $scope.words = $scope.dataFromQR = info.data;
```
cmgustavo: Variable `$scope.dataFromQR` is not used.

cmgustavo:
ACK

Issue #4557, opened 2016-07-08
Copay hanging when canceling fingerprint scanner


isocolsky:
To reproduce: 
* Set up fingerprint protection.
* Sign a payment proposal but hit "Cancel" on fingerprint popup.
* "Signing transaction" spinner stays on forever...

matiu:
Could not reproduce it this one (on Galaxy S7) not in 1-1 wallet nor in 2-2 wallets.

isocolsky:
Sorry, it seems to only happen on multi-sig. I created a payment proposal from my PC and tried to apply the last signature on my phone. Canceling the fingerprint verification lead to this screen:

![image](https://cloud.githubusercontent.com/assets/2073468/16732119/ef5119a2-4751-11e6-8a8b-5d8c13e4d918.png)

PR #4648, opened 2016-08-04
Fix displaying errors in transaction proposal details

JDonadio:
Fixes https://github.com/bitpay/copay/issues/4557

### DIFF
--- a/src/js/controllers/modals/txpDetails.js
+++ b/src/js/controllers/modals/txpDetails.js
@@ -1,6 +1,6 @@
 'use strict';
 
-angular.module('copayApp.controllers').controller('txpDetailsController', function($scope, $rootScope, $timeout, $interval, $ionicModal, platformInfo, txStatus, $ionicScrollDelegate, txFormatService, fingerprintService, bwcError, gettextCatalog, lodash, profileService, walletService) {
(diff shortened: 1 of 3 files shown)


src/js/controllers/modals/txpDetails.js
```diff
@@ -72,18 +70,30 @@ angular.module('copayApp.controllers').controller('txpDetailsController', functi
...
       walletService.rejectTx(fc, txp, function(err, txpr) {
-        $scope.loading = null;
+        ongoingProcess.set('rejectTx', false);
 
         if (err) {
           $scope.$emit('UpdateTx');
-          $scope.error = bwcError.msg(err, gettextCatalog.getString('Could not reject payment'));
-          $scope.$digest();
+          setError(err, gettextCatalog.getString('Could not reject payment'));
         } else {
```
isocolsky: do not use `else`, instead do ```if (err) return setError(....```

src/js/services/onGoingProcess.js
```diff
@@ -14,6 +14,8 @@ angular.module('copayApp.services').factory('ongoingProcess', function($log, $ti
     'sendingTx': gettext('Sending transaction'),
     'signingTx': gettext('Signing transaction'),
     'broadcastingTx': gettext('Broadcasting transaction'),
+    'rejectTx': gettext('Rejecting payment proposal'),
+    'removeTx': gettext('Deleting Payment proposal'),
```
isocolsky: use lowercase `p`

JDonadio: Good catch!

JDonadio:
Updated. Thanks @isocolsky !

isocolsky:
ACK

PR #4653, opened 2016-08-05
Notify when click on clickable element

### LINKED ISSUE #4637: [Desktop] Notify when click on clickable element (eg. bitcoin address)
When clicked on a bitcoin address in the popup, the user don't know that it was copied to the clipboard. We need to add a notification or something to show that the address is copied to clipboard

gabrielbazan7:
fixes #4637


JDonadio:
ACK

PR #4667, opened 2016-08-09
fix scroll on ios devices

gabrielbazan7:


gabrielbazan7:
Fix in next versions.

PR #4700, opened 2016-08-18
Move hardcoded insight button url to config

JohnEz:
The main purpose of this pull request is to add an advanced option that lets users change where the "See it on the blockchain" button links to. This will allow users to choose their blockchain explorer like they can currently choose their wallet service. At present, it's hard coded in the button itself to either link to production or test. 

Originally it was checking if the short network name started with the word "test" and if so, added "test-" to the start of the url. I have removed this feature as you can now enter any url you want to. However, if you want to retain the check for this, just let me know.

I noticed that the BWS url doesn't have a trailing /, so at first I wrote the transaction url to be the same,  where it just added it for them. The problem with this is that, potentially, different blockchain explorers may use the txid differently such as part of a query or a hash. I have opted to use a token instead so that the id can be placed anywhere in the url string and as many times as the user needs.

### DIFF
--- a/public/views/modals/tx-details.html
+++ b/public/views/modals/tx-details.html
@@ -149,8 +149,7 @@
 
       <div ng-show="btx.txid" class="tx-details-blockchain">
         <div class="text-center m20t">
-          <button class="button outline round dark-gray tiny" ng-click="$root.openExternalLink('https://' +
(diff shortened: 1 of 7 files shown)


matiu:
nice! thanks the the contribution

matiu:
This PR was outdated by a mayor refactor. Please resubmit. Thanks!

PR #4720, opened 2016-08-29
Use more explicit logic in fee level check

JohnEz:
Bitcore-Wallet-Service allows for a minimum fee of `0`, however if this is the case it will currently cause Copay errors. These problems are caused by the `!feeLevelBalue.feePerKb` checks that return true on a case of `0`.

This PR only solves half of the problem as there is a similar check in Bitcore-Wallet-Client which I have also created a [pull request](https://github.com/bitpay/bitcore-wallet-client/pull/320) for.

### DIFF
--- a/src/js/services/feeService.js
+++ b/src/js/services/feeService.js
@@ -30,7 +30,7 @@ angular.module('copayApp.services').factory('feeService', function($log, $stateP
       var feeLevelValue = lodash.find(levels, {
         level: feeLevel
       });
-      if (!feeLevelValue || !feeLevelValue.feePerKB)
+      if (!feeLevelValue || feeLevelValue.feePerKB == null)
         return cb({
           message: 'Could not get dynamic fee for level: ' + feeLevel
         });


matiu:
Allowing zero fee in a general purpose wallet will generate problem of transactions never confirming.

chrisprice:
N.B. This is to fix a logical flaw rather than attempt to change fee levels. Should zero tx fees ever be allowed, the code in its current form would cause an unexpected failure.

bitjson:
Hey @chrisprice – thanks for the pull request! It seems like it's probably a good idea to use this more explicit logic. Would you mind rebasing your pull request so it can be merged?

chrisprice:
Great. It's not actually my PR, I'll have a word with @JohnEz when he's back from holiday.

src/js/services/feeService.js
```diff
@@ -28,7 +28,7 @@ angular.module('copayApp.services').factory('feeService', function($log, bwcServ
       var feeLevelValue = lodash.find(levels, {
         level: feeLevel
       });
-      if (!feeLevelValue || !feeLevelValue.feePerKB)
+      if (!feeLevelValue || feeLevelValue.feePerKB == null)
```
JohnEz: Normally i would agree, using `===` over `==` is usually safer however, my reason for doing it like this here is that `=== null` won't catch `undefined` .

bitjson:
Thanks!

PR #4801, opened 2016-10-15
Locking amount on paypro only if sum of outputs great than zero

homeopatchy:
Allows user-specified amounts on paypro as per  [bip70](https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki#paymentdetailspaymentrequest):

> 
**outputs**: one or more outputs where Bitcoins are to be sent. **If the sum of outputs.amount is zero, the customer will be asked how much to pay**, and the bitcoin client may choose any or all of the Outputs (if there are more than one) for payment. If the sum of outputs.amount is non-zero, then the customer will be asked to pay the sum, and the payment shall be split among the Outputs with non-zero amounts (if there are more than one; Outputs with zero amounts shall be ignored).

### DIFF
--- a/src/js/controllers/walletHome.js
+++ b/src/js/controllers/walletHome.js
@@ -616,7 +616,7 @@ angular.module('copayApp.controllers').controller('walletHomeController', functi
       this.lockAddress = true;
     }
 
-    if (amount) {
+    if (amount > 0) {
       form.amount.$setViewValue("" + amount);
       form.amount.$isValid = true;
       form.amount.$render();


homeopatchy:
Not particularly happy of the way I fix it (simply using coercion), the issue comes from using `toFixed()` here, as a string it would always validate to true inside `setForm()`:

https://github.com/homeopatchy/copay/blob/02164c108e90b7f3432c6a38026548f0095bd87a/src/js/controllers/walletHome.js#L728

matiu:
Thanks for your PR. We want to merge this in, but it was outdated by a mayor refactor. Could you  resubmit it?

PR #4898, opened 2016-10-26
remove to show backup needed modal automatically when clicking on receive view

gabrielbazan7:
I think we should keep it simple. There are so much issues with this feature. 

-  Clicking on Receive tab and immediately on other shows the modal, but sometimes no.  Sometimes the 'Wallet/Changed' event fires when we are out that view.
-  Sometimes the 'Wallet/Changed' is not fired and the modal do not show at all.
- When the modal is not shown the top label is also not shown.

Maybe we could create a an issue to add this feature again later.


cmgustavo:
ACK

PR #4919, opened 2016-10-26
feat(resources): automatically generate Windows ICO file

bitjson:
Also corrects the location/adds several other resources


matiu:
nice!

PR #4939, opened 2016-10-28
Create Dockerfile

amitkumarj441:
Initial commit to Dockerfile for #4940

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,8 +3,8 @@ FROM dockerfile/nodejs-bower-grunt
 MAINTAINER Amit Kumar Jaiswal "amitkumarj441@gmail.com"
 
 RUN apt-get install -y build-essential
-RUN mkdir /opt/copay
-WORKDIR /opt/copay
+RUN mkdir /bitpay/copay
+WORKDIR /bitpay/copay
 RUN \
     curl -s -L https://github.com/justcoin/copay/tarball/master | tar zx -C /opt/copay/ --strip-components=1 && \
     npm install && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM dockerfile/nodejs-bower-grunt
+
+MAINTAINER Amit Kumar Jaiswal "amitkumarj441@gmail.com"
+
+RUN apt-get install -y build-essential
+RUN mkdir /opt/copay
+WORKDIR /opt/copay
+RUN \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM dockerfile/nodejs-bower-grunt
+
+MAINTAINER Amit Kumar Jaiswal "amitkumarj441@gmail.com"
+
+RUN apt-get install -y build-essential
+RUN mkdir /opt/copay
+WORKDIR /opt/copay
+RUN \
+    curl -s -L https://github.com/justcoin/copay/tarball/master | tar zx -C /opt/copay/ --strip-components=1 && \
```
matiu: use the official repo

amitkumarj441: @matiu I'll change it right away! Thanks for letting me know about this info!

amitkumarj441:
@matiu Please review my PR!

bitjson:
Not sure if I understand the goal here – is this for setting up to work on development of Copay?

`npm start` is a very unsafe way to use this project for real money. That opens the app in the system default browser, and could be compromised by nearly any browser extension, as well as any "XSS vulnerabilities" which may or may not exist. This project is not – and should not be used as – a web app. (If you're storing real money on a website, you're gonna have a bad time...)

I think we want contributors to Copay using the very simple `npm install && npm start` method – it's already very stable and repeatable (not to mention the `start:ios`, `start:android`, etc scripts). I don't see a lot of value in complicating that with Docker.

amitkumarj441:
@bitjson This docker image is for the development of copay!

bitjson:
@amitkumarj441 – definitely appreciate your interest in contributing to Copay.

> I think we want contributors to Copay using the very simple `npm install && npm start` method – it's already very stable and repeatable (not to mention the `start:ios`, `start:android`, etc scripts). I don't see a lot of value in complicating that with Docker.

Is there a specific use case for why a developer shouldn't clone this repo normally to contribute to Copay? In order to share their changes, they'll still need to send a pull request. This just really doesn't seem like a place for Docker.

bitjson:
Closing for now, but please let me know if this should remain open.

amitkumarj441:
@bitjson You're right! But why did you closed this PR, even I tested this dockerfile and it's working fine, even this can help other to extend this dockerfile according to their use.

amitkumarj441:
@bitjson Got it! Thanks for the suggestion!

Issue #4881, opened 2016-10-25
Show the "unconfirmed balance" everywhere


JDonadio:
_From @bitjson on October 25, 2016 3:44_

The Home tab and wallet views still display the "available balance" as the primary visible number. This should be the "unconfirmed balance".

This requires quite a bit of development work still (since Copay focuses on the "available balance"):
- implement the "Confirming" transaction items from the mocks (https://github.com/bitpay/bitpay-wallet/issues/539) – this will help to make it clearer why there are some funds which can't be spent
- Provide helpful error messages on the amount-entry view and the tx confirm view explaining when the "available balance" is lower than the required amount
- Consider enabling the "Use Unconfirmed Funds" by default, and instead of preventing it, issue a modal dialog when the user tries to spend uncofirmed funds (asking the to confirm they would like to send the transaction, even though it might be delayed)

A relevant conversation:
https://app.intercom.io/a/apps/2a3a0d58dd7819f5938cf908db897ecebb22df70/inbox/a/james@bitpay.com/conversations/6468816342

_Copied from original issue: bitpay/bitpay-wallet#668_

bitjson:
This is working as it was originally designed, but I'm marking it as a bug, since it's unexpected and potentially scary.

PR #4944, opened 2016-10-28
Make available/confirm balance concept less scary

bitjson:
A small fix for #4881, we should re-examine in another pass.

### DIFF
--- a/src/js/services/walletService.js
+++ b/src/js/services/walletService.js
@@ -220,9 +220,9 @@ angular.module('copayApp.services').factory('walletService', function($log, $tim
       cache.totalBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat) + ' ' + cache.unitName;
       cache.lockedBalanceStr = txFormatService.formatAmount(cache.lockedBalanceSat) + ' ' + cache.unitName;
       cache.availableBalanceStr = txFormatService.formatAmount(cache.availableBalanceSat) + ' ' + cache.unitName;
-      cache.pendingBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat + cache.pendingAmount) + ' ' + cache.unitName;
+      cache.pendingBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat + (cache.pendingAmount === null? 0 : cache.pendingAmount)) + ' ' + cache.unitName;
 
-      if (cache.pendingAmount) {
+      if (cache.pendingAmount !== null) {
         cache.pendingAmountStr = txFormatService.formatAmount(cache.pendingAmount) + ' ' + cache.unitName;
       } else {
         cache.pendingAmountStr = null;
--- a/src/js/services/walletService.js
+++ b/src/js/services/walletService.js
@@ -220,6 +220,7 @@ angular.module('copayApp.services').factory('walletService', function($log, $tim
       cache.totalBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat) + ' ' + cache.unitName;
       cache.lockedBalanceStr = txFormatService.formatAmount(cache.lockedBalanceSat) + ' ' + cache.unitName;
       cache.availableBalanceStr = txFormatService.formatAmount(cache.availableBalanceSat) + ' ' + cache.unitName;
+      cache.pendingBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat + cache.pendingAmount) + ' ' + cache.unitName;
 
       if (cache.pendingAmount) {
         cache.pendingAmountStr = txFormatService.formatAmount(cache.pendingAmount) + ' ' + cache.unitName;
--- a/src/js/services/walletService.js
+++ b/src/js/services/walletService.js
@@ -222,7 +222,7 @@ angular.module('copayApp.services').factory('walletService', function($log, $tim
       cache.availableBalanceStr = txFormatService.formatAmount(cache.availableBalanceSat) + ' ' + cache.unitName;
       cache.pendingBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat + (cache.pendingAmount === null? 0 : cache.pendingAmount)) + ' ' + cache.unitName;
 
-      if (cache.pendingAmount !== null) {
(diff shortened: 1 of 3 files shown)


bitjson:
This solves the immediate problem of the app making it seem like larger-than-intended payments are being made.

# Flow:


### Initial balances
screen shot 2016-10-28 at 11 40 31 am

### 100,000 bits sent from one wallet to the other
screen shot 2016-10-28 at 11 41 08 am

### New balances on home tab
screen shot 2016-10-28 at 11 42 45 am

### user taps testnet wallet
screen shot 2016-10-28 at 11 43 12 am

### user taps Savings wallet
screen shot 2016-10-28 at 11 43 01 am

I think there are a lot of design improvements that could be made to help (particularly new bitcoin users) understand the concept of "confirming", though this fixes the immediate issue of frightening users in certain situations.

Note: I intentionally did not make any changes to the Receive view – I assume that if the user is "holding up a QR code", they're not simultaneously trying to make a transaction from one of their wallets. If someone makes a payment to the QR code, the balance won't change until it's confirmed (to avoid misleading uninformed users into believing they have received the payment already).

screen shot 2016-10-28 at 11 41 43 am
screen shot 2016-10-28 at 11 41 36 am

So while the transaction above was going on, this view would still only show the confirmed balance, which seems like a balance drop. The user can flip back to their wallet to see more info though:
screen shot 2016-10-28 at 11 43 01 am

### After confirmation, the wallets views return to normal:
screen shot 2016-10-28 at 12 01 59 pm
screen shot 2016-10-28 at 12 02 09 pm


I think this solves the immediate issue, but we should review this fundamentally. Hopefully we can eliminate the imprecise "Use Unconfirmed Funds" Advanced setting, and always intelligently differentiating between unconfirmed funds under the user's control, and unconfirmed funds that the user can't yet rely on. (@isocolsky has solid ideas on this)

src/js/services/walletService.js
```diff
@@ -220,6 +220,7 @@ angular.module('copayApp.services').factory('walletService', function($log, $tim
       cache.totalBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat) + ' ' + cache.unitName;
       cache.lockedBalanceStr = txFormatService.formatAmount(cache.lockedBalanceSat) + ' ' + cache.unitName;
       cache.availableBalanceStr = txFormatService.formatAmount(cache.availableBalanceSat) + ' ' + cache.unitName;
+      cache.pendingBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat + cache.pendingAmount) + ' ' + cache.unitName;
```
bitjson: Good catch, thanks! Fixed in the latest commit.

src/js/services/walletService.js
```diff
@@ -220,8 +220,9 @@ angular.module('copayApp.services').factory('walletService', function($log, $tim
       cache.totalBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat) + ' ' + cache.unitName;
       cache.lockedBalanceStr = txFormatService.formatAmount(cache.lockedBalanceSat) + ' ' + cache.unitName;
       cache.availableBalanceStr = txFormatService.formatAmount(cache.availableBalanceSat) + ' ' + cache.unitName;
+      cache.pendingBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat + (cache.pendingAmount === null? 0 : cache.pendingAmount)) + ' ' + cache.unitName;
 
-      if (cache.pendingAmount) {
+      if (cache.pendingAmount !== null) {
```
cmgustavo: Warning: `cache.pendingAmount` could be 0 (zero) from ```balance.totalAmount - balance.totalConfirmedAmount;```

cmgustavo:
ACK

src/js/services/walletService.js
```diff
@@ -220,8 +220,9 @@ angular.module('copayApp.services').factory('walletService', function($log, $tim
       cache.totalBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat) + ' ' + cache.unitName;
       cache.lockedBalanceStr = txFormatService.formatAmount(cache.lockedBalanceSat) + ' ' + cache.unitName;
       cache.availableBalanceStr = txFormatService.formatAmount(cache.availableBalanceSat) + ' ' + cache.unitName;
+      cache.pendingBalanceStr = txFormatService.formatAmount(cache.totalBalanceSat + (cache.pendingAmount === null? 0 : cache.pendingAmount)) + ' ' + cache.unitName;
 
-      if (cache.pendingAmount) {
+      if (cache.pendingAmount !== null && cache.pendingAmount !== 0) {
```
bitjson: I was thinking it would be a good idea to make the logic a little more explicit – reading through the code doesn't make it very obvious. (Maybe we can leave it explicit until we have unit tests for this view?)

www/views/walletDetails.html
```diff
@@ -37,10 +37,11 @@
...
 
           <div ng-click='updateAll(true)' ng-show="!updateStatusError && wallet.walletScanStatus != 'error' && !wallet.balanceHidden" on-hold="hideToggle()">
-            <strong class="size-36">{{status.totalBalanceStr}}</strong>
-            <div class="size-14 amount-alternative" ng-if="status.totalBalanceAlternative">{{status.totalBalanceAlternative}} {{status.alternativeIsoCode}}</div>
-            <div class="size-14" ng-if="status.pendingAmount">
-              <span translate>Pending Confirmation</span>: {{status.pendingAmountStr}}
+            <strong ng-if="!status.pendingAmount" class="size-36">{{status.totalBalanceStr}}</strong>
+            <div ng-if="!status.pendingAmount" class="size-14 amount-alternative" ng-if="status.totalBalanceAlternative">{{status.totalBalanceAlternative}} {{status.alternativeIsoCode}}</div>
+            <div class="size-20" ng-if="status.pendingAmount">
+              <div style="margin-bottom:.5rem"><span translate>Available</span>: {{status.totalBalanceStr}}</div>
```
isocolsky: Shouldn't `status. totalBalanceAlternative` also be shown in case `status.pendingAmount` is `true`?

PR #4957, opened 2016-10-30
Allow user-defined amounts when paypro request amount is zero

homeopatchy:
Same as #4801 but updated for the last version. Routes through the amount controller if required before going into confirm.


bitjson:
Thanks for the PR @homeopatchy! @matiu do you want to review?

homeopatchy:
Please note https://github.com/bitpay/bitcore-wallet-client/pull/337 is required to test user-defined amounts, as it is now the tx proposal won't succeed as the paypro check in bwc marks it as invalid.

bitjson:
Maybe @cmgustavo can review this?

PR #4959, opened 2016-10-31
Add webkit prefixed transforms (android 4)

homeopatchy:
This adds missing webkit prefixes for the css property `transform`. These are required by Android 4.x


bitjson:
ACK – this is great, thanks!

PR #4963, opened 2016-11-01
Feedback feature

gabrielbazan7:
Needs design.


JDonadio:
Code was cleaned up. @bitjson can you take a look ? I think this is ready! 🎉

bitjson:
Could we get some info on how best to manually test this? How can I modify my local storage to retry, etc? Looking great so far!

On first review, the 5/5 feedback flow dropped me right into the text entry screen, when it should have taken me to the Thank you / rate view. I still need to step through the other flows.

JDonadio:
@bitjson try modifying this:

copay_-_copay_bitcoin_wallet

gabrielbazan7:
@bitjson sorry, just fixed.

PR #4966, opened 2016-11-01
URI handler support for macOS

cmgustavo:
* Fix slide to pay for desktop after click on registered uri
* Adds uri handler for macOS (Info.plist)
* Fix hidden back button

Note: could fix some cases when the back button disappears.

### DIFF
--- a/www/views/confirm.html
+++ b/www/views/confirm.html
@@ -7,7 +7,7 @@
     </ion-nav-back-button>
   </ion-nav-bar>
 
-  <ion-content ng-class="{'slide-to-pay': !hasClick && !insuffientFunds && !noMatchingWallet}">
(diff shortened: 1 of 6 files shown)


www/views/confirm.html
```diff
@@ -7,7 +7,7 @@
     </ion-nav-back-button>
   </ion-nav-bar>
 
-  <ion-content ng-class="{'slide-to-pay': !hasClick && !insuffientFunds && !noMatchingWallet}">
```
bitjson: I think we want to be determining this functionality based on device capabilities, rather than the specific platform. Maybe @msalcala11 should look at this too?

cmgustavo: Agree, but it 's not working at first launch (after clicking on a bitcoin: URI). We should tweak that function in another PR.

matiu: @bitjson this is a bug fix, it is currently broken.

PR #4969, opened 2016-11-02
Adds "Confirmations" to tx-details

cmgustavo:


JDonadio:
This field should be shown if action != 'received', right?

cmgustavo:
No, I think it 's useful both in `sent` as in `receive`

JDonadio:
ACK

PR #5015, opened 2016-11-11
Implement multiple cards.

ajp8164:
Manages multiple cards per account and multiple cards across multiple bitpay accounts.
Implements a simple framework for upgrading storage data (upgrades attached debit card storage).


cmgustavo:
ACK! Tested with an *old* card. Added 2 additional accounts successfully.

PR #5043, opened 2016-11-15
Fix android pull to refresh icon

msalcala11:
A quick fix to ensure the pull to refresh icon is still visible on android after recent changes.


JDonadio:
ACK

PR #5049, opened 2016-11-15
After send feedback from settings-tab, go back and clear history

cmgustavo:
* Show alert, click ok and return to tab-settings automatically


bitjson:
ACK

PR #5059, opened 2016-11-16
Status bar routing

msalcala11:
Hides the status bar in the feedback flow via a flag on the state declaration in routes.js


cmgustavo:
ACK! Tested on iOS

PR #5066, opened 2016-11-16
fix(feedback): improve design of feedback flow, correct several logic issues

bitjson:


www/views/feedback/complete.html
```diff
@@ -7,81 +7,52 @@
...
-        </div>
-        <div class="subtitle" ng-if="score <= 3 || !isCordova">
-          <span translate>If you have additional feedback, please let us know by tapping the "Send feedback" option in the Settings tab.</span>
-        </div>
-        <div ng-if="score <= 3 || !isCordova">
-          <div class="text-center">
-            <i class="icon icon-svg">
-              <img src="img/illustration-send-feedback.png"/>
-            </i>
+    <a class="close-button" ng-click="close()"><i class="icon ion-ios-close-empty close-home-tip"></i></a>
```
cmgustavo: This line is duplicated. The back button is working as expected.

Anyway, we could remove the `nav-bar`

cmgustavo: 5 stars, click on "Not now", the "Share BitPay" page is broken (it has 2 close button (X) and the `nav-bar`).

bitjson:
I think this is almost ready – final two issues:
- `complete.html` – `A member of the team will review your feedback as soon as possible.` is not displaying when the view shows for scores 1-3
- `Settings` -> `Share BitPay` – the back button is not displaying

Issue #5020, opened 2016-11-13
Receive View Enhancements, new Wallet Address view


bitjson:
Current high-level issues with the Receive view:

- First-time users tend to not understand that wallets contain multiple addresses. At best, they think they need to create a new wallet to get a new address. At worst, they think the address change is a critical bug (in which payments to old addresses might be lost).
- `Next Address` doesn't mean anything to even many experienced users. It's also a very "niche" feature. It's very unlikely to be necessary in normal operation.
- There's no obvious way for advanced users to "dig into the wallet". This is actually available in `Settings` -> the wallet's settings -> `More Options` -> `Wallet Information`. 

Minor issues (merging from #4891):
- Font sizes are generally too small – we should be using device-standard sizes
- On first-impression, It's not immediately obvious which wallet I'm looking at

Mocks are below for a new section. The user can tap the bitcoin address in the receive view to be taken forward to the "Wallet Addresses" view. We also provide some better help text, where we try to be thorough (and provide info even "experience bitcoiners" might appreciate).

We also move the `Next Address` functionality into the Wallet Addresses view (and make it a little more useful).

<table><tr>
<td><img src="https://cloud.githubusercontent.com/assets/904007/20247772/734a7542-a9a3-11e6-9f2d-c7610b5ecc5b.png" width="200"></td>

<td><img src="https://cloud.githubusercontent.com/assets/904007/20247778/8e0c39ce-a9a3-11e6-8a3f-70e119fcbf0b.png" width="200"></td>

<td><img src="https://cloud.githubusercontent.com/assets/904007/20247799/ec449e14-a9a3-11e6-8ea1-3592a2181f22.png" width="200"></td>
</tr><tr>

<td><img src="https://cloud.githubusercontent.com/assets/904007/20247796/dc14008e-a9a3-11e6-9618-9ca96e0a9b74.png" height="500"></td>

<td><img src="https://cloud.githubusercontent.com/assets/904007/20247804/ffefccf4-a9a3-11e6-9691-c297d71cee4c.png" height="500"></td>

<td><img src="https://cloud.githubusercontent.com/assets/904007/20247808/0a360a3e-a9a4-11e6-95a4-3733b33c89d6.png" height="500"></td>

<tr/><tr>
<td><img src="https://cloud.githubusercontent.com/assets/904007/20247812/49f09478-a9a4-11e6-865f-ca8e466d85ab.png" height="300"></td>
</tr></table>

bitjson:
[...]

PR #5070, opened 2016-11-17
Feat/address view

JDonadio:
Fixes https://github.com/bitpay/copay/issues/5020

* Addresses limit = 5 (for both lists)
* View all addresses **button** displayed if *any list length > 5*
* View all addresses > list of **all addresses**
* Remove duplicated information in *preferences > wallet > adnvanced > wallet information*
* Add popover menu with options (below)
* Migrate **scan/send** addresses functionality to addresses views

### DIFF
--- a/src/js/controllers/addresses.js
+++ b/src/js/controllers/addresses.js
@@ -1,6 +1,6 @@
 'use strict';
 
-angular.module('copayApp.controllers').controller('addressesController', function($scope, $stateParams, $timeout, $ionicScrollDelegate, configService, popupService, gettextCatalog, ongoingProcess, lodash, profileService, walletService) {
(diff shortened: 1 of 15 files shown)


bitjson:
@JDonadio for the icon, you can use `www/img/tab-icons/ico-receive-selected.svg` in the same way `addressbook.html` does:

```
      <i class="icon zero-state-icon">
        <img src="img/tab-icons/ico-receive-selected.svg"/>
      </i>
```

gabrielbazan7:
On "Wallet Information" view, "Latest Wallet Addresses" option should be remove ?

screenshot_20161117-155522

What about "Scan address for fund" and "Send addresses by email" buttons

screenshot_20161117-155515

JDonadio:
Fixed icon @bitjson.

Good point @gabrielbazan7, @bitjson what do you think about that? seems like we have duplicated info in both views

cmgustavo:
ACK. Good job!

PR #5080, opened 2016-11-17
Detect and convert localStorage values to string.

ajp8164:
Change guarantees that localStorage always writes a string.


cmgustavo:
Good catch! ACK

PR #5100, opened 2016-11-22
initialize 5 tab layout

msalcala11:


ionic2/src/pages/home/home.scss
```diff
@@ -1,3 +1,132 @@
 page-home {
+    //@extend .deflash-blue;
```
bitjson: Can we do this as a move rather than an add? (We can delete this file from the old `sass` dir.) Would help to be able to track where things are coming from.

ionic2/src/theme/custom-variables.scss
```diff
@@ -0,0 +1,56 @@
+/* constants */
```
bitjson: Same here, I think it would be good to delete the old one as we bring this one in to this structure.

bitjson:
(Actually, I'll submit a PR that only moves the files, then we can merge this)

bitjson:
https://github.com/bitpay/copay/pull/5101

bitjson:
🚀

marvelou:
I don't want to receive emails from now

PR #5127, opened 2016-12-01
Update translations

JDonadio:


dabura667:
@JDonadio If all is good, we should merge this with master, then update Crowdin.

Then wait for a bit for new translations, and make a second pull request with all the new translated po files.

If you don't know how to do this, I will update Crowdin once this is merged.

Thanks.

Just to check, is this template for the new Bitpay app or for the old Copay app? Also, how would we separate the two pot files for the two versions on Crowdin. (these are some questions I have)

JDonadio:
@dabura667 there is some fix remaining here, I'll try to clean all of them today and then merge this with master as you say. Would be awesome your help here ;)

> Just to check, is this template for the new Bitpay app or for the old Copay app?
Bitpay app, but will be the same for both in a future.

>Also, how would we separate the two pot files for the two versions on Crowdin
Good question, @bitjson, @cmgustavo what do you think?

gabrielbazan7:
ACK

dabura667:
@JDonadio

Re: separating the files

How is the repo handling the Bitpay and Copay app diff? Will Master branch be Bitpay and Copay branch be Copay or something?

If we can decide something like that, then we could just change the current template.pot on crowdin to template_copay.pot or something, and add the new template_bitpay.pot as a new file (Translation Memory will auto-fill 100% matches from other files in the project) then I could change my script to check the current local branch and unpack only the translations for the correct pot file for the current branch in git.

However, this will be a bit of work, and I am very busy with taking care of my wife and my newborn son.

For now, my script will just unpack what it is specifically told to unpack, so you could leave everything as is and just add a new file like template_bitpay.pot and my script will ignore it. You can then manually package it by downloading the zip file from the crowdin page after "building the project" in the control panel.

PR #5139, opened 2016-12-02
Removes *Recent Transaction Card* from experimental features

cmgustavo:


bitjson:
NACK – design-wise, we don't think this is the direction we want to go currently. It somewhat limits our options for extending the view with other kinds of data/activity. It's ok to leave experimental, but I'd rather not enable it now just to potentially disable/replace it in a few weeks.

Issue #4943, opened 2016-10-28
Hide "Import from Hardware Wallet" on non-Chrome apps 


jameswalpole:
The BitPay app/Copay UIs  show the "Import from Hardware Wallet" option regardless of platform. Hardware wallet imports are only possible in the Chrome app. We need to hide "Import from Hardware Wallet" to avoid confusion.

Issue #4909, opened 2016-10-26
Development needed before Chrome Apps are deprecated


bitjson:
Google is discontinuing support for Chrome Apps: https://blog.chromium.org/2016/08/from-chrome-apps-to-web.html

> We will be removing support for packaged and hosted apps from Chrome on Windows, Mac, and Linux over the next two years. 

The Chrome App is the primary platform for the hardware wallet integrations – we/hardware wallet providers need to figure out how best to make those integrations available within the native desktop apps.

ruimarinho:
@bitjson does this affect the nw.js-based Copay binaries as well?

bitjson:
@ruimarinho no, this means Copay will **only** be available for desktop via the native desktop apps for each platform (currently nw.js).

JDonadio:
Please check: https://blog.bitpay.com/copay-chrome-end/

We have no plans to support hardware wallets on desktop platforms for now. If we have news about that we'll reopen this.

ruimarinho:
Got it. I will keep subscribed to this issue. Thanks!

nathankaiser:
@bitjson Question for my understanding: does this mean that the Chrome App for ChromeOS will continue to exist? 

(in other words, does the deprecation "only" affect the Chrome App on Windows, etc.?)

Thanks!

PR #5153, opened 2016-12-06
Disable hardware wallets

JDonadio:
Related https://github.com/bitpay/copay/issues/4909
Closes https://github.com/bitpay/copay/issues/4943

Disable hardware options for **BitPay distribution**:

* Create single/shared wallet
* Join wallet
* Import wallet


cmgustavo:
ACK

Issue #5194, opened 2016-12-09
Wallet balance UX suggestion


ajp8164:
Complaint:

"still, nowhere on this screen does it tell me what is available to spend right now.  I have to manually subtract "Available" minus "Locked" to get that number. My settings are to Not spend unconfirmed funds.  So exactly what is available for me? there are 3 numbers on the screen related to my balance and none of them are what I need."

![screen shot 2016-12-09 at 9 11 26 am](https://cloud.githubusercontent.com/assets/1063765/21052443/1c736382-bdf3-11e6-9485-9e3360f3e3e3.png)

Possible solution:

“Available” should be “what i can spend right now”; everything else should be totaled into “Pending”.  If I click on “Pending” (make it look like a button) then I get a modal that shows me a break-down “Locked”, “Confirming”, etc..  with a bit of explantation about each category.

ajp8164:
See https://github.com/bitpay/copay/pull/5196

PR #5196, opened 2016-12-09
Initial implementation for balance modal.

ajp8164:
When the total wallet balance is not equal to the available (spendable) balance then a button appears in the wallet home header.  The button text is the available (spendable) balance.

The total wallet balance is not equal to the available (spendable) balance when there are unconfirmed or locked funds.

Clicking the button presents four amounts and some explanation for each:

- Total balance
- Available balance
- Confirming amount
- Locked amount

Testing/fixing - this PR needs some UX work wrt. the collapsing wallet header and the size of the header when the button is displayed vs. not.

See https://github.com/bitpay/copay/issues/5194.


src/js/controllers/modals/walletBalance.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+angular.module('copayApp.controllers').controller('walletBalanceController', function($scope) {
+
+  $scope.close = function() {
+    $scope.walletBalanceModal.hide();
```
JDonadio: Please, if the controller will not have more logic just use the next line on *Close* button in the HTML tag

> ng-click="walletBalanceModal.hide()"

And delete the controller

ajp8164:
It may have more logic later.  I had thought about adding a "refresh" button to refresh balances.  We may also want to consider providing details behind confirming and locked amounts (like lists of tx and txp's).

But we can remove if for now and add the t back later as we need.

--
Andy Phillipson
Director of Engineering Services, BitPay, Inc.
844A CCA5 F277 30C0 E86F  9799 0CEE 502F 2484 0854

www/views/modals/wallet-balance.html
```diff
@@ -0,0 +1,90 @@
...
+              </span>
+            </div>
+          </span>
+        </div>
+      </div>
+      <div class="wallet-balance__description" translate><b>Total</b> is the total amount of bitcoin stored in this wallet. It is the sum of Available, Confirming, and Locked amounts.</div>
+      <div class="wallet-balance__item item">
+        <i class="wallet-balance__icon icon ion-ios-checkmark-outline"></i>
+        <div class="wallet-balance__content no-border">
+          <div class="wallet-balance__title">Available</div>
```
gabrielbazan7: needs translate

ajp8164: Done.

JDonadio:
ACK. Great job @ajp8164 !

tonygallippi:
ACK.  works exactly as expected.  I approve

Issue #5223, opened 2016-12-14
Contact name not showing in proposals view


JDonadio:
From @tonygallippi: If I have more than 3 proposals, I get prompted to see more in a Proposals screen.  on this screen the address book entry is not showing

From Home view:

<img width="420" alt="copay_-_copay_bitcoin_wallet" src="https://cloud.githubusercontent.com/assets/13851807/21187030/47fd8e3c-c1f5-11e6-9ab0-5e3a7d171cf8.png">

From Proposals view:

<img width="433" alt="copay_-_copay_bitcoin_wallet" src="https://cloud.githubusercontent.com/assets/13851807/21187049/5a8a2e34-c1f5-11e6-8318-bb5df3627873.png">

PR #5226, opened 2016-12-14
Use contact name in proposal card

JDonadio:
Closes https://github.com/bitpay/copay/issues/5223

### DIFF
--- a/src/js/controllers/proposals.js
+++ b/src/js/controllers/proposals.js
@@ -25,4 +25,10 @@ angular.module('copayApp.controllers').controller('proposalsController',
     });
 
     $scope.openTxpModal = txpModalService.open;
+
+    $scope.createdWithinPastDay = function(time) {
+      var now = new Date();
+      var date = new Date(time * 1000);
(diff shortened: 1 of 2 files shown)


www/views/includes/txp.html
```diff
@@ -32,8 +32,7 @@
       </span>
     </span>
     <div>
-      <time class="wallet-details__tx-time" ng-if="tx.createdOn && createdWithinPastDay(tx.createdOn)">{{tx.createdOn * 1000 | amTimeAgo}}</time>
-      <time class="wallet-details__tx-time" ng-if="tx.createdOn && !createdWithinPastDay(tx.createdOn)">{{tx.createdOn * 1000 | date:'MMMM d, y'}}</time>
```
cmgustavo: If that function is not used any more, I think that it should be removed from controller

JDonadio: `createdWithinPastDay` function missing on _tab-home.js_ and _proposals.js_. Already added 👍

cmgustavo:
ACK

PR #5229, opened 2016-12-14
Fix creating transaction with no priv key

### LINKED ISSUE #5222: Error: Can not create a proposal without private key
From a wallet exported without private key.

```
No signing proposal: No private key
ionic.bundle.min.js:150 ReferenceError: txp is not defined
    at m.$scope.approve (copay.js:13488)
```

cmgustavo:
fixes #5222


gabrielbazan7:
ACK !

PR #5247, opened 2016-12-15
Passes back error message from BWC rather than the object.

ajp8164:
This PR needs to be coordinated with an update to BWC error message strings.  The string presented to the UI is coming from BWC error messages, which is ok but those messages are terse and not informative for a user.

Also see https://github.com/bitpay/bitcore-wallet-client/pull/343.

### DIFF
--- a/src/js/services/walletService.js
+++ b/src/js/services/walletService.js
@@ -144,7 +144,7 @@ angular.module('copayApp.services').factory('walletService', function($log, $tim
           if (err instanceof errors.NOT_AUTHORIZED) {
             return cb('WALLET_NOT_REGISTERED');
           }
-          return cb(err);
+          return cb(err.message);
         }
         return cb(null, ret);
       });


cmgustavo:
Not sure about this change. `walletService` (or any other service) should return error as an object, not plain text message.

Then the controller should be able to process that error to show the user (we are using `bwcError` service to process and translate errors from bwc).

ajp8164:
I discussed with Ivan (I think, could have been Javier).  Other paths in this method return an error string and this path was previously returning an object which results in a rather ugly error on the UI (string representation of a Json object).

matiu:
To display an error we should use:
```
          errorString = bwcError.msg(err, gettextCatalog.getString('BWS Error'));
```
While not ideal, is the way it is used on somewhere of the code already, and it gives more flexibility. I am sending a PR to update some parts we should be using it.

PR #5261, opened 2016-12-16
Wallet service errors

JDonadio:


ajp8164:
We may not need the new "if" statements in walletService.js because the errors are not changed from BWC (I think they will end up being noops).  bwcErrors should be able to work fine with the just the `return cb(err)` statement.

JDonadio:
You're right @ajp8164, thanks!
Updated.

JDonadio:
We need to do a complete review of errors at BWC. For now, errors like `WALLET_NOT_REGISTERED` are not be handling correctly.

gabrielbazan7:
ACK

gabrielbazan7:
ACK

Issue #5268, opened 2016-12-16
Bogus BWS - empty send view


JDonadio:
<img width="436" alt="copay_-_copay_bitcoin_wallet" src="https://cloud.githubusercontent.com/assets/13851807/21269126/4641d566-c390-11e6-8fa6-89cd308abe42.png">

Also, we should add something like:

<img width="430" alt="copay_-_copay_bitcoin_wallet" src="https://cloud.githubusercontent.com/assets/13851807/21269186/832b00b0-c390-11e6-9078-e7ef43272b1d.png">

PR #5289, opened 2016-12-19
Empty send view - refresh

JDonadio:
With a bogus bws: If we wait on home for a while and then go to send, the page remains blank.

Related https://github.com/bitpay/copay/issues/5268


gabrielbazan7:
This timeout with the apply : https://github.com/bitpay/copay/pull/5289/files#diff-9df2a6ae82ed07c56cb66b0af31efa0bL151 should be moved inside this if statement : 
`if (index == wallets.length) $scope.checkingBalance = false;`

PR #5295, opened 2016-12-20
fix format amount

### LINKED ISSUE #5162: Format alternative amount - Confirm view
Alternative amount is not correctly formated

* USD

copay_-_copay_bitcoin_wallet

copay_-_copay_bitcoin_wallet

* EUR

copay_-_copay_bitcoin_wallet

copay_-_copay_bitcoin_wallet

JDonadio:
Fixes #5162

### DIFF
--- a/src/js/services/txFormatService.js
+++ b/src/js/services/txFormatService.js
@@ -1,6 +1,6 @@
 'use strict';
 
-angular.module('copayApp.services').factory('txFormatService', function(bwcService, rateService, configService, lodash) {
+angular.module('copayApp.services').factory('txFormatService', function($filter, bwcService, rateService, configService, lodash) {
   var root = {};
 
   root.Utils = bwcService.getUtils();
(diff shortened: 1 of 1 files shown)


JDonadio:
Good catch @cmgustavo. Updated now.

src/js/services/txFormatService.js
```diff
@@ -48,10 +48,11 @@ angular.module('copayApp.services').factory('txFormatService', function(bwcServi
     var config = configService.getSync().wallet.settings;
 
     var val = function() {
-      var v1 = rateService.toFiat(satoshis, config.alternativeIsoCode);
+      var v1 = parseFloat((rateService.toFiat(satoshis, config.alternativeIsoCode)).toFixed(2), 10);
```
cmgustavo: What does the number `10` means?

JDonadio: This is an error, that parameters does nothing. 👍

PR #5326, opened 2016-12-22
Feat/app identity

ajp8164:
All binding of the identity maintained by the app to the bitpay server is embedded in the bitpay card implementation.  This PR refactors the bitpayCardCredentials to a more general purpose appIdentity.

A new service is created that exposes an interface to the appIdentity (appIdentityService).  A single method, getAppIdentity(), returns either the single existing identity or it creates, stores, and returns a new app identity.

This app identity should be used for all user account binding between the app and the bitpay server.

A second new service (bitpayService) is extracted from the bitpay card implementation and made reusable for integrating future services provided by bitpay.


ajp8164:
Ok, I'll have a look at it today

cmgustavo:
ACK

src/js/services/bitpayService.js
```diff
@@ -0,0 +1,193 @@
+'use strict';
+
+angular.module('copayApp.services').factory('bitpayService', function($log, $http, platformInfo, appIdentityService, bitauthService, storageService, gettextCatalog, popupService) {
+  var root = {};
+
+  var NETWORK = 'livenet';
+  var BITPAY_API_URL = NETWORK == 'livenet' ? 'https://bitpay.com' : 'https://test.bitpay.com';
+
+  root.getEnvironment = function() {
```
matiu: To match the rest of Copay's code, this would probably be either: getNetwork
or getEnvironment and return `{ network: "livenet" }`.

src/js/services/appIdentityService.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+angular.module('copayApp.services').factory('appIdentityService', function($log, lodash, storageService, bitauthService) {
+  var root = {};
+
+  root.getIdentity = function(network, cb) {
+    var pubkey, sin, isNew;
+    storageService.getAppIdentity(network, function(err, data) {
+      if (err) return cb(err);
+      if (lodash.isString(data)) {
```
matiu: Probably lines 10 to 12 should be inside getAppIdentity

src/js/services/storageService.js
```diff
@@ -78,6 +78,14 @@ angular.module('copayApp.services')
     //
     // UPGRADING STORAGE
     //
+    // TODO: prior to upgrading storage a backup of all storage should be captured and made available
```
matiu: Could you explain this more? We cannot merge this to master, if this is a requirement.

src/js/services/storageService.js
```diff
@@ -119,6 +131,123 @@ angular.module('copayApp.services')
...
+            if (err) return cb(err);
+            storage.remove(key, function() {
+              cb(null, 'replaced with \'appIdentity\'');
+            });
+          });
+        } else {
+          cb();
+        }
+      });
+    };
```
matiu: I think it would be safer that when we upgrade a key, we ALWAYS use a new key for the upgraded data,
as it is done for bitpayCardCredentials and setBitpayDebitCards.

Could you change the new of the key of the upgraded value of `bitpayAccounts` ?

src/js/controllers/bitpayCardIntro.js
```diff
@@ -1,64 +1,48 @@
...
-                $state.go('tabs.home').then(function() {
-                  if (data.cards[0]) {
-                    $state.transitionTo('tabs.bitpayCard', {
-                      id: data.cards[0].id
-                    });
-                  }
-                });
-              });
+      var pairingReason = gettextCatalog.getString('add your BitPay Visa<sup>&reg;</sup> card(s)');
+      bitpayService.pair(pairData, pairingReason, function(err, paired, apiContext) {
```
ajp8164: Thanks!

src/js/services/bitpayService.js
ajp8164: Done.

src/js/services/appIdentityService.js
ajp8164: Done.

src/js/services/storageService.js
ajp8164: That TODO is not a requirement; just something I thought about for the future.  I can remove the comment if you like.

src/js/services/storageService.js
matiu: thanks!

PR #5332, opened 2016-12-22
fix design issue on specific amount feature

### LINKED ISSUE #5195: Request Specific Amount feature needs design work
The button seems to be general for all the view. But, it use the particular address that is shown at that moment.

copay_-_copay_bitcoin_wallet

gabrielbazan7:
fixes #5195


JDonadio:
Nice work. However, if you complete the process and click on "Finish" button navigate back to addresses view and the back arrow disappear.

gabrielbazan7:
Updated

PR #5349, opened 2016-12-27
BitPay card: Fix value in USD on confirm view

### LINKED ISSUE #5317: Toping a debit card in USD, sometimes does not match exactly the value entered
I have entered 400USD, and I got 399.95 on the transaction confirmation view.

cmgustavo:
fixes #5317


gabrielbazan7:
ACK

Issue #5302, opened 2016-12-20
[BitPay 1.2.1] UI Bug - Menu popover


Kirvx:
### Android 6.0.1 - Z3 Compact

![362](https://cloud.githubusercontent.com/assets/6627786/21358495/286df2c6-c6d9-11e6-91db-bea4badefdcc.jpg)

PR #5351, opened 2016-12-27
remove menu popover

JDonadio:
Fixes https://github.com/bitpay/copay/issues/5302


gabrielbazan7:
I think we should remove the check icon, and also change the directive and html name, so we could add for example the paste from clipboard option in the future on that action sheet.

JDonadio:
This is a selectable menu with `only one item`, why do you want to add a paste option there?

gabrielbazan7:
In the future

Issue #5307, opened 2016-12-20
Support Trezor / Ledger on Bitpay App


matiu:
While Trezor / Ledger support in the future is still undefined due to the end of Chrome app  support  announced by Google, we need to support them in the mean time, for our Copay users, when they upgrade to the new platform.

PR #5363, opened 2016-12-28
Chrome app for Copay

matiu:
- Restore Chrome App for Copay. 
- Fixes error when trying to import empty a mnemonic
- Fixes importing and creating form HW wallets

ToDo:
- Join from HW wallets
- Send TXs from HW wallets


JDonadio:
related https://github.com/bitpay/copay/issues/5307

matiu:
updated.

PR #5370, opened 2016-12-29
Update translation files

JDonadio:


JDonadio:
Many translations are missing, please let me update this

JDonadio:
Updated.

PR #5386, opened 2017-01-02
Fix wallet deletion without push

antonio-fr:
Fix a BWS crash when deleting a wallet with push disabled. No need of push unsubscribe when push is disabled. Else this leads to a full crash and stop of BWS server incase it doesn't support push.


JDonadio:
Thanks a lot @antonio-fr !

Issue #5404, opened 2017-01-04
Specify service being reviewed in feedback flow


jameswalpole:
From @ajp8164: "Collect the identity of the service(s) for which the feedback applies. Ideally we could automatically forward this feedback directly to the service provider (e.g., Glidera) in real-time so they get a sense of the issues and urgency to provide fixes."

![feedback view](https://cloud.githubusercontent.com/assets/9257396/21652729/f5d38fe8-d27b-11e6-84fe-6c71692f6d6e.png)

jameswalpole:
@bitjson @matiu Have a look at this part of the solution to the feedback confusion issue. I think @ajp8164 wants to forge ahead.

ajp8164:
See https://github.com/bitpay/copay/pull/5414

ajp8164:
![screen shot 2017-01-05 at 12 32 19 pm](https://cloud.githubusercontent.com/assets/1063765/21690628/0f70bbac-d343-11e6-849e-0542baa9c8c1.png)

PR #5414, opened 2017-01-05
Feat/services feedback

ajp8164:
Requires design review.  Feedback form allows user to check one or more enabled services (which are ideally the subject of the feedback).  The data collected now includes "enabledServices" and "feedbackServices".  Enabled services lists all the services enabled in the users app.  Feedback services lists the services that were checked by the user on the feedback form.


ajp8164:
Also see #5404

ajp8164:
@bitjson closing this. Re-open if you think it's usable..

PR #5421, opened 2017-01-06
Improved bitpay account pairing and management of paired state and data.

ajp8164:
Creates a new bitpayAccountService to manage all behavior with pairing a bitpay account and obtaining account information.  The bitpay account is elevated to the UI so users can see the account is paired and remove it.  Account pairing UX is unchanged.  If the user has already paired (by adding a card) then when the user attempts to (re)connect the card (or connect again to fetch a new card) an action sheet presents that shows the existing account pairing as well as an option fetch cards from another (not yet paired) account.  Also, in settings (Wallets and Integrations) the Bitpay Debit Cards entry is renamed to BitPay with the 'b' icon and when clicked reveals a view that shows all connected cards and accounts.  The user may remove one card at a time or remove all cards and unlink the account by just removing the account.  For example, if the user has three cards on their account and they want to unlink their account they can perform one action to remove the account (which will also remove all the cards associated with that account).

This PR is a predecessor to introducing new services that require this framework.


ajp8164:
Home view with multiple cards across multiple accounts (no UX changes).

screen shot 2017-02-09 at 3 10 23 pm

Add/pair card base view (no UX changes).

screen shot 2017-02-09 at 3 11 01 pm

Add/pair card view after clicking "Connect my BitPay card" when at least one account is already paired (new account selector slides up).  No account selector if no accounts paired (launches browser as expected). Clicking "Add account" launches browser to pair as expected.

screen shot 2017-02-09 at 3 10 43 pm

Settings view when at least one account is paired (entry is updated).

screen shot 2017-02-09 at 3 11 13 pm

Settings view now lists each card and account separately. Can remove cards or accounts. Removing an account removes all cards attached to the account. If card removed it can be quickly re-added using new account selector seen above (no new pairing hassle since account is still paired).

screen shot 2017-02-09 at 3 11 25 pm

PR #5425, opened 2017-01-06
Disable hw from mobile

JDonadio:


JDonadio:
You're right, fixed now. Thanks

JDonadio:
Agree. Updated

PR #5438, opened 2017-01-11
Fix nav back button

tuladhar:
The nav back button doesn't show up when there's one or more contacts. Thus, only way to go back is to exit the app :(

### DIFF
--- a/www/views/addressbook.html
+++ b/www/views/addressbook.html
@@ -1,10 +1,10 @@
 <ion-view id="view-address-book" hide-tabs>
   <ion-nav-bar class="bar-royal">
-    <ion-nav-back-button>
-    </ion-nav-back-button>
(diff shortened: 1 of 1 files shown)


JDonadio:
@tuladhar thank you very much for you contributions. This is an knows issue that we are trying to fix for the next release. Unfortunately this changes are not solving all the scenarios. Please take a look this steps:

1. Go to "Home Tab".
2. Go to "Wallet Details" clicking on a wallet.
3. Click on backbutton.
4. Go to "Send Tab".
5. Go to "Amount" clicking on a wallet.
6. Click on backbutton
7. Repeat 1, 2, 3, 4, 5

Ionic history seems to remove some state or so during the navigation, we continue investigating this case.

PR #5439, opened 2017-01-11
Use appropriate icon

tuladhar:
Update icon from receive to scan which is shown when camera authorization is required.


JDonadio:
NACK.

The icon is correct. If you use the scan icon users may confuse that with the main and try to tap there to scan.

bitpay_wallet_screens_-_invision

tuladhar:
Well, even if they click, it will not work :)

Thanks for the reply!

JDonadio:
Sure, but shouldn't be confusing for anyone 😉. Thanks!

Issue #4935, opened 2016-10-27
Use a native hook to avoid "testing" cameras on desktop


bitjson:
This will have to wait until after https://github.com/bitpay/copay/issues/4890.

From [this reddit thread](https://www.reddit.com/r/Copay/comments/59q0uz/this_is_somewhat_disturbingi_downloaded/): 

> ### This is somewhat disturbing...I downloaded Copay/Bitpay wallet for mac
> 
> ...and when I start it up the camera light flashes 3 times. Why is your wallet taking pictures of me????

[My response](https://www.reddit.com/r/Copay/comments/59q0uz/this_is_somewhat_disturbingi_downloaded/):

> I'm probably in the best position to answer this since I wrote [the code that does it](https://github.com/bitpay/cordova-plugin-qrscanner/blob/master/src/browser/src/createQRScannerInternal.js#L27)!
> 
> The code is part of the new [BitPay-developed QR scanning library](https://github.com/bitpay/cordova-plugin-qrscanner/).
> 
> Browsers try to protect users' privacy by refusing to "tell" applications anything about the cameras available to your computer. (I would argue this is just privacy-theater – as I have shown by my workaround – but that's another discussion.)
> 
> Basically, applications must "ask" for a camera of "at least width X and height X", and the browser will either provide access to a camera, or respond "no matching camera available".
> To work around this, I wrote a short "startup" process which quickly "tests" all the cameras connected to your computer, and remembers the best two. When you go to scan a QR code, we can instantly choose the best one. If you "rotate" the camera, we switch to the second-best camera.
> 
> It's silly, but necessary for a fast user experience.
> 
> In the future, I'm hoping we can instead tie into native functionality on each platform, and completely avoid the blinking privacy lights.
> 
> Thank you for the feedback! It's valuable to know that you 1) noticed and 2) were concerned. I've just [opened an issue](https://github.com/bitpay/copay/issues/4935) in the Copay GitHub project to get it on the list for future development.

The change would require changes in both Copay and [cordova-plugin-qrscanner](https://github.com/bitpay/cordova-plugin-qrscanner).

huangbong:
I agree that this is a huge issue. The first thing I noticed when I open the app on my Mac was the 3 green flashes from the webcam light and I'm like wtf is going on. It feels like it's taking 3 snapshots of your face like you're passing under a toll booth in your car...

[...]

Issue #5172, opened 2016-12-08
Process pegs cpu (linux)


gabegattis:
Not totally sure what happened here, but I think it is related to the scanner.

The cpu spike was temporary, lasting for maybe 2 minutes or so. I looked in the dev console, and the spike coincided with this log message

`Scanner not in use for 300 seconds, destroying...`

JDonadio:
Can you check it again after this improvement?

Issue #4967, opened 2016-11-02
Desktop: after scan a QR code, the camera remain active


cmgustavo:
Note: I have to close the app

bitjson:
@cmgustavo could you provide a little more detail on this? Do you mean the privacy light on your laptop stays active? Or the video preview is still active? (Is this still an issue with the new scanner work?)

cmgustavo:
The light on the laptop stays active (for long time, I had to close the app)

Note: I tested it with a simple bitcoin address (bitcoin:1xxx...xxx)

bitjson:
Right now it should stay "enabled" for [10 seconds](https://github.com/bitpay/copay/blob/master/src/js/services/scannerService.js#L133) after visiting the scan tab. Though if you scan something successfully, it should go off immediately.

Looks like there are some strange edge cases where scanning won't be properly deactivated after scanning. Definitely needs to be reviewed 👍

gabegattis:
can confirm on linux

Issue #4946, opened 2016-10-28
Android: performance – scanner remaining active for too long?


Kirvx:
I can hear the auto focus of the camera on my Sony Z3 Compact (Android 6.0.1) in others tabs than the Scan tab :smile:

The problem is that my phone heats as if I was playing a game ~~, and it causes a significant decrease of performance in the app.~~

So it looks like a big app for my phone, requires a lot of resource :zap:

bitjson:
Hey @Kirvx – thanks for the report! It's very possible the camera is staying activated longer than it should be.

With the new scanner, we've somewhat disconnected the UI the scanner itself to improve performance. The scanner first turns on when you open the scan view, but it remains functional (with scanning disabled) [for 10 seconds](https://github.com/bitpay/copay/blob/master/src/js/services/scannerService.js#L133) after leaving the view. This makes it extremely quick to reactivate if you switch back to the scan tab.

After that 10 seconds though, scanning should stop, and resource consumption should fall to nearly zero. (After a little longer, the scanner is also "destroyed" completely, and no longer consumes any resources.)

If the scanner is still clearly active on your device, it might be a bug with the native Android implementation. I'll test on some other Android devices and see if I can figure out the issue.

Kirvx:
Thanks :)

Actually the scanner starts in background with the launch of the app, I don't have to run the scanner on the Scan tab an go back to notice the problem.

bitjson:
Good to know. I should have also mentioned that we attempt to [pre-initialize the scanner right after the app starts](https://github.com/bitpay/copay/blob/master/src/js/services/scannerService.js#L83) to improve the initial load time a little more.

So that would also point to an issue with [the android implementation](https://github.com/bitpay/cordova-plugin-qrscanner/tree/master/src/android) of cordova-plugin-qrscanner.

Kirvx:
Just to inform that the camera doesn't cause the "significant decrease of performance in the app" I described, it's the same when the camera is disabled (it still cause heat release).

I hope Ionic 2 makes the app faster :)

Kirvx:
### How the camera working in the background cause problem for the battery life : 

> ### BitPay 1.2.1 consumes 47.8% more energy than "reddit is fun" for a duration of use lower by 81.50%.

![26](https://cloud.githubusercontent.com/assets/6627786/21750334/266a7bde-d5b1-11e6-84da-158ac2090b2c.jpg)

[...]

PR #5440, opened 2017-01-12
Improve scanner handling on all platforms

bitjson:
Fixes: #4935, #4967, #5172, #4946, and #4831

### DIFF
--- a/src/js/controllers/tab-scan.js
+++ b/src/js/controllers/tab-scan.js
@@ -57,8 +57,10 @@ angular.module('copayApp.controllers').controller('tabScanController', function(
 
   $scope.$on("$ionicView.afterEnter", function() {
     // try initializing and refreshing status any time the view is entered
-    scannerService.gentleInitialize();
-    scannerService.resumePreview();
+    if(!scannerService.isInitialized()){
+      scannerService.gentleInitialize();
+    }
+    activate();
   });
 
   function activate(){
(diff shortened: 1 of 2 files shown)


matiu:
Please note that this seems to also happen on mobile.

eg.  is you turn on the flash on mobile and navigate away from the scanner, the flash will stay on indefinitely.

bitjson:
@matiu this PR is only for the "privacy light" (little green light by the camera), and there are currently no mobile devices with privacy lights. So I think this only concerns the desktop build.

On what you mentioned – would you expect the flash to work differently? If you tap the scan tab again, the flash button is still highlighted, right?

I guess we could disable the light if the user navigated away, but I think it's probably a better experience for the light to only activate/deactivate when the user taps the button. If the user is in a dark room and activated the light to scan a printed QR code, I would think the light should stay on until the user deactivates it.

bitjson:
I'm still reviewing on mobile devices. These are small changes, but the logic is pretty complicated, so I'd like to be very sure it's completely correct now.

bitjson:
This is ready for review. It would be good to test a bit more on Android, particularly.

cmgustavo:
It works for me: (iOS and macOS).

But in the mobile I can see that after turning on the light, this still active after changing tab.

bitjson:
@cmgustavo – see [above](https://github.com/bitpay/copay/pull/5440#issuecomment-273573385) - I don't think that's a bug. 

> I guess we could disable the light if the user navigated away, but I think it's probably a better experience for the light to only activate/deactivate when the user taps the button. If the user is in a dark room and activated the light to scan a printed QR code, I would think the light should stay on until the user deactivates it.

Would you expect it to work differently? I tried a bit both ways, and it seems more intuitive like this? (I don't have a strong opinion on it though, I could switch it)

PR #5449, opened 2017-01-12
Display updated memo on save

tuladhar:
Updated memo isn't displayed to the user on save on same view. User has to perform some other action such as re-edit or go back, to see the changes reflect. To fix this, added $scope.$apply() to make sure changes are displayed to the user on save.


JDonadio:
Good catch! I could reproduce it on my phone. Thank you!

PR #5451, opened 2017-01-13
cordova wp8 build

gabrielbazan7:
Not solved issues: 

- Scanner is not working
- npm i fail. (Node rebuild error of bitauth library)
- Needs BitPay app assets
- Some images are not shown
- Needs publisher ID and wp Main Page Id to build BitPay wallet


gabrielbazan7:
Close this one. Replaced by #5579

PR #5456, opened 2017-01-14
Allow user to cancel on going progress and hide spinner dialog

tuladhar:
User doesn't have ability to cancel on going progress, thus spinner dialog runs forever. Otherwise, during no network connection, if user goes to view their transaction the spinner dialog will run forever and requires app restart to get back. Set root.clear as callback, so user are able to cancel the progress and hide the spinner dialog.


JDonadio:
Thank you @tuladhar, definitely we have a lot of work on non-connection scenarios

PR #5460, opened 2017-01-16
Fix update url

JDonadio:


matiu:
????

JDonadio:
My bad in [this](https://github.com/bitpay/copay/pull/5450/files) previous PR.

`release url` returns _vX.X.X_ and we should open the latest release page for update the app

PR #5476, opened 2017-01-18
Fix first link to your account. Removes unused variables. Fix spinner

cmgustavo:
* Logout function also removes txs

### DIFF
--- a/src/js/controllers/coinbase.js
+++ b/src/js/controllers/coinbase.js
@@ -18,7 +18,7 @@ angular.module('copayApp.controllers').controller('coinbaseController', function
           if (err) {
             $log.error(err);
             err = err.errors ? err.errors[0].message : err;
-            popupService.showAlert('Error', err);
+            popupService.showAlert('Error connecting to Coinbase', err);
           }
           return;
         }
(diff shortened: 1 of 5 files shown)


src/js/controllers/coinbase.js
```diff
@@ -16,6 +16,8 @@ angular.module('copayApp.controllers').controller('coinbaseController', function
         $scope.loading = false;
         if (err || lodash.isEmpty(data)) {
           if (err) {
+            $log.error(err);
+            err = err.errors ? err.errors[0].message : err;
             popupService.showAlert('Error', err);
```
matiu: Should be: 'Error connecting to Coinbase'

src/js/controllers/coinbase.js
```diff
@@ -79,18 +81,15 @@ angular.module('copayApp.controllers').controller('coinbaseController', function
...
-          return;
-        }
-        $scope.accessToken = accessToken;
-        init();
-      });
-    }, 100);
+    coinbaseService.getToken(code, function(err, accessToken) {
+      ongoingProcess.set('connectingCoinbase', false);
+      if (err) {
+        popupService.showAlert('Error', err);
```
matiu: Should be: 'Error connecting to Coinbase'

cmgustavo:
Updated. Thanks!

PR #5479, opened 2017-01-18
fix delete tx history

matiu:
- fixes Delete Tx History
- Add wallet indicator in preferences inner views.

### DIFF
--- a/src/js/services/walletService.js
+++ b/src/js/services/walletService.js
@@ -551,6 +551,21 @@ angular.module('copayApp.services').factory('walletService', function($log, $tim
     };
   };
 
+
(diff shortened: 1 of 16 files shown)


src/js/services/walletService.js
```diff
@@ -551,6 +551,21 @@ angular.module('copayApp.services').factory('walletService', function($log, $tim
     };
   };
 
+
+  root.clearTxHistory = function(wallet, cb) {
```
JDonadio: This function never returns `err`

matiu: updated, thanks.

Issue #5483, opened 2017-01-19
[UI - BUG] Enable/disable integrations for buy & sell bitcoins


JDonadio:
Fresh client or never linked an account to any service:

**Enable Glidera or Coinbase service**

* Next step card visible on home with Buy or Sell Bitcoin option
* Click to see the enabled services and always shows both
* If navigates from services list (tab home) to advanced settings (tab settings) disable both and returning to home you can still see the services list. Should navigates to home overview page.

PR #5486, opened 2017-01-19
Fix enabled services list

JDonadio:
Fixes https://github.com/bitpay/copay/issues/5483

### DIFF
--- a/src/js/controllers/advancedSettings.js
+++ b/src/js/controllers/advancedSettings.js
@@ -24,9 +24,6 @@ angular.module('copayApp.controllers').controller('advancedSettingsController', 
     $scope.recentTransactionsEnabled = {
       value: config.recentTransactions.enabled
     };
-    // $scope.frequentlyUsedEnabled = {
-    //   value: config.frequentlyUsed.enabled
-    // };
   };
 
   $scope.spendUnconfirmedChange = function() {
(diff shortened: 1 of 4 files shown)


src/js/controllers/buyandsellController.js
```diff
@@ -0,0 +1,19 @@
...
+
+angular.module('copayApp.controllers').controller('buyandsellController', function($scope, $ionicHistory, $state, configService) {
+
+  $scope.$on("$ionicView.beforeEnter", function(event, data) {
+    configService.whenAvailable(function(config) {
+      $scope.isCoinbaseEnabled = config.coinbase.enabled;
+      $scope.isGlideraEnabled = config.glidera.enabled;
+
+      if (!$scope.isCoinbaseEnabled && !$scope.isGlideraEnabled)
+        goHome();
```
cmgustavo: Why don't use `goBack()` instead?

src/js/controllers/advancedSettings.js
```diff
@@ -24,9 +24,9 @@ angular.module('copayApp.controllers').controller('advancedSettingsController',
     $scope.recentTransactionsEnabled = {
       value: config.recentTransactions.enabled
     };
-    $scope.frequentlyUsedEnabled = {
-      value: config.frequentlyUsed.enabled
-    };
+    // $scope.frequentlyUsedEnabled = {
+    //   value: config.frequentlyUsed.enabled
+    // };
```
cmgustavo: If it is not used any more, please remove it.

JDonadio:
Updated. Thanks!

PR #5501, opened 2017-01-21
Fix delete wallet view cache and transition. Fix UI details

cmgustavo:
* After delete wallet, the view "deleteWallet" remain in cache, getting error in console log.

### DIFF
--- a/src/js/controllers/preferencesDelete.js
+++ b/src/js/controllers/preferencesDelete.js
@@ -1,12 +1,19 @@
 'use strict';
 
 angular.module('copayApp.controllers').controller('preferencesDeleteWalletController',
-  function($scope, $stateParams, $ionicHistory, gettextCatalog, lodash, profileService, $state, ongoingProcess, popupService) {
-    var wallet = profileService.getWallet($stateParams.walletId);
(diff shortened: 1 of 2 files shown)


src/js/controllers/preferencesDelete.js
```diff
@@ -1,12 +1,19 @@
...
-    var wallet = profileService.getWallet($stateParams.walletId);
-    $scope.wallet = wallet;
-
-    $scope.alias = lodash.isEqual(wallet.name, wallet.credentials.walletName) ? null : wallet.name + ' ';
-    $scope.walletName = wallet.credentials.walletName;
+  function($scope, $ionicHistory, gettextCatalog, lodash, profileService, $state, ongoingProcess, popupService) {
+    
+    $scope.$on("$ionicView.beforeEnter", function(event, data) {
+      if (!data.stateParams || !data.stateParams.walletId) {
+        popupService.showAlert(null, gettextCatalog.getString('Bad param'), function() {
```
JDonadio: I think 'Bad param' is not clear

cmgustavo: Updated. Thanks!

PR #5505, opened 2017-01-22
Synchronize all storage upgrade.

ajp8164:
Fixes disappearance of debit cards from UI.  Users may have to pair again.
This needs to be tested on a release prior to 1.3 and a "broken" 1.3 release.


matiu:
Also, please do not modify `Object.` prototype inside storageService.

ajp8164:
I'll make it a private function.

PR #5542, opened 2017-01-27
Refresh external service on settings-tab

cmgustavo:


JDonadio:
Can you try removing the 100 ms in the timeouts? Because if the process takes less than 100 ms the redrawing view experience will be faster.

cmgustavo:
Yes, it works without adding 100ms. Thanks.

JDonadio:
Actually if you are using `apply` or `digest` within an async function you don't need the timeouts (untested)

cmgustavo:
Code obsolete. Closes #5544

Issue #5411, opened 2017-01-05
Deleting tx proposal doesn't hide the balance button


JDonadio:
<img width="410" alt="copay_-_copay_bitcoin_wallet" src="https://cloud.githubusercontent.com/assets/13851807/22521130/b31b05e8-e895-11e6-848f-6f6ab912e5e9.png">

<img width="413" alt="copay_-_copay_bitcoin_wallet" src="https://cloud.githubusercontent.com/assets/13851807/22521164/d0b00478-e895-11e6-8337-4d9d2c17aba4.png">

PR #5557, opened 2017-02-01
Balance button hide

JDonadio:
Fixes https://github.com/bitpay/copay/issues/5411


src/js/controllers/walletDetails.js
```diff
@@ -271,7 +271,7 @@ angular.module('copayApp.controllers').controller('walletDetailsController', fun
   function refreshAmountSection(scrollPos) {
     $scope.showBalanceButton = false;
     if ($scope.wallet.status) {
```
cmgustavo: Should it be `$scope.status` also?

JDonadio: Your'e right. Thanks

PR #5560, opened 2017-02-02
feat(cordova): add screen privacy for iOS and Android

bitjson:
Tested on iOS and Android.

Works really well on iOS The Android implementation is a bit strange (just leaves the app preview blank), but in testing other apps, it doesn't seem like there's a better way to do it. So this is probably good for now.


JDonadio:
ACK

Blank screen on android, but it's better than watch the balance 😅

Issue #5497, opened 2017-01-20
Add PIN code or password for viewing backup phrase after initial backup


jameswalpole:
We need to add password protection or PIN code protection on the backup phrase viewing screen (in Settings > Wallet > Backup Wallet) if we want to keep it. Right now anyone with access to a phone (unlocked) could open a BitPay wallet and view its backup phrase. 

@cmgustavo says that users with spending passwords already enabled will be prompted to provide those passwords when trying to view this screen. The problem is that users are not told that they need to set spending passwords in order to protect this information within the app. It's not clear that "Spending Passwords" are also "Passwords for Protecting Sensitive Info from Prying Eyes."

One solution is to prompt **all** users to create spending password for their new wallets in the wallet creation flow so as to include this security by default. Other solution is PIN code for accessing the app interface. Finally, you could just

matiu:
This was discussed before at:
https://github.com/bitpay/copay/issues/4555
and
https://github.com/bitpay/copay/issues/4617

JDonadio:
> @cmgustavo says that users with spending passwords already enabled will be prompted to provide those passwords when trying to view this screen. The problem is that users are not told that they need to set spending passwords in order to protect this information within the app. It's not clear that "Spending Passwords" are also "Passwords for Protecting Sensitive Info from Prying Eyes."

Maybe we should add some description to clarify it, like:

<img width="414" alt="copay_-_copay_bitcoin_wallet" src="https://cloud.githubusercontent.com/assets/13851807/22375357/71f1a39e-e488-11e6-9f36-7f2db40d107f.png">

darakaye:
Can someone help me?  I did not set a spending password.  I cannot access my btc.

BGalliotti:
Hello @darakaye ,

Thanks for reaching out to us.

If you are being prompted for a password, that means you set up a password. 
Very likely you did it on the app onboarding and do not remember. Certainly the app cannot generate a password and lock wallets with random passwords.

Since the password is stored locally on your device, there is nothing we can do to restore it.

There are two ways to solve this situation:

1. Remember the password. Keep in mind there are no special requirements for the password such as numbers, symbols, uppercase letters, etc.

2. Restore the wallet with the 12-word recovery phrase.

For option 2:

**What you can do is restore the wallet.**

**A)** First, make sure the recovery phrase you have, corresponds to the wallet you want to restore.

i) **IF YOU HAVE ONLY ONE WALLET IN THE APP**

[...]

Issue #5548, opened 2017-01-30
Disable spending password


JDonadio:
Spending password should be disabled if the backup flow was not completed.

dabura667:
Or require a spending password upon wallet creation and tell the user they may remove the password after writing down their backup by going into the menu.

This way it will make them keenly aware of having set a password for spending.

A side note: How can someone get money on to Copay if the backup hasn't been written down? Copay doesn't show any addresses without writing down the backup.

JDonadio:
> Or require a spending password upon wallet creation and tell the user they may remove the password after writing down their backup by going into the menu.

Could be another option. Personally I prefer disable 🤓 

> A side note: How can someone get money on to Copay if the backup hasn't been written down? Copay doesn't show any addresses without writing down the backup.

It's mandatory for Copay but not for BitPay app. There was, are, and surely will be a lot of discussion about if this is correct or not, you probably knows this story. Honestly I don't have the lastest directive or ticket to tell you explicitly why we made it in this way.

sadpandas:
I feel really dumb but this is how I lost $150.00
My Lastpass note didn't save so my generated password was lost

Issue #5547, opened 2017-01-30
Spending password issue


slinky2:
I never set up a spending password and I can no longer find my mnemonic for this specific wallet.  Is there a default password, assuming I never set one up, there should not a a password for spending. Without the password my bitcoin is essentially frozen without a way to transfer.  I have another wallet I can use, but I cannot transfer the bitcoin to it. Thank you.

JDonadio:
Do you have the backup?

slinky2:
Unfortunately not for this wallet, otherwise I would delete and reset. My issue is not understanding why it requests a sending password when I never set one. It only makes sense that there would be no password associated at this point. Understandably,  it is my duty to have the recovery phrase, but it is not my duty to have a password that was never set. This concerns me bc I can no longer access my Bitcoin.

dabura667:
Have you ever left your phone unattended? Unlocked?

My son messed with my unlocked phone and managed to add a spending password to a wallet on Copay once... Almost as if he did it on purpose.

Luckily, I kept my phrase backups safe, so no harm was done.

JDonadio:
Unfortunately without a backup there is nothing we can do.

slinky2:
No, never.  I am the only person with access to my device and no password was set up. If I could send to my other wallet or even sell the bitcoin I would be fine with deleting and using my other wallet that has no spending password associated. Really do not understand the password request since there was never one set.

dabura667:
If you really think Copay is lying to you:

1. Export the wallet. You will need to set a password, "a" is fine.
2. use sjcl to decrypt the wallet using the "a" password you just set.
3. If you see your phrase in there, you didn't have a password set, if you don't see it, then you have encrypted the phrase and need the password.

https://bitwiseshiftleft.github.io/sjcl/demo/

Paste the exported wallet in Ciphertext, and type "a" in the password box to see the raw wallet data.

slinky2:
I know the wallets I have are separate identities, but aren't they connected in some way? 

And as far as my password, I am saying that I haven't set one up. I will see if that works.

slinky2:
Also, my wallet in question has an option to delete recovery phrase. Obviously the mnemonic phrase is still in existence. Is there any way of showing/recovering the mnemonic phrase? 

I don't know how to export my wallet, as I'm not brushed up on all the tech points. I'm just a simple man with enough knowledge to know that there are doors that have been opened and not yet closed so to speak.

slinky2:
[...]

PR #5570, opened 2017-02-06
Ref/disable spending password

JDonadio:
Fixes https://github.com/bitpay/copay/issues/5497
Related https://github.com/bitpay/copay/issues/5548
Related https://github.com/bitpay/copay/issues/5547

copay_-_copay_bitcoin_wallet_y_slack_-_bitpay

### DIFF
--- a/www/views/preferences.html
+++ b/www/views/preferences.html
@@ -41,6 +41,9 @@
       </div>
       <a class="item item-icon-right" ui-sref="tabs.preferences.backupWarning({from: 'tabs.preferences'})" ng-hide="wallet.isPrivKeyExternal()">
         <span translate>Backup</span>
+        <span class="right text-light assertive" ng-show="wallet.needsBackup">
+          {{'Backup needed' | translate}}
+        </span>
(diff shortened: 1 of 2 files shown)


gabrielbazan7:
Maybe "Spending Password" is not the correct label?. Why it ask me for the "Spending Password" when I try to see the backup words ? I'm not spending nothing. We should call it "Security Password"? 
Also in the description we should explain what setting the spending password really does. Then the user will understand why a request password prompt is shown when he tries to see the words and when he tries to send money.

JDonadio:
"Spending password" is clear for send action, but not for the access to the recovery phrase or private key options. I like "Security password", and maybe the info message should be something like: "If enabled, a request password prompt will be shown for spending, backup phrase or private key options.

isocolsky:
I'm not sure about calling it "security password". It could be argued that all passwords are meant for security one way or another. I think we could call it something like "wallet password" and state that "this password is used to protect all sensitive information (private keys, menmonics, etc) and actions (spending, exporting, etc) associated with the wallet".

PR #5574, opened 2017-02-06
New release v1.3.5

cmgustavo:


ivanrojasorg:
I cant install bitpay walley with the Commander ./Bitpay, how install bitpay walley un Ubuntu?

PR #5582, opened 2017-02-07
Enable scss overriding (e.g. for Copay vs. BitPay apps).

ajp8164:
These changes allow for the Copay app (npm run apply:copay) to provide its own styling to app (images and css).

You can place a scss at app-template/copay/sass/app.scss and have it automatically applied and included during the build.  CSS will override all other app CSS.  You can also use locally defined images in the app-template/copay/img directory and refer to them in any overriding scss defined in app.scss.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,7 +107,7 @@ src/js/externalServices.js
 chrome-app/manifest.json
 www/img/app
 www/index.html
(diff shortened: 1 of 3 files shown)


ajp8164:
Merged into #5612

PR #5584, opened 2017-02-08
Set image fill to none to allow css background to color the icon.

ajp8164:
Makes this icon the same as the others on the `add` view.


JDonadio:
I tried but really don't see the difference, can you add screenshots or something?

ajp8164:
You won't see any change on the app.  The image is stored with a fill color which is the same as the bitpay app css. If you change the css color the image color remains the same (because the fill color is stored with the image).

ajp8164:
Merged into #5612

PR #5606, opened 2017-02-10
Allow for wallet colors to be set by css.

ajp8164:
This change moves the definitions of wallet colors from javascript to css making it very easy to define a distinct and arbitrary set of wallet colors for copay vs. bitpay apps via css changes only.


ajp8164:
Merged into #5612

PR #5624, opened 2017-02-17
Ability to configure the bundle path.

ajp8164:


matiu:
What is the purpose of this PR?

ajp8164:
The build process assumes that the package id will always be`com.bitpay.*`.  For anyone else wanting to use this software as a base they will have to modify the build process for their own domain rather than just change a string in an overriding config file.  This pr allows the developer to easily set what ever package id they need without modifying the build process locally.

PR #5631, opened 2017-02-20
fix screen.lockOrientation is not a function

gabrielbazan7:


gabrielbazan7:
close in order to #5632

Issue #5600, opened 2017-02-09
After buying a gift card, this service appears duplicated on tab-home


cmgustavo:
![bitpay_-_secure_bitcoin_wallet](https://cloud.githubusercontent.com/assets/237435/22787196/3b153b24-eeba-11e6-8778-edf7a5b6444b.png)

PR #5639, opened 2017-02-21
Fix duplicated home integration items

cmgustavo:


JDonadio:
Fixes https://github.com/bitpay/copay/issues/5600

PR #5647, opened 2017-02-21
remove unnecessary translators view

gabrielbazan7:


cmgustavo:
Why? We should link to Crowdin from somewhere inside the app.

Note: `route.js` has a lot of unnecessary changes (spaces)

gabrielbazan7:
This is the page I remove: 

cursor_and_copay_-_copay_bitcoin_wallet

Have no new information, is the same text here: 

cursor_and_copay_-_copay_bitcoin_wallet

PR #5663, opened 2017-02-22
Swipe back 

gabrielbazan7:
When swipe back is set false it stays that way

### DIFF
--- a/src/js/controllers/confirm.js
+++ b/src/js/controllers/confirm.js
@@ -7,9 +7,14 @@ angular.module('copayApp.controllers').controller('confirmController', function(
   var countDown = null;
   var cachedSendMax = {};
   $scope.isCordova = platformInfo.isCordova;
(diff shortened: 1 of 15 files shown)


src/js/controllers/confirm.js
```diff
@@ -7,7 +7,6 @@ angular.module('copayApp.controllers').controller('confirmController', function(
   var countDown = null;
   var cachedSendMax = {};
   $scope.isCordova = platformInfo.isCordova;
-  $ionicConfig.views.swipeBackEnabled(false);
```
gabrielbazan7: @cmgustavo you add this here for a reason ?

cmgustavo: Because some mobile user that were trying to do "swipe to pay" and it goes back to previous view.

gabrielbazan7:
close in order to #5706

PR #5612, opened 2017-02-13
Initial Copay sass

ajp8164:
Includes sass for copay, updates to apply app sass to override base css, and ability to define wallet colors from sass.

screen shot 2017-02-14 at 4 43 18 pm

screen shot 2017-02-14 at 4 49 37 pm

PR #5672, opened 2017-02-23
Implementation for copay distribution sass.

ajp8164:
Recovery of #5612 on a new branch


JDonadio:
👏 great job!

bitjson:
Design ACK – awesome work!

There are a few small color/style things we'll probably polish up, but I think those would be better to do in smaller PRs.

Thanks for refactoring. Let's get this in a beta 🚀

PR #5697, opened 2017-03-02
Fix styling on mobile (slide to accept)

cmgustavo:


src/sass/variables.scss
```diff
@@ -93,6 +93,12 @@ $v-button-secondary-outline-color:                   $v-dark-gray !default;
 $v-button-secondary-outline-active-bg:               $v-dark-gray !default;
 $v-button-secondary-outline-active-color:            #ffffff !default;
 
+/* Slide */
+
+$v-slide-bg-color:                                   $v-accent-color;
```
ajp8164: Should be `$v-accent-color !default`

src/sass/views/includes/slideToAcceptSuccess.scss
```diff
@@ -1,5 +1,4 @@
 slide-to-accept-success {
-  $slider-bg-color: #5063B9;
   $success-bg-color: #11D1A6;
```
ajp8164: This color definition should be added to the variables.scss

PR #5711, opened 2017-03-03
Fix header bar background.

ajp8164:


gabrielbazan7:
Could you add a picture please ?

PR #5728, opened 2017-03-07
Ref tx-details

cmgustavo:
* Transition smoother
* Show date on receive tx
* Hide creator if not shared wallet
* Pre-load exchange rate (click on alternative amount)
* Click on confirmation row to refresh tx

### DIFF
--- a/www/views/tx-details.html
+++ b/www/views/tx-details.html
@@ -5,6 +5,11 @@
     </ion-nav-title>
     <ion-nav-back-button>
     </ion-nav-back-button>
+    <ion-nav-buttons side="secondary">
+      <button class="button back-button" ng-show="btx" ng-click="updateTx()">
+        <i class="icon ion-ios-refresh-empty"></i>
+      </button>
+    </ion-nav-buttons>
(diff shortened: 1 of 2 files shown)


www/views/tx-details.html
```diff
@@ -56,10 +58,16 @@
             <div>{{wallet.name}}</div>
           </div>
         </div>
-        <div class="item single-line" ng-if="btx.action != 'received'">
+        <div class="item single-line" ng-if="btx.action != 'received' && isShared">
           <span class="label" translate>Created by</span>
           <span class="item-note">
-            {{btx.creatorName}} <time>{{ (btx.ts || btx.createdOn ) * 1000 | amDateFormat:'MM/DD/YYYY hh:mm a'}}</time>
```
JDonadio: 👏

JDonadio:
Nice work!

Can you add the ability to change the fee value to unit/fiat?
The update tx (by clicking on confirmations) isn't intuitive, what about add the refresh icon at right-top?

cmgustavo:
Updated. Thanks!

PR #5740, opened 2017-03-08
Fix import QR code from main scan button

JDonadio:
Fixes https://github.com/bitpay/copay/issues/5738

Tested on Android and Browser (devel)


gabrielbazan7:
ACK tested on IOS

PR #5751, opened 2017-03-10
Fix Tx-defailts on iOS. Smoother transition

cmgustavo:
Replaced ng-show by ng-if to improve responsive on mobile


gabrielbazan7:
Code ACK

Issue #5773, opened 2017-03-15
Receive tab not showing QR


isocolsky:
When the active wallet in the receive tab is removed, no other wallet gets selected and there is no QR (see screenshot). The expected behavior would be to make another wallet the active wallet in that tab.

![image](https://cloud.githubusercontent.com/assets/2073468/23949174/069100fc-0965-11e7-91c9-da6418e22df5.png)

PR #5776, opened 2017-03-15
Set wallet properly - receive tab

JDonadio:
Fixes https://github.com/bitpay/copay/issues/5773

### DIFF
--- a/src/js/controllers/tab-receive.js
+++ b/src/js/controllers/tab-receive.js
@@ -34,12 +34,6 @@ angular.module('copayApp.controllers').controller('tabReceiveController', functi
     });
   };
 
-  $scope.loadAddresses = function(wallet, index) {
-    walletService.getAddress(wallet, false, function(err, addr) {
-      $scope.walletAddrs[wallet.id] = addr;
-    });
-  }
-
   $scope.goCopayers = function() {
     $ionicHistory.removeBackView();
     $ionicHistory.nextViewOptions({
@@ -88,9 +82,9 @@ angular.module('copayApp.controllers').controller('tabReceiveController', functi
     });
   };
 
-  $scope.setWallet = function(index) {
-    $scope.wallet = $scope.wallets[index];
-    $scope.walletIndex = index;
+  $scope.setWallet = function(wallet) {
+    $scope.wallet = wallet;
+    $scope.walletIndex = $scope.wallets.indexOf(wallet);
     if ($scope.walletAddrs[$scope.wallet.id].addr) $scope.addr = $scope.walletAddrs[$scope.walletIndex].addr;
     else $scope.setAddress(false);
   }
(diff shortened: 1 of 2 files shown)


src/js/controllers/tab-receive.js
```diff
@@ -127,26 +108,47 @@ angular.module('copayApp.controllers').controller('tabReceiveController', functi
...
+      $scope.wallet.status = status;
+      $scope.setAddress();
     });
   };
 
   $scope.$on("$ionicView.beforeEnter", function(event, data) {
     $scope.wallets = profileService.getWallets();
+    if (!$scope.wallets[0]) return;
+    if ($scope.wallet && findWallet($scope.wallet))
+      $scope.setWallet($scope.wallets.indexOf($scope.wallet));
```
isocolsky: Since you already have `$scope.wallet` why not just use it as the argument to `setWallet()`?

isocolsky: Even better: accept a wallet object as the argument to `setWallet()`

JDonadio:
Updated. thanks for your comments!

src/js/controllers/tab-receive.js
```diff
@@ -88,9 +82,9 @@ angular.module('copayApp.controllers').controller('tabReceiveController', functi
     });
   };
 
-  $scope.setWallet = function(index) {
-    $scope.wallet = $scope.wallets[index];
-    $scope.walletIndex = index;
+  $scope.setWallet = function(wallet) {
+    $scope.wallet = wallet;
+    $scope.walletIndex = $scope.wallets.indexOf(wallet);
     if ($scope.walletAddrs[$scope.wallet.id].addr) $scope.addr = $scope.walletAddrs[$scope.walletIndex].addr;
```
isocolsky: What would happen if `$scope.walletAddrs[$scope.wallet.id].addr` was `undefined`? I think you should reset `$scope.addr` first.

src/js/controllers/tab-receive.js
```diff
@@ -88,9 +82,9 @@ angular.module('copayApp.controllers').controller('tabReceiveController', functi
     });
   };
 
-  $scope.setWallet = function(index) {
-    $scope.wallet = $scope.wallets[index];
-    $scope.walletIndex = index;
+  $scope.setWallet = function(wallet) {
```
isocolsky: This looks better but the view is still using this function passing the wallet index: https://github.com/bitpay/copay/blob/master/www/views/tab-receive.html#L80

src/js/controllers/tab-receive.js
JDonadio: $scope.addr is cleaned in `setAddress` function se branch

src/js/controllers/tab-receive.js
JDonadio: You're right. Changed.

src/js/controllers/tab-receive.js
```diff
@@ -88,9 +82,9 @@ angular.module('copayApp.controllers').controller('tabReceiveController', functi
...
+    $scope.wallet = wallet;
+    $scope.walletIndex = $scope.wallets.indexOf(wallet);
```
JDonadio: Unnecesary, it's not used in the view. Chenged to `var walletIndex`

JDonadio:
Closed in favor of https://github.com/bitpay/copay/pull/5884

PR #5794, opened 2017-03-20
Show wallets with sufficient funds in order to top-up the bitpay card…

### LINKED ISSUE #5791: Card Loading Not Working
Debit card top-ups are not working for me in version 3.0.5 of BitPay app on Android 7.1.1. I’m taken to a confirmation screen and slide right, then I get a success screen. But no funds are broadcast, I am never prompted to select a wallet source for the top-up, and no transaction proposal is created.

Here are the steps I'm seeing now:

screenshot_20170318-173616

1. First, I'm not immediately prompted to choose a wallet to load from. I had to select Personal Wallet to view the wallet options. It excludes/doesn't show my other, very full wallet that I would normally use to load my card. There is no way to change this.

screenshot_20170318-173626

2. Confirming load. No warning is shown because funds from personal wallet were insufficient for the load amount. 

3. 
screenshot_20170318-173640

4. 
screenshot_20170318-173725

Funds are not added to the card, despite success state.

cmgustavo:
…. Improves error messages

Fixes #5791

### DIFF
--- a/src/js/services/bitpayCardService.js
+++ b/src/js/services/bitpayCardService.js
@@ -1,6 +1,6 @@
 'use strict';
 
-angular.module('copayApp.services').factory('bitpayCardService', function($log, $rootScope, $filter, lodash, storageService, bitauthService, platformInfo, moment, appIdentityService, bitpayService, nextStepsService, configService, txFormatService, appConfigService) {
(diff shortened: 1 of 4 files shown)


src/js/services/bitpayCardService.js
```diff
@@ -59,6 +61,7 @@ angular.module('copayApp.services').factory('bitpayCardService', function($log,
     return {
       amount: amount, 
       currency: currency, 
+      amountSat: amountSat,
```
matiu: OK. approx amount

matiu:
Good job!

PR #5797, opened 2017-03-21
Baseline Intel TEE support

ajp8164:
Relies on a simple BWC change and Intel software to drive TEE chipset.

### DIFF
--- a/src/js/controllers/import.js
+++ b/src/js/controllers/import.js
@@ -35,7 +35,7 @@ angular.module('copayApp.controllers').controller('importController',
       if ($scope.isChromeApp || $scope.isDevel) {
         $scope.seedOptions.push({
           id: walletService.externalSource.trezor.id,
-          label: walletService.externalSource.ledger.longName,
+          label: walletService.externalSource.trezor.longName,
         });
         $scope.formData.seedSource = $scope.seedOptions[0];
       }
--- a/src/js/controllers/import.js
+++ b/src/js/controllers/import.js
@@ -3,6 +3,8 @@
 angular.module('copayApp.controllers').controller('importController',
   function($scope, $timeout, $log, $state, $stateParams, $ionicHistory, $ionicScrollDelegate, profileService, configService, sjcl, ledger, trezor, derivationPathHelper, platformInfo, bwcService, ongoingProcess, walletService, popupService, gettextCatalog, appConfigService) {
 
+    var isChromeApp = platformInfo.isChromeApp;
+    var isDevel = platformInfo.isDevel;
     var reader = new FileReader();
     var defaults = configService.getDefaults();
     var errors = bwcService.getErrors();
@@ -25,15 +27,15 @@ angular.module('copayApp.controllers').controller('importController',
 
       if ($scope.isChromeApp) {
         $scope.seedOptions.push({
-          id: 'ledger',
-          label: 'Ledger Hardware Wallet',
+          id: walletService.externalSource.ledger.id,
+          label: walletService.externalSource.ledger.longName,
         });
       }
 
       if ($scope.isChromeApp || $scope.isDevel) {
         $scope.seedOptions.push({
-          id: 'trezor',
-          label: 'Trezor Hardware Wallet',
+          id: walletService.externalSource.trezor.id,
+          label: walletService.externalSource.ledger.longName,
         });
         $scope.formData.seedSource = $scope.seedOptions[0];
       }
@@ -267,7 +269,7 @@ angular.module('copayApp.controllers').controller('importController',
           return;
         }
 
-        lopts.externalSource = 'trezor';
+        lopts.externalSource = walletService.externalSource.trezor.id;
         lopts.bwsurl = $scope.formData.bwsurl;
         ongoingProcess.set('importingWallet', true);
         $log.debug('Import opts', lopts);
@@ -293,7 +295,7 @@ angular.module('copayApp.controllers').controller('importController',
 
       var account = $scope.formData.account;
 
-      if ($scope.formData.seedSource.id == 'trezor') {
+      if ($scope.formData.seedSource.id == walletService.externalSource.trezor.id) {
         if (account < 1) {
           popupService.showAlert(gettextCatalog.getString('Error'), gettextCatalog.getString('Invalid account number'));
           return;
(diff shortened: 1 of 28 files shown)


src/js/routes.js
```diff
@@ -579,6 +579,15 @@ angular.module('copayApp').config(function(historicLogProvider, $provide, $logPr
           }
         }
       })
+      .state('tabs.preferences.preferencesExternal', {
```
JDonadio: I don't see the point here for this view. I connected/disconnected both devices and always have "No hardware information available". Is this necessary for trezor or ledger?

src/js/controllers/preferencesExternal.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+angular.module('copayApp.controllers').controller('preferencesExternalController', function($scope, $stateParams, lodash, gettextCatalog, popupService, profileService, walletService) {
+  var wallet = profileService.getWallet($stateParams.walletId);
+  $scope.wallet = wallet;
```
JDonadio: Not needed

src/js/controllers/preferences.js
```diff
@@ -89,9 +89,6 @@ angular.module('copayApp.controllers').controller('preferencesController',
         value: $scope.wallet.balanceHidden
       };
 
-      if (wallet.isPrivKeyExternal)
-        $scope.externalSource = wallet.getPrivKeyExternalSourceName() == 'ledger' ? 'Ledger' : 'Trezor';
```
JDonadio: This option is not used in the view, however we must hide all security options from waller preferences if it is an external source.

src/js/controllers/import.js
```diff
@@ -25,15 +27,15 @@ angular.module('copayApp.controllers').controller('importController',
...
       if ($scope.isChromeApp) {
         $scope.seedOptions.push({
-          id: 'ledger',
-          label: 'Ledger Hardware Wallet',
+          id: walletService.externalSource.ledger.id,
+          label: walletService.externalSource.ledger.longName,
         });
       }
 
       if ($scope.isChromeApp || $scope.isDevel) {
```
JDonadio: Same as above

src/js/controllers/import.js
```diff
@@ -25,15 +27,15 @@ angular.module('copayApp.controllers').controller('importController',
 
       if ($scope.isChromeApp) {
```
JDonadio: Should compare as https://github.com/bitpay/copay/pull/5797/files#diff-c5f2d6282049bf9d04bcb177c5912451R64

JDonadio:
Trezor and Ledger tested > OK

src/js/routes.js
ajp8164: It is necessary for the Intel platform.

src/js/controllers/preferencesExternal.js
ajp8164: Is needed.

src/js/controllers/preferences.js
ajp8164: Sorry, I don't understand.

src/js/controllers/preferencesExternal.js
JDonadio: $scope.wallet is not used either in the view or controller.

src/js/controllers/preferences.js
JDonadio: copay_-_copay_bitcoin_wallet_y_preferencesexternal_html_ ___copay

src/js/routes.js
JDonadio: Should be hidden for trezor and leger? (most users)

src/js/controllers/preferencesExternal.js
ajp8164: Oh. Thought you were referring to the whole file ;-)
Yes, if it's not referenced it should be removed.

JDonadio: Just the $scope.wallet reference :)

src/js/controllers/import.js
ajp8164: ?

src/js/controllers/import.js
ajp8164: Chrome or dev?  I recall this being exactly what was in the baseline I started with but i can add `|| isDevel`.  Is that what you're asking about?

src/js/controllers/preferences.js
ajp8164: "More options" should be there but not "Security" ?

src/js/controllers/import.js
JDonadio: In join.js you're comparing agains `walletService.externalSource.ledger.supported` and `walletService.externalSource.trezor.supported` I'm just wondering if you should use the same method in create.js.

ajp8164: oh. yes. good catch!

src/js/routes.js
JDonadio: Would be nice to add L&T information there, but for now maybe a conditional to hide it will avoid the L&T users confusing or asking "why the information is not available if I have my hardware connected" (like me lol)

src/js/controllers/preferences.js
JDonadio: Yep, security options are available for wallets which have the privkey locally not externally. So these wallets cannot sign, encrypt, request fingerprint, access to 12 words, etc

src/js/controllers/import.js
JDonadio: Sorry, I mean the `alletService.externalSource.ledger.supported` comparison

src/js/routes.js
ajp8164: Done

src/js/controllers/preferencesExternal.js
ajp8164: Done

src/js/controllers/preferences.js
ajp8164: Done

src/js/controllers/import.js
ajp8164: Done

src/js/controllers/import.js
ajp8164: Done

PR #5841, opened 2017-03-29
Using new version of connect

runn1ng:
Old version of connect is not correctly signing transactions with version v2.


matiu:
Thanks a lot! We will review it soon

matiu:
Code ACK.

JDonadio:
ACK

JDonadio:
Thanks @runn1ng!

PR #5852, opened 2017-04-05
temporary fix for the tabs confusion

gabrielbazan7:
This is a temporary fix to avoid the users to copy a wrong address when they enter to the wallet details of a wallet and then enter to the receive tab.

Confusion: 
```
So I click on the wallet in the home screen
Then I click on "Receive" at the bottom
It shows me an address. But this is the address for a TOTALLY DIFFERENT WALLET.
```

### DIFF
--- a/www/views/walletDetails.html
+++ b/www/views/walletDetails.html
@@ -1,4 +1,4 @@
-<ion-view id="walletDetails">
+<ion-view id="walletDetails" hide-tabs>
 
   <ion-nav-bar ng-class="{'wallet-background-color-default': !wallet.color}" ng-style="{'background-color': wallet.color}">
     <ion-nav-title>{{wallet.name}}</ion-nav-title>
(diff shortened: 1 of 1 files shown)


JDonadio:
LGTM. Hiding tabs will prevent to assume they has dependencies each other.

bitjson:
Awesome 👍

toomim:
Does someone have a screenshot handy of the new UI? Without running this code, I'm guessing that it makes this change:

- From the home screen, if you click on a wallet
- There will no longer be tabs available at the bottom of the screen

If so, this is certainly an improvement to prevent the most egregious part of the error. I'm looking forward to any help I can give for the deeper redesign.

JDonadio:
Before this change
copay - copay bitcoin wallet - google chrome

After this change
copay - copay bitcoin wallet - google chrome_2

toomim:
Great, thanks! I think this is an insightful quick fix, and look forward to deeper redesign.

PR #5858, opened 2017-04-06
fixing desktop redirect from starting to home view

gabrielbazan7:


JDonadio:
Goot catch @gabrielbazan7

Issue #5693, opened 2017-03-01
Low fee unconfirmed incoming payment insufficient info on screen


fortran77:
A wallet configured in Copay received a small payment with a very low fee.  I clicked on the displayed line for this incoming payment (using the Copay Chrome app) and then clicked on the SEE IT ON THE BLOCKCHAIN button. My web browser was directed towards a long URL on insight.bitpay.com and then immediately redirected to the top-level page [https://insight.bitpay.com/](https://insight.bitpay.com/) where I saw an error message "Transaction Not Found". On that screen I saw no other transaction-specific information other than this general error message.

From the naive end-user's point of view, here are the problems.

1. If the Copay wallet shows an incoming transaction, then the transaction must exist. Viewing it on the blockchain should show the transaction. (So the user would think.)
2. Since Copay gives us no option to use a different blockchain explorer, the hard-coded blockchain explorer should provide more information. It should at least show the transaction id for the transaction that was not found. The error message ought to be "Transaction xxxxxxxxxxxxx not found" where the transaction ID is included in the error message. Now at least the user can make further attempts to track down this existing-but-not-found transaction.

I happen to have the same 12-word seed for this Copay wallet also configured in the Electrum wallet. Electrum listed the transaction as "⚠ Low fee" (a useful hint that Copay did not provide) and when I right-clicked and asked for Details it gave me the transaction id and the values of two outputs, and the helpful text: "Expected confirmation time: unknown (low fee)".

It would be desirable for Copay to give the user more information in such cases. Not necessarily as much as Electrum does (although that would be good too), but more than just a generic error message "Transaction Not Found".

matiu:
Hey, 

Great points, thanks for reporting it. Yes, we will definitely add the (!low fee) indication and enhance the handling of the UI if the TX is not found in Insight.

thanks a lot.

PR #5875, opened 2017-04-12
Feat/Add warning for low fee

JDonadio:
Fixes https://github.com/bitpay/copay/issues/5693

copay_-_copay_bitcoin_wallet

copay_-_copay_bitcoin_wallet


www/views/tx-details.html
```diff
@@ -84,6 +84,10 @@
             <span ng-if="toggleFeeFiat">{{feeFiatStr}}</span>
           </span>
         </div>
+        <div class="item low-fees" ng-if="btx.action == 'received' && btx.lowFees">
```
gabrielbazan7: also : `btx.confirmations === 0`

dabura667: I agree `btx.confirmations === 0` should also be a condition.

gabrielbazan7: Ok. It's done here https://github.com/bitpay/bitcore-wallet-service/pull/650

isocolsky:
ACK

Issue #5239, opened 2016-12-15
Suggestion: Proposal list animation


JDonadio:
Proposal list/cards should appear with a smooth animation. In the current version it's anything but smooth (see the video).

https://youtu.be/Y8md1D6UduA

PR #5886, opened 2017-04-14
adding loading placeholders to avoid the screen resize in tab-home view

gabrielbazan7:


JDonadio:
https://github.com/bitpay/copay/issues/5239

PR #5910, opened 2017-04-20
Request PIN to disable it

JDonadio:
Fixes https://github.com/bitpay/copay/issues/5905

* Disable unselected option
* Remove popup confirmation

### DIFF
--- a/src/js/controllers/lockSetup.js
+++ b/src/js/controllers/lockSetup.js
@@ -12,6 +12,7 @@ angular.module('copayApp.controllers').controller('lockSetupController', functio
         method: 'pin',
         label: gettextCatalog.getString('Lock by PIN'),
         needsBackup: null,
+        disabled: null,
       },
     ];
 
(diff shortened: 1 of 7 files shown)


src/js/controllers/lockSetup.js
```diff
@@ -20,22 +21,38 @@ angular.module('copayApp.controllers').controller('lockSetupController', functio
...
-    });
+    checkAndSelectOption();
     processWallets();
   };
 
   $scope.$on("$ionicView.beforeEnter", function(event) {
     init();
   });
 
+  function checkAndSelectOption() {
```
matiu: name should be something like:

setMethodSelector
or
configureMethodSelect

src/js/controllers/lockSetup.js
```diff
@@ -73,39 +90,28 @@ angular.module('copayApp.controllers').controller('lockSetupController', functio
...
-    else if (selectedMethod == 'fingerprint') {
-      if (savedMethod == 'pin') {
-        askForDisablePin(function(disablePin) {
-          if (disablePin) saveConfig('fingerprint');
-          else init();
-        });
-      } else saveConfig('fingerprint');
+    if (!selectedMethod) {
+      if (!savedMethod) return;
+      if (savedMethod == 'pin') $state.transitionTo('tabs.pin', {
```
matiu: It dont follow how after transitioning to tab.pin, the new  (null) `selectedMethod` will be saved.

Also, disabling fingerprint should also ask for the fingerprint.

JDonadio: Because I'm telling to the next view (pin) that the action is `unlocking` with:

> {fromSettings: true, locking: false}

src/js/controllers/lockSetup.js
JDonadio: Done. Changed to `setMethodSelector`

PR #5915, opened 2017-04-21
Translation/updates po pt

cmgustavo:


dabura667:
There were a few more translations added to crowdin about PIN lock / Finger lock etc.

Those are translated in Japanese now.

PR #6104, opened 2017-05-23
improvements on the card UI

gabrielbazan7:
Still missing "View Confirmation Status" label. The transaction id is not coming on the invoice request.  

bitpay_card

image_uploaded_from_ios__15__jpg_and_downloads


www/views/bitpayCard.html
```diff
@@ -7,7 +7,7 @@
 
   <ion-content>
 
-    <div class="box-notification warning m0" ng-show="network == 'testnet'">
+    <div class="box-notification warning m0" ng-show="network == 'testnet'" translate>
```
cmgustavo: Not necessary `translate` here.

gabrielbazan7:
Updated

PR #6116, opened 2017-05-26
Problem opening Copay on Android from another Cordova app

plotnik:
Hello,

We've noticed a problem opening Copay on Android from another Cordova app 
when using bitpay/cordova-sdk. Copay is opened within the same window as the calling app, 
making hard to switch back to the calling app. The similar problem is described at
https://stackoverflow.com/questions/28058633/deep-linking-into-a-cordova-app-from-a-web-browser-on-android-opens-app-within-t

Regards, Egor

### DIFF
--- a/app-template/config-template.xml
+++ b/app-template/config-template.xml
@@ -11,7 +11,7 @@
         BitPay Inc.
     </author>
     <content src="index.html" />
-    <access launch-external="yes" origin="*" />
+    <access origin="*" />
     <preference name="AndroidLaunchMode" value="singleTask" />
     <preference name="AndroidPersistentFileLocation" value="Internal" />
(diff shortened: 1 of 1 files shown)


gabrielbazan7:
I get Wallet Service not found, if I remove `launch-external="yes" ` works as expected.

plotnik:
You are right, `access launch-external="yes"` not needed here.

JDonadio:
Thanks @plotnik !

gabrielbazan7:
Hey! I'm reverting this one. We are sending an urgent release now with a UI fix. We need more time to test all the services Glidera, Conbaise, Bitpay Card, payment protocol and bit21 with this PR. We could merge it later.

PR #6161, opened 2017-06-01
Adds support for EUR and GBP currencies

cmgustavo:


Kirvx:
Sentences should also be modified in the Card Intro to include the other currencies:
https://github.com/bitpay/copay/blob/master/www/views/bitpayCardIntro.html

PR #6191, opened 2017-06-06
Feat - Check transaction params on third party block explorers

### LINKED ISSUE #5994: Check incoming transactions on other blockexplorers
To mitigate possible attacks is Insight is compromised,  recheck incoming transactions in TX detail views in different public block explorers, like https://blockr.io/.

If the TX is reported confirmed by Insight, and it does not appear at other block explorer: Show an important warning.

If the TX is reported unconfirmed, just a minor warning.

if the TX is at other block explorer, and match (amount, destination) add a Check ("also seen at xxxxxxx.com").

JDonadio:
fixes #5994


src/js/services/blockExplorerService.js
```diff
@@ -0,0 +1,83 @@
...
+    var request = {
+      url: url,
+      method: 'GET',
+      json: true
+    };
+
+    $http(request).then(function(resp) {
+      $log.debug('Tx params:', resp);
+      return cb(null, resp);
+    }, function(err) {
```
matiu: We should separate the err here:

if (err ... = 404) then return cb(null, { notFound: true})
else return(err).

src/js/controllers/tx-details.js
```diff
@@ -1,20 +1,36 @@
 'use strict';
```
matiu: There is way to much code in tx-details. js regarding this feature. it should be only a couple of lines.

src/js/controllers/tx-details.js
```diff
@@ -36,6 +54,69 @@ angular.module('copayApp.controllers').controller('txDetailsController', functio
...
+    $scope.loadingService[name] = state;
+    $timeout(function() {
+      $scope.$apply();
+    });
+  };
+
+  function verifyByThirdParties() {
+    if (lodash.isEmpty($scope.availableServices) || $scope.wallet.network == 'testnet') return;
+
+    lodash.each($scope.availableServices, function(service) {
```
matiu: this loop should be inside the service.

src/js/controllers/tx-details.js
```diff
@@ -36,6 +54,69 @@ angular.module('copayApp.controllers').controller('txDetailsController', functio
...
+      setLoading(service.name, true);
+
+      verifyByThirdPartyService.getTx(service, txId, function(err, params) {
```
matiu: The service api should be something like:


thirdPartyExplorers.verify(tx, function(err, result) {
   if (err) return ...;

   $scope.verifiedOK = result.ok;
   $scope.verifiedNOK = result.nok;
   $scope.verifiedNA = result.na;
});

src/js/controllers/tx-details.js
```diff
@@ -36,6 +54,69 @@ angular.module('copayApp.controllers').controller('txDetailsController', functio
...
+          checkError();
+        } else
+          compareAndUpdateParams(params);
+
+        serviceCounter += 1;
+      });
+    });
+  };
+
+  function compareAndUpdateParams(params) {
```
matiu: inside the service.

Please do not use `params`. Parameters are a function input.  It is a generic name.

src/js/services/verifyByThirdPartyService.js
```diff
@@ -0,0 +1,91 @@
...
+  var getTxFromBlockchainInfo = function(txid, cb) {
+    var service = lodash.find(configService.getDefaults().blockExplorerServices, function(s) {
+      return (s.name == 'Blockchain.info');
+    });
+
+    requestTx(service.name, service.url + txid, function(err, resp) {
+      if (err) return cb(err);
+      if (resp.notFound) return cb(null, resp);
+
+      var params = {};
```
matiu: please dont use `params` name  as a response.

src/js/services/verifyByThirdPartyService.js
```diff
@@ -0,0 +1,91 @@
...
+  var explorers = {
+    'Blockchain.info': {
+      getTx: getTxFromBlockchainInfo,
+    },
+    'Blockr.io': {
+      getTx: getTxFromBlockrIo,
+    }
+  };
+
+  function requestTx(serviceName, url, cb) {
```
matiu: why parameter serviceName here?

src/js/services/configService.js
```diff
@@ -68,6 +68,21 @@ angular.module('copayApp.services').factory('configService', function(storageSer
       enabled: false,
     },
 
+    blockExplorerServices: [
```
matiu: the service listing  suggest that adding a service is just adding a new item in the config, and it is not.

src/js/services/verifyByThirdPartyService.js
```diff
@@ -0,0 +1,91 @@
...
+        params.out.push({
+          address: out.address,
+          amount: parseInt(out.amount * unitToSatoshi)
+        });
+      });
+      return cb(null, params);
+    });
+  };
+
+  var explorers = {
```
matiu: here there shoudl be all the info about the services:


name,
url,
parser: function(response) {
  return normalizedJason;
}

JDonadio:
Updated. Thanks for your comments!

www/views/advancedSettings.html
```diff
@@ -25,6 +25,15 @@
 
       <div class="item item-divider"></div>
 
+      <ion-toggle class="has-comment" ng-model="verifyTransactionEnabled.value" toggle-class="toggle-balanced" ng-change="verifyTransactionChange()">
+        <span class="toggle-label" translate>Verify transaction</span>
+      </ion-toggle>
+      <div class="comment" translate>
+        If enabled, each transaction will be automatically verified by third party Block Explorer.
```
matiu: Block Explorer*s*

JDonadio: Updated. thanks!

Issue #5771, opened 2017-03-14
Windows scanner build issue


gabrielbazan7:
- Steps to make the scanner works 
1. npm run apply:bitpay
2. npm run start:windows
3. cd plugins/cordova-plugin-qrscanner
4. remove ^ from webpack dependency on package.json (https://github.com/bitpay/cordova-plugin-qrscanner/pull/78) 
5. npm i
6. npm run build
7. Go to root
8. rm -rf platforms/windows
9. cordova platform add windows
10. npm run start:windows
11. add to cordova project index.html file:

`<script type="text/javascript" src="plugins/cordova-plugin-qrscanner/www/www.min.js"></script>`

12. Check windows 10 manifest file and add:
`<DeviceCapability Name="webcam" />`

PR #6195, opened 2017-06-07
add old scanner for windows platform

gabrielbazan7:


Derrick-:
Does this cover all the remaining requirements for a WP compatible release?
Is there an issue that covers the WP problems?

gabrielbazan7:
@Derrick- we were trying to use the BitPay scanner (https://github.com/bitpay/cordova-plugin-qrscanner)  without success. Here is an issue related to that https://github.com/bitpay/copay/issues/5771. So we are using the old one (https://github.com/phonegap/phonegap-plugin-barcodescanner). I think we are ready to lunch the app for windows 10

Derrick-:
> I think we are ready to lunch the app for windows 10
Thanks for your efforts on this. It's been a long patient wait here for a WP10 upgrade.

PR #6197, opened 2017-06-07
Unifies card history view with wallet view

ajp8164:
- Makes the card history view look the same (similar) to the wallet history view.
- Relative or absolute tx time (less than a day = relative).
- Removes the running balance presentation (UI change only, logic still in place).
- Introduces timeService to share common functions between wallet and card views.
- Enhances svg directive to work with images using ngSrc.
- Adds card `country` to the cards collection (not used at the moment, will need for future policy statements in the app).

screen shot 2017-06-07 at 6 54 43 pm

### DIFF
--- a/src/js/controllers/bitpayCard.js
+++ b/src/js/controllers/bitpayCard.js
@@ -163,7 +163,7 @@ angular.module('copayApp.controllers').controller('bitpayCardController', functi
     runningBalance -= parseFloat(tx.amount);
   };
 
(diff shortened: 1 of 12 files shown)


tonygallippi:
relative time has been fixed, icons resized, and affected styles to top-ups fixed.  I like it.

vanhue:
Could we arrange the transaction icons to look a little more like the wallet? Example:

screenshot 2017-06-08 11 44 12

PR #6259, opened 2017-06-16
Bug/fix card and send views

ajp8164:
Fix the send search background color on blur.
Fix the card mcc icons so they color properly (circle).
Fix the card pre-auth header to include a (?) link and eliminate the verbose text explanation (same with confirming).
Fix pre-auth present-header logic to choose the proper header when a pre-auth is detected.

### DIFF
--- a/src/js/controllers/bitpayCard.js
+++ b/src/js/controllers/bitpayCard.js
@@ -92,10 +92,10 @@ angular.module('copayApp.controllers').controller('bitpayCardController', functi
           return !tx.pending && tx.type.indexOf('93') == -1;
         });
         self.bitpayCardTransactionHistoryConfirming = lodash.filter(txs, function(tx) {
-          return tx.pending;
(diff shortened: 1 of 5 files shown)


src/js/controllers/bitpayCard.js
```diff
@@ -92,10 +92,10 @@ angular.module('copayApp.controllers').controller('bitpayCardController', functi
           return !tx.pending && tx.type.indexOf('93') == -1;
         });
         self.bitpayCardTransactionHistoryConfirming = lodash.filter(txs, function(tx) {
-          return tx.pending;
+          return tx.pending && tx.type.indexOf('93') == -1;
         });
-        self.bitpayCardTransactionHistoryPreAuth = lodash.includes(txs, function(tx) {
-          return tx.type.indexOf('93') > -1;
+        self.bitpayCardTransactionHistoryPreAuth = lodash.filter(txs, function(tx) {
```
gabrielbazan7: +1

PR #6316, opened 2017-06-27
Ref topup. Fix sendmax

cmgustavo:


cmgustavo:
Updated. Ready for review

src/js/services/txFormatService.js
```diff
@@ -23,6 +23,26 @@ angular.module('copayApp.services').factory('txFormatService', function($filter,
     return root.formatAmount(satoshis) + ' ' + config.unitName;
   };
 
+  root.formatToCode = function(satoshis, code, cb) {
```
matiu: maybe a better name would be "toFiat"

src/js/controllers/topup.js
```diff
@@ -50,29 +60,146 @@ angular.module('copayApp.controllers').controller('topUpController', function($s
...
+        return cb({
+          title: gettextCatalog.getString('Could not create transaction'),
+          message: bwcError.msg(err)
+        });
+      }
+      return cb(null, ctxp);
+    });
+  };
+
+  var parseAmount = function(wallet, cb) {
```
matiu: Maybe this function should be `calculateAmount`...

```
if (useSendMax)  {
  ...
} else 
  return cb(null, amount);
```

and on the result of the function use `txFormatService.parseAmount`

cmgustavo: Good idea. Thanks

cmgustavo:
Updated.

Issue #5597, opened 2017-02-09
On settings view with no way out


ajp8164:
![img_5543](https://cloud.githubusercontent.com/assets/1063765/22767545/f813955c-ee49-11e6-9962-f6b6e4ab9eed.PNG)

Opened app, went to settings>about>session log>back and not tab menu.

Update: tab menu appeared after resuming app from background.  Still an issue though ;-)

Update2: backgrounding the app and resuming once again and the tab menu is gone again.  Can't get it back now.

JDonadio:
v1.3.6 right?

ajp8164:
Yes, 1.3.6.
---
Andy Phillipson
Director of Engineering Services, BitPay, Inc.
844A CCA5 F277 30C0 E86F  9799 0CEE 502F 2484 0854

JDonadio:
I was able to reproduce in v1.3.6 - Android in the related issue above. I'm not sure why it's happening in your case but definitely has to be the redrawing of the view.

JDonadio:
Waiting for https://github.com/bitpay/copay/pull/5864

gabrielbazan7:
Is this issue solve in the recent version ?

JDonadio:
I think no. Probably the same problem as https://github.com/bitpay/copay/issues/6482#issuecomment-318717623

gabrielbazan7:
Ok. Lets close this one in order to #6482

PR #6322, opened 2017-06-28
avoid flickers on transitions and fixes hidden tab bug

### LINKED ISSUE #6262: Enter amount view init jumpy
When going from Send view to Enter amount view the Enter amount view init causes the view to expand/jump down. Issue between the "Amount" text and the "0.00" value text.

gabrielbazan7:
fixes #6262 #5597


cmgustavo:
Missing coinbase.html

PR #6323, opened 2017-06-28
Update translations

cmgustavo:


JDonadio:
`"No wallets with funds"` Should be `No wallets available`?

cmgustavo:
Ok. Agree. Updated

PR #6408, opened 2017-07-13
Confirmed transactions notifications as global settings

### LINKED ISSUE #6401: Notifications when confirm
Make this option available by default on settings.
And maybe the user could get a second notification when the txs reaches 6 confirmations

Gamboster:
fixes #6401

### DIFF
--- a/src/js/services/configService.js
+++ b/src/js/services/configService.js
@@ -80,7 +80,7 @@ angular.module('copayApp.services').factory('configService', function(storageSer
 
     pushNotificationsEnabled: true,
 
-    notifyIfTxConfirmed: {
+    confirmedTxsNotifications: {
       enabled: true,
     },
 
--- a/src/js/services/configService.js
+++ b/src/js/services/configService.js
@@ -80,6 +80,10 @@ angular.module('copayApp.services').factory('configService', function(storageSer
 
     pushNotificationsEnabled: true,
 
+    notifyIfTxConfirmed: {
+      enabled: true,
+    },
+
     emailNotifications: {
       enabled: false,
     },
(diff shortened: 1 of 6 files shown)


www/views/tx-details.html
```diff
@@ -108,13 +108,15 @@
...
           </span>
         </div>
-        <ion-toggle ng-show="!btx.confirmations || btx.confirmations == 0"
-                    class="toggle-unconfirmed"
-                    ng-model="txNotification.value"
-                    toggle-class="toggle-balanced"
-                    ng-change="txConfirmNotificationChange()">
-          <span class="toggle-label" translate>Notify me if confirmed</span>
-        </ion-toggle>
+        <div ng-if="!notifyIfTxConfirmed">
```
JDonadio: Should be "txsUnsubscribedForNotifications"

www/views/preferencesNotifications.html
```diff
@@ -19,6 +19,12 @@
         </div>
       </div>
 
+      <div ng-if="PNEnabledByUser">
+        <ion-toggle ng-model="notifyIfTxConfirmed.value" toggle-class="toggle-balanced" ng-change="notifyIfTxConfirmedChange()" ng-if="usePushNotifications" ng-show="pushNotifications.value">
```
JDonadio: should not use `ng-if` and `ng-show` in the same component

src/js/services/configService.js
```diff
@@ -80,6 +80,10 @@ angular.module('copayApp.services').factory('configService', function(storageSer
 
     pushNotificationsEnabled: true,
 
+    notifyIfTxConfirmed: {
```
JDonadio: Should be "confirmedTxsNotifications: {"

PR #6418, opened 2017-07-14
Updates pot file

cmgustavo:


i18n/po/template.pot
```diff
@@ -3104,6 +3221,14 @@ msgstr ""
 msgid "Your bitcoin wallet is ready!"
 msgstr ""
 
+#: www/views/modals/chooseFeeLevel.html:38
+msgid "Your could not set a fee higher than 1000 satoshis/byte."
```
JDonadio: Typo: "your"

i18n/po/template.pot
```diff
@@ -3104,6 +3221,14 @@ msgstr ""
...
+msgstr ""
+
+#: www/views/modals/chooseFeeLevel.html:34
+msgid "Your fee is lower than the recommended super economy fee ({{getMinimumRecommeded()}} satoshis/byte). Your transaction may never get confirmed."
```
JDonadio: Should be: "lower than recommended..." and "The transaction may never get confirmed"

i18n/po/template.pot
cmgustavo: Thanks!

i18n/po/template.pot
cmgustavo: Thanks!

cmgustavo:
Updated.

PR #6548, opened 2017-08-09
Join items "network" and "invoice" in an unique "miner fee" item

cmgustavo:
It is applied in:

* Top up
* Buy gift card


cmgustavo:
Closed in favor of #6576

PR #6684, opened 2017-09-01
Bug/send max fee levels

cmgustavo:
* Fix sendMax, support bch. Fee levels cache for bch

### DIFF
--- a/src/js/controllers/confirm.js
+++ b/src/js/controllers/confirm.js
@@ -28,7 +28,7 @@ angular.module('copayApp.controllers').controller('confirmController', function(
   function refresh() {
     $timeout(function() {
       $scope.$apply();
-    }, 1);
+    }, 10);
   }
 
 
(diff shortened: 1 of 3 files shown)


src/js/controllers/confirm.js
```diff
@@ -69,15 +69,15 @@ angular.module('copayApp.controllers').controller('confirmController', function(
 
   $scope.$on("$ionicView.beforeEnter", function(event, data) {
 
-    function setWalletSelector(coin, network, minAmount, cb) {
+    function setWalletSelector(coin, network, minAmount, sendMax, cb) {
```
matiu: maybe better to refacto and send coin ==null  or send  the whole `tx`

cmgustavo: Good idea. Thanks

cmgustavo:
Updated.

PR #6718, opened 2017-09-08
Feat/bitcoin cash url support

cmgustavo:
Also https://github.com/cmgustavo/Custom-URL-scheme has been updated.

### DIFF
--- a/src/js/services/incomingData.js
+++ b/src/js/services/incomingData.js
@@ -69,8 +69,8 @@ angular.module('copayApp.services').factory('incomingData', function($log, $stat
       }, 100);
     }
     // data extensions for Payment Protocol with non-backwards-compatible request
-    if ((/^bitcoin[cash]*:\?r=[\w+]/).exec(data)) {
-      data = decodeURIComponent(data.replace(/bitcoin[cash]*:?r=/, ''));
+    if ((/^bitcoin(cash)?:\?r=[\w+]/).exec(data)) {
+      data = decodeURIComponent(data.replace(/bitcoin(cash)?:\?r=/, ''));
       $state.go('tabs.send', {}, {
         'reload': true,
         'notify': $state.current.name == 'tabs.send' ? false : true
@@ -85,9 +85,9 @@ angular.module('copayApp.services').factory('incomingData', function($log, $stat
     data = sanitizeUri(data);
 
     // Bitcoin or Bitcoin Cash URL
-    if ((/^bitcoin[cash]*:/).exec(data)) {
+    if ((/^bitcoin(cash)?:/).exec(data)) {
       var coin = 'btc';
-      if ((/^bitcoincash*:/).exec(data)) {
+      if ((/^bitcoincash:/).exec(data)) {
         coin = 'bch';
         data = data.replace(/bitcoincash*:/, 'bitcoin:');
       }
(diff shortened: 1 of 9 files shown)


src/js/services/incomingData.js
```diff
@@ -84,27 +84,37 @@ angular.module('copayApp.services').factory('incomingData', function($log, $stat
 
     data = sanitizeUri(data);
 
-    // BIP21
-    if (bitcore.URI.isValid(data)) {
-      var parsed = new bitcore.URI(data);
+    // Bitcoin or Bitcoin Cash URL
+    if ((/^bitcoin[cash]*:/).exec(data)) {
```
dabura667: this means: "match a string that starts with "bitcoin" then may or may not have 0-infinity repetitions of any of the following letters ["c","a","s","h"] then a colon.

Notice what matches and what doesn't:
http://refiddle.com/refiddles/59b3564b75622d096f6d0000

dabura667: This is the correct regex: `/^bitcoin(cash)*:/` parenthesis instead of brackets

http://refiddle.com/refiddles/59b3569075622d0967760000

src/js/services/incomingData.js
```diff
@@ -84,27 +84,37 @@ angular.module('copayApp.services').factory('incomingData', function($log, $stat
...
+      var coin = 'btc';
+      if ((/^bitcoincash*:/).exec(data)) {
```
dabura667: This is also incorrect... this means "starts with bitcoincas and has zero to infinite "h" repetitions then a colon.
http://refiddle.com/refiddles/59b3575075622d0967780000

This should just be `/^bitcoincash:/`

src/js/services/incomingData.js
```diff
@@ -84,27 +84,37 @@ angular.module('copayApp.services').factory('incomingData', function($log, $stat
...
+        coin = 'bch';
+        data = data.replace(/bitcoincash*:/, 'bitcoin:');
```
dabura667: same as above

src/js/services/incomingData.js
matiu: +1

www/views/customAmount.html
```diff
@@ -21,7 +21,7 @@
   <ion-content scroll="false">
     <div class="address" ng-if="address && amountBtc">
       <div class="qr-code" copy-to-clipboard="copyToClipboard()">
-        <qrcode size="220" data="bitcoin:{{address + '?amount=' + amountBtc + '&coin=' + coin}}" color="#334"></qrcode>
+        <qrcode size="220" data="bitcoin{{ wallet.coin == 'bch' ? 'cash' : '' }}:{{address + '?amount=' + amountBtc}}" color="#334"></qrcode>
```
matiu: I think it would be better to have a fiunciton in walletService:

function getProtocolHandler(wallet) {
  if wallet.coin== 'bch' return 'bitcoincash';
  if wallet.coin=='btc' return 'bitcoin';
  <error>
}

and pass that info to the view, and do not put 'bitcoincash' in any other place.

src/js/services/incomingData.js
cmgustavo: Fixed. Thanks for review.

src/js/services/incomingData.js
cmgustavo: Updated. Thanks

www/views/customAmount.html
cmgustavo: Updated. Thanks.

PR #6735, opened 2017-09-14
Feat: basic receive and custom amount implemented

Gamboster:


src/providers/persistence/persistence.ts
```diff
@@ -52,7 +52,7 @@ export let persistenceProviderFactory = (platform: PlatformProvider, log: Logger
   }
    */
   // Testing in RAM
-  storage = new RamStorage(log);
+  storage = new LocalStorage(log);
```
cmgustavo: Leave this as RamStorage please.

Gamboster: Fixed, ty

Issue #6825, opened 2017-09-30
Please change title of app and clarify BCH as a separate currency.


dabura667:
Currently, Copay is registered as "A Bitcoin Wallet" which is misleading. Currently Copay now supports currencies other than Bitcoin, so "A Cryptocurrency Wallet" or "A Multi-currency Wallet" should be more appropriate.

Calling Copay "A Bitcoin Wallet" and adding an option for wallet creation that shows Bitcoin Cash as "just a parameter change, no biggie" rather than "Choose your Currency" (clearly stating Bitcoin Cash is a separate currency) can land Bitpay in hot water legally in many countries, as it could be interpreted as trying to pass off Bitcoin Cash as Bitcoin.

I'm not trying to say "Bitcoin is the real Bitcoin and Bitcoin Cash is not" but rather "These are separate currencies, and should be portrayed as such. Otherwise Bitpay is liable for criminal charges in many countries (if any lawyer cared to bring it up)"

For instance, the current UI would be very helpful to a scammer that was trying to sell someone Bitcoin Cash at a higher price by showing them Bitcoin's price. When the same scam was reported by people being tricked into buying Testnet coins, your reaction was to change the UI enough to make it obvious that Testnet coins were different and removed the fiat price display to remove confusion...

Then the following days you add Bitcoin Cash as if it were Bitcoin, saying "oh, this is just the same thing, but you're just changing the "Coin" parameter, nothing to see here."

fortran77:
This type of proposal should be based on technical merit, not on made up nonsense like the "criminal charges" assertion.

dabura667:
> technical merit

UX decisions are almost never made on technical merit... in fact most UX decisions are made at the cost of technical merit. Thanks for your input though.

This is a UX problem. And in the worst case scenario, a legal problem. Bitpay is a company that is officially sponsoring this software, so they are an easy target for litigation. The legal comment was just a kind reminder, as it seems their legal team is asleep at the wheel, or fired due to massive layoffs before Bitmain threw them a lifevest in exchange for their souls.

Ok, that last part was a jab.

but still, this UX is confusing, and I have actually had a user who bought 1 BCH from someone for $5000. The wallet they used was Bitpay wallet.

Bitpay is losing people money and helping scammers with their new UI.

Sorry if your political affiliations are blinding you from seeing the horrible UI, @fortran77

matiu:
[...]

PR #6826, opened 2017-09-30
Clear up confusion for new users.

dabura667:
Also see #6825


matiu:
We are not changing the name. Copay supports Bitcoin, Bitcoin Cash, and Bitcoin Testnet. "Bitcoin Wallet" is an appropriate description.

dabura667:
Bitcoin Cash is not Bitcoin. It is an altcoin.

If your measurement of what constitutes Bitcoin is having the word bitcoin in the name... then you will be adding any and all altcoins that use bitcoin in their name... correct?

PR #6874, opened 2017-10-09
Receive: wallets separated by coins

cmgustavo:


JDonadio:
"Select a wallet" header should be removed?

copay_-_copay_bitcoin_wallet

Note: changing tabs do not hide the wallet selector.

cmgustavo:
Title deleted in some cases. Thanks.

PR #6920, opened 2017-10-17
fix item arrow

JDonadio:
home

### DIFF
--- a/src/theme/variables.scss
+++ b/src/theme/variables.scss
@@ -67,15 +67,16 @@ $v-visible-radius:                                   6px !default;
 
 
 
-
 // App Material Design Variables
 // --------------------------------------------------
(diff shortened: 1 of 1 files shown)


JDonadio:
By default, <button> and <a> elements with the ion-item attribute will display a right arrow icon on ios mode. 
This feature is not enabled by default for md and wp modes, but it can be enabled by setting the Sass variables $item-md-detail-push-show and $item-wp-detail-push-show, respectively, to true.

JDonadio:
Source: https://ionicframework.com/docs/api/components/item/Item/

JDonadio:
Default Ionic values:

home
home
home

By default, all iOS devices use the `push-arrow` icon.

PR #7098, opened 2017-11-09
Copy to clipboard directive

cmgustavo:


src/app/app.module.ts
```diff
@@ -94,10 +94,18 @@ import { TxConfirmNotificationProvider } from '../providers/tx-confirm-notificat
...
 
+/* Directives */
+
+import { CopyToClipboard } from '../directives/copy-to-clipboard/copy-to-clipboard';
+
 export function createTranslateLoader(http: HttpClient) {
   return new TranslatePoHttpLoader(http, 'assets/i18n', '.po');
 }
 
+let declarations: any = [
```
gabrielbazan7: Baybe this array should be name as "directives" ?

cmgustavo: Ok, I agree.

cmgustavo: Updated. Thanks!

gabrielbazan7:
ACK

PR #7104, opened 2017-11-10
Enable BCH wallet by default

cmgustavo:
This PR has the option enabled by default in order to create a Bitcoin Cash wallet without need to enable the option on "Settings" -> "Bitcoin Cash Support" -> "Enable Bitcoin Cash".


dabura667:
1. Removing the scan feature and encouraging "duplication" by default is bad, it is reusing public keys. One of the major protections of Bitcoin (hiding public keys with two hashes) is nullified.
2. Bcash activation by default is confusing to the user with the current UI, icons, naming "Bitcoin Cash", etc.

What is the motivation for this, if I may ask?

Kirvx:
I think BitPay will allow invoice payments with Bitcoin Cash very soon... And they will have to recommend a wallet for BCH...

This is the best explanation IMO.

cmgustavo:
Updated. Do not create bch wallet on fresh installations

dabura667:
NACK

Rather than removing what little protection users have from scammers (selling BCH at BTC prices to newbies, we get tons of these support queries every day: "Why isn't my coin worth $6000???") you should probably work on making the UI less confusing. It is not obvious to a new user that these are two currencies with two prices, and your app's title says "A Bitcoin Wallet" and makes no mention of Bitcoin Cash.

I will ACK this PR once UI is fixed.

matiu:
Hi @dabura667 in the current version we had modified the UI to make it more clear there ar different currencies. Now wallets appear in different cards, and in the wallet selector, in other tabs, appear in different sections also.

Besides that, Copay is the only Bitcoin Cash wallet that enforces different address versions for Bitcoin Cash, to avoid confusion, and there is no way inside the app to bypass that format.

Kirvx:
Seems like my explanation is the good one... ;)

Kirvx:
@matiu You should view your private messages on reddit...

PR #7157, opened 2017-11-14
add infinite scroll

JDonadio:


gabrielbazan7:
ACK !

src/providers/wallet/wallet.ts
```diff
@@ -368,18 +368,22 @@ export class WalletProvider {
 
   public getTxHistory(wallet: any, opts: any) {
     return new Promise((resolve, reject) => {
-      opts = opts ? opts : {};
+      opts = opts || {};
```
Gamboster: Why is this change necessary?

JDonadio: Just habit 😂

PR #7161, opened 2017-11-14
Feat: Import wallet V4

Gamboster:


src/pages/add/import-wallet/import-wallet.ts
```diff
@@ -85,19 +120,259 @@ export class ImportWalletPage implements OnInit {
...
+          this.importErr = true;
+        } else {
+          this.popupProvider.ionicAlert('Error', err); // TODO: gettextcatalog
+        }
+        return;
+      });
+    }, 100);
+  }
+
+  /*
```
JDonadio: Should be removed

src/pages/add/import-wallet/import-wallet.ts
```diff
@@ -85,19 +120,259 @@ export class ImportWalletPage implements OnInit {
...
+      this.navCtrl.setRoot(HomePage);
+      this.navCtrl.popToRoot();
+    }).catch((err: any) => {
+      this.logger.warn(err);
+    });
+  }
+
+  private importExtendedPrivateKey(xPrivKey, opts) {
+    this.onGoingProcessProvider.set('importingWallet', true);
+    setTimeout(() => {
```
JDonadio: Is this timeout needed?

src/pages/add/import-wallet/import-wallet.ts
```diff
@@ -85,19 +120,259 @@ export class ImportWalletPage implements OnInit {
...
+
+          $state.go('tabs.home');
+        });
+      }, 100);
+    };
+    */
+
+  private importMnemonic(words: string, opts: any): void {
+    this.onGoingProcessProvider.set('importingWallet', true);
+    setTimeout(() => {
```
JDonadio: Is this timeout needed?

src/pages/add/import-wallet/import-wallet.ts
```diff
@@ -85,19 +120,259 @@ export class ImportWalletPage implements OnInit {
...
+        this.popupProvider.ionicAlert('Error', 'Wrong number of recovery words: ' + wordList.length);
+        return;
+      }
+    }
+
+    opts.passphrase = this.formData.passphrase || null;
+    this.importMnemonic(words, opts);
+  }
+
+  public toggleShowAdvOpts(): void {
```
JDonadio: If this function only switch to true/false should be here?

src/pages/add/import-wallet/import-wallet.ts
```diff
@@ -85,19 +120,259 @@ export class ImportWalletPage implements OnInit {
...
+
+    if (info.type == '1' && info.hasPassphrase)
+      this.popupProvider.ionicAlert('Error', 'Password required. Make sure to enter your password in advanced options'); //TODO gettextcatalog
+
+    this.formData.derivationPath = info.derivationPath;
+    this.formData.testnetEnabled = info.network == 'testnet' ? true : false;
+    this.formData.words = info.data;
+  }
+
+  public switchTestnetOff(): void {
```
JDonadio: Never used?

Gamboster: Not at the moment, but we will use it in a future feature, when we add BCH

src/pages/add/import-wallet/import-wallet.ts
Gamboster: Ok

src/pages/add/import-wallet/import-wallet.ts
Gamboster: I prefer to have this type of functions in the controller, to not dirty the html

src/pages/add/import-wallet/import-wallet.ts
JDonadio: Then should be added with BCH

Gamboster: I removed it, thanks

JDonadio:
CODE ACK

JDonadio:
Error at import:

* Testnet wallet
* From recovery phrase
* Do not set the testnet switch on

PR #7188, opened 2017-11-16
Change URL support (Copay)

cmgustavo:


matiu:
no yet, we need to enhance the forum yet.

PR #7230, opened 2017-11-21
[V4] Feat: Create shared wallets

Gamboster:


JDonadio:
If you create a shared wallet but just go back (from copayer page to home) then you cannot reach the copayer page anymore.

JDonadio:
Clicking in the invitation code display the toast message twice.

src/pages/add/copayers/copayers.html
```diff
@@ -0,0 +1,62 @@
...
+    <ion-title *ngIf="wallet">{{ wallet.name }}</ion-title>
+  </ion-navbar>
+
+</ion-header>
+
+
+<ion-content>
+  <div *ngIf="wallet" [hidden]="wallet.notAuthorized">
+    <ion-list class="copayers-secret">
+      <ion-item translate>
```
JDonadio: Translate directive should be used as pipe, not in ion labels.

src/pages/add/copayers/copayers.ts
```diff
@@ -0,0 +1,113 @@
...
+    private walletProvider: WalletProvider
+  ) {
+
+  }
+
+  ionViewDidLoad() {
+    console.log('ionViewDidLoad CopayersPage');
+    this.wallet = this.profileProvider.getWallet(this.navParams.data.walletId);
+    this.updateWallet();
+    this.shareIcon = this.platformProvider.isIOS ? 'iOS' : 'Android';
```
JDonadio: Is it used?
We should use one, not depending on the platform.

Gamboster: It isn't used at the moment. I will remove it

src/pages/add/copayers/copayers.html
Gamboster: Got it. Thanks

PR #7258, opened 2017-11-23
[V4] Feat: Join shared Wallet

Gamboster:


JDonadio:
Code ACK.

Missing: scan QR to join

Gamboster:
>>>Missing: scan QR to join

It will be implemented when we improve the scan component. @gabrielbazan7 is working on it

PR #7267, opened 2017-11-24
[V4] Feat/wallet info

JDonadio:
Display wallet info:

* M-N (label)
* Read only (icon and label)
* Testnet (icon and label) > hide fiat amount
* Auditable (label)


src/pages/includes/wallet-info.ts
```diff
@@ -0,0 +1,18 @@
+import { Component, Input } from "@angular/core";
+
+@Component({
+  selector: 'page-wallet-info',
+  templateUrl: 'wallet-info.html',
+})
+export class WalletInfoPage {
+  public _wallet: any;
```
Gamboster: If this variable is not used in the view, it should be private. I think the '_' is unnecessary due to typing

src/pages/includes/wallet-info.html
```diff
@@ -0,0 +1,22 @@
...
+  <div class="actions" *ngIf="(walletNotRegistered || updateStatusError)">
+    <a *ngIf="walletNotRegistered" (click)="recreate()" translate>Tap to recreate</a>
+    <a *ngIf="updateStatusError" (click)="updateAll(true)" translate>Tap to retry</a>
+  </div>
+
+  <!-- TODO -->
+  <ion-spinner class="spinner-dark recent right" icon="crescent" *ngIf="(updatingStatus || updatingTxHistory) && !walletNotRegistered && !updateStatusError && !updateTxHistoryError">
+  </ion-spinner>
+
+  <img style="height:3rem; width:3rem;" *ngIf="wallet.network == 'testnet'" src="assets/img/icon-wallet-testnet.svg">
```
Gamboster: The styles should be in the css file, to avoid dirtying the html. The same with the following lines

src/theme/global.scss
```diff
@@ -17,6 +17,12 @@ $global-colors: (
   }
 }
 
+.wallet-info {
```
Gamboster: I think wallet info should have its own CSS file

JDonadio: Done

src/pages/includes/wallet-info.ts
JDonadio: Changed to private.
The underscore is needed to avoid conflicts with the method names.

src/pages/includes/wallet-info.html
JDonadio: Done

Gamboster:
ACK

JDonadio:
Thanks for the review!

PR #7457, opened 2017-12-13
[V4] fix send searchbox and divide bch and btc wallets

gabrielbazan7:


src/pages/send/send.ts
```diff
@@ -22,10 +22,14 @@ import * as _ from 'lodash';
...
   public search: string = '';
-  public wallets: any;
-  public walletList: any;
-  public contactsList: any;
-  public hasWallets: boolean;
+  public walletsBtc: any;
+  public walletsBch: any;
+  public walletBchList: any;
+  public walletBtcList: any;
+  public contactsList: Array<object> = [];;
```
cmgustavo: double `; `

PR #8035, opened 2018-02-13
[v4] Fix: Camera permissions view and buttons behavior

Gamboster:


Gamboster:
screenshot_20180213-182430
screenshot_20180213-182437

PR #8075, opened 2018-02-20
REF: using clear to remove process

gabrielbazan7:


Gamboster:
ACK (Tested on Android)

PR #8089, opened 2018-02-21
[v3] Set transactionCurrency to create BitPay invoice

cmgustavo:
If the BitPay.com invoice has BCH turned on, this PR prevents to get an error while trying to top-up (with BTC).


Gamboster:
ACK

PR #8123, opened 2018-02-27
Fix formatting of table in README

Pastafarianist:
Please check out the diff.


Gamboster:
Good catch, but that will not fix the problem, since the readme will ignore the spaces.

Gamboster:
ACK
Good catch. Thank you!

Pastafarianist:
The problem was that the second line contained too many columns, it had nothing to do with whitespace.

PR #8134, opened 2018-02-28
Added card styling to home view

vanhue:
I realize it may look cleaner to not have cards. But we have quite a few upcoming changes on the design roadmap. Let's, for now, try to get the UI on par with whats currently in production and then discuss possible enhancements for a later release. Which I imagine will be more design/feature focused.

### DIFF
--- a/src/app/app.scss
+++ b/src/app/app.scss
@@ -16,7 +16,9 @@
 
 // Position
 // --------------------------------------------------
-
+ion-app.app-root{
+  background-color:#F5F5F5 !important;
+}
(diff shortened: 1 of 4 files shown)


src/app/app.scss
```diff
@@ -16,7 +16,9 @@
 
 // Position
 // --------------------------------------------------
-
+ion-app.app-root{
+  background-color:#F5F5F5 !important;
```
bitjson: should this use `$app-bg-color;`?

PR #8135, opened 2018-02-28
V4 issues

vanhue:
This change mostly has to do with the card styling on the home view. I realize it may look cleaner to not have cards. But we have quite a few upcoming changes on the design roadmap. Let's, for now, try to get the UI on par with whats currently in production and then discuss possible enhancements for a later release. Which I imagine will be more design/feature focused.


Gamboster:
ACK

PR #8150, opened 2018-03-02
REF: cards class added to home page and send page. 

rastajpa:
Also Text overflow added to item subtitle class and item title class

### DIFF
--- a/src/pages/home/home.scss
+++ b/src/pages/home/home.scss
@@ -37,9 +37,5 @@ page-home {
     justify-content: center;
     text-align: center;
   }
-  .wallet-warning {
-    font-size: 11px;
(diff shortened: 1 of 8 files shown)


src/pages/home/home.scss
```diff
@@ -37,19 +37,9 @@ page-home {
...
-    box-shadow: 0 1px 2px rgba(0, 0, 0, 0.25);
-    overflow: hidden;
-    margin: 20px auto;
-    display: block;
-    border-radius: 4px;
-    background-color: #ffffff;
-    .item-divider{
-      background-color:#ffffff;
-    }
+  .wallet-warning {
```
gabrielbazan7: dup

PR #8210, opened 2018-03-08
fix: border line added to all qr code

rastajpa:


Keensyst:
I need help with my wallet

PR #8279, opened 2018-03-22
Ref/logger for iOS and updates tx/txp on home view

cmgustavo:
* Also removes external logger plugin.
* Update home on resume event (mobile).


cmgustavo:
Updated.

PR #8320, opened 2018-03-27
Fix multi login and add wallet loading

ro-savage:
This PR removes an issue where the home page would load multiple time and multiple logins would occur.

This was occurring because the page would load once, before data was retrieve, and then it would reload the page to show the wallets.

To solve this, instead we show a 'loading...' wallet screen and then fire an event once a wallet is loaded.
We also timeout after 2.5s and assume there is no wallet to load. 

## Testing required
- [ ] Code review 
- [ ] Optional regression to ensure the wallets show up when loaded


ro-savage:
Sorry. We moved our repo, and our default branch got changed. This PR is not meant for this repo.

PR #8338, opened 2018-03-29
Add environments, e2e network request mocking, improved home view, Scan view e2e test

bitjson:
Closes #8337

PR #8352, opened 2018-04-02
Fix syntax for tslint

cmgustavo:


bitjson:
The tests are failing because CircleCI is caching the requirements from my PR:  https://github.com/bitpay/copay/pull/8338, which makes the visual diff testing more strict (now that we can mock API calls). Tests from other PRs will probably keep failing occasionally until we merge #8338.

PR #8402, opened 2018-04-10
Use Yarn

bitjson:


bitjson:
replaced by https://github.com/bitpay/copay/pull/8662

PR #8454, opened 2018-04-17
FEAT: NWJS

gabrielbazan7:


src/providers/release/release.ts
```diff
@@ -17,8 +17,15 @@ export class ReleaseProvider {
     return this.appVersion;
   }
 
-  public getLatestAppVersion() {
-    return this.http.get(this.LATEST_RELEASE_URL).map(x => x['tag_name']);
+
+  public getLatestAppVersion(): Promise<any> {
```
bitjson: Why wrap this in a new Promise? It's much simpler to just return and use the observable.

(Also, this breaks unit tests)

PR #8477, opened 2018-04-20
Fix: URL Handler on NW

Gamboster:


src/pages/home/home.ts
```diff
@@ -212,7 +221,10 @@ export class HomePage {
     // Used at the startup of Copay
     var argv = gui.App.argv;
     if (argv && argv[0]) {
-      this.handleOpenUrl(argv[0]);
+      // The timeout waits for the components to be initialized
+      setTimeout(() => {
```
bitjson: Isn't there a less brittle way to do this than setTimeout? At the very least it makes testing messy.

src/pages/home/home.ts
```diff
@@ -200,9 +206,12 @@ export class HomePage {
 
     // This event is sent to an existent instance of Copay (only for standalone apps)
     gui.App.on('open', (pathData) => {
-      if (pathData.indexOf(/^bitcoin(cash)?:/) != -1) {
+      if (pathData.indexOf('bitcoincash:/') != -1) {
```
bitjson: Can we add some unit tests here? This is really important logic, and we need to make sure it doesn't ever break

Gamboster: Hi @bitjson , Could you check this PR? We need to merge it to be able to send desktop. The home page tests are already being done in a separate PR.

src/pages/home/home.ts
Gamboster: In a future PR, which is almost ready, we will merge a fix that will allow us to eliminate the timeout, but it needs to be tested yet.

PR #8479, opened 2018-04-20
Ref config. Add error message in file-storage. Add tests

cmgustavo:


src/providers/config/config.spec.ts
```diff
@@ -46,28 +44,25 @@ describe('Provider: Config Provider', () => {
...
+      expect(defaults).not.toBeNull();
     });
     it('should set config from storage', () => {
-      persistenceProvider.getConfig().then((config) => {
+      persistenceProvider.getConfig().then(config => {
         expect(this.configCache).not.toBeNull();
       });
     });
-    it('should set default config if file is corrupted', async () => {
+    it('should set default config if file is corrupted', () => {
```
msalcala11: It seems like the description for this test is no longer accurate, since the code being tested no longer seems to reset to the default config on a load error.

Issue #8507, opened 2018-04-26
Add back E2E testing


bitjson:
Broken here:
https://github.com/bitpay/copay/commit/cade2bf8ef464baa049a33cfb6fd7d42e0d75824

PR #8510, opened 2018-04-26
Restore E2E Testing

bitjson:
This mostly reverts commit adaa126552089188a15f961280c9c7e23244a4c6, but avoids changing the "apply" process for now.

Fixes https://github.com/bitpay/copay/issues/8507


cmgustavo:
Code ACK! Just a comment there, I 'd revert this file `src/pages/send/confirm/confirm.ts` because it only has syntax changes and it 's not related with this PR (I think that it 's the reason why codecov/patch doesn't pass)

bitjson:
Awesome – @msalcala11 is actually working on a PR to get unit test coverage on `confirm.ts`, so we can wait for that 😄

bitjson:
Thanks for the unit tests @msalcala11!

@cmgustavo this is ready to merge 🚀

PR #8544, opened 2018-05-02
Improve Wallet Details Styling, and Implement Sticky Month Headers

msalcala11:


gabrielbazan7:
ACK ! Tested on IOS.

Gamboster:
ACK for Android!

But when I tested on desktop version, the Sticky Month Headers does not work like it does on mobile. Is it a feature only for mobile devices ?

msalcala11:
Interesting @Gamboster — sticky month headers will only work in browsers that support css `position: sticky`, and it looks like the NW.js version we're currently using (0.19.5) does not support it. 

Looks like support for `position: sticky` was added in NW.js version 0.20.0: 
https://nwjs.io/blog/v0.20.0/

PR #8641, opened 2018-05-14
FIX: bug fixed when the verification by fingerprint is canceled

rastajpa:


src/providers/touchid/touchid.ts
```diff
@@ -91,6 +91,8 @@ export class TouchIdProvider {
             );
             return resolve();
           } else this.logger.debug("Didn't authenticate!");
+        }, () => {
+          return reject();
```
msalcala11: It seems like this line is unnecessary if the `androidFingerprintAuth` Promise were directly returned instead of wrapped in a `new Promise`. Also this approach will not pass the rejection error to the `.catch` block.

src/providers/touchid/touchid.ts
```diff
@@ -143,6 +145,8 @@ export class TouchIdProvider {
           this.check()
             .then(() => {
               return resolve();
+            },() => {
+              return reject();
```
msalcala11: same here

src/providers/wallet/wallet.ts
```diff
@@ -1181,6 +1181,8 @@ export class WalletProvider {
         }).catch((err) => {
           return reject(err);
         });
+      },() => {
+        return reject();
```
msalcala11: same here

src/pages/send/confirm/confirm.ts
```diff
@@ -645,6 +645,8 @@ export class ConfirmPage {
 
   private setSendError(msg: string) {
     if (this.isCordova) this.slideButton.isConfirmed(false);
+    if (msg === "FINGERPRINT_CANCELLED") return;
```
msalcala11: This looks like it can work for android, but not for iOS because the error codes on iOS are different: https://ionicframework.com/docs/native/touch-id/

From the link above, here are the two possible "User Canceled" error codes`-2 or -128 - User tapped the 'Cancel' button`.

It might be worth consolidating the different iOS and Android messages in the `check` method of the touchid provider and then returning a consistent error message to all callers of the method when a users cancels touchid.

msalcala11:
ACK 🚀

PR #8662, opened 2018-05-17
Better builds

bitjson:


bitjson:
@cmgustavo @gabrielbazan7 this is ready for review 🚀

cmgustavo:
LGTM, but I 'd like to remove all the `^` for each package, because some developers do not respect the correct nomination and apply changes without backward compatibility as a minor version

PR #8667, opened 2018-05-17
Improve PIN Screen

msalcala11:
Improves the styling of the PIN screen, adds a shake animation, and provides vibration feedback on incorrect PIN entry.

Also refactors how the pin modal is displayed and hidden by migrating away from the ionic events approach and toward the simpler `modalCtrl.present()/navCtrl.pop()` method. This will improve overall app performance, since the pin modal page will only exist when it is needed.

# Before

# After


gabrielbazan7:
ACK

PR #8678, opened 2018-05-21
Fix & ref: warning modal added when user share session logs and refactoring of all modals

rastajpa:


src/pages/settings/about/session-log-warning-modal/session-log-warning-modal.scss
```diff
@@ -0,0 +1,48 @@
+page-session-log-warning-modal {
```
msalcala11: It looks like this css was copied and pasted from another file, which is bad practice and makes code very hard to maintain. Is there a way we can reuse the pre-existing css without having to copy and paste it into a new file?

src/components/custom-modal/custom-modal.html
```diff
@@ -0,0 +1,103 @@
...
+        </ion-col>
+      </ion-row>
+    </div>
+  </div>
+  <div class="height-content" *ngSwitchCase="'backup-needed'">
+    <div class="modal-img background-color-danger">
+      <img src="assets/img/warning.svg">
+    </div>
+    <div class="modal-content">
+      <div class="modal-heading" translate>Screenshots are not secure</div>
```
msalcala11: Text should be `Backup Needed`

src/components/custom-modal/custom-modal.html
```diff
@@ -0,0 +1,103 @@
...
+          <button ion-button clear color="danger" (click)="close(true)">
+            {{ 'Backup now' | translate }}
+          </button>
+        </ion-col>
+      </ion-row>
+    </div>
+  </div>
+  <div class="height-content" *ngSwitchCase="'backup-ready'">
+    <div class="modal-img background-color-success">
+      <img src="assets/img/warning.svg">
```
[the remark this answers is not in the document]
msalcala11: Still the wrong icon here.

msalcala11:
ACK! (Diff coverage not hit, but that seems to be the result of using prettier, so ack from me).

PR #8697, opened 2018-05-24
fixed newsletter email collection from onboarding checkbox.

vehcklox:
The email collection was never getting triggered when a user said they wanted to receive a newsletter during the onboarding process. I fixed the validation check to use the data from the checkbox, thus allowing a true/false to be set based on the input.


cmgustavo:
Thanks @vehcklox ! I close this PR because it has been fixed here ( #8666 ) and also deployed in v4.3.6.

PR #8705, opened 2018-05-28
Bug/email collector 03

cmgustavo:
* If `device.version` is not available, use `platform`
* Fixes column name


cmgustavo:
Column name in the spreadsheet was wrong. This PR is not needed any more.

PR #8717, opened 2018-05-30
REF: changes in the workflow of locking option

rastajpa:


src/pages/settings/settings.ts
```diff
@@ -182,12 +198,53 @@ export class SettingsPage {
...
+  }
+
+  private openLockModal(): void {
+    let config: any = this.configProvider.get();
+    let lockMethod =
+      config && config.lock && config.lock.method
+        ? config.lock.method.toLowerCase()
+        : null;
+    if (!lockMethod || lockMethod == 'disabled')
+      this.navCtrl.push(LockPage, {}, { animate: false });
```
msalcala11: The fact that there's no animation here feels a bit weird on iOS. Simplifying to `this.navCtrl.push(LockPage);` will make the page transition feel more native on iOS.

PR #8720, opened 2018-05-31
refactor(prettier): npm run fix:prettier

bitjson:


msalcala11:
ACK 🚀 🚀 🚀  (I ran prettier locally and got the same result)

PR #8730, opened 2018-06-01
Fixes for recent and txps transactions cards

gabrielbazan7:
Before:

copay_-_copay_bitcoin_wallet

Now: 
copay_-_copay_bitcoin_wallet

### DIFF
--- a/src/pages/includes/txp/txp.html
+++ b/src/pages/includes/txp/txp.html
@@ -16,8 +16,13 @@
     </span>
   </span>
   <div class="item-subtitle">
-    <span *ngIf="tx.creatorName">{{ tx.creatorName}}@</span>
-    <span>{{tx.wallet.name}}</span>
+    <div *ngIf="tx.creatorName">{{ tx.creatorName}}</div>
+    <div>{{tx.wallet.name}}</div>
+  </div>
+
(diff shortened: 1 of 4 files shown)


rastajpa:
ACK! (tested on desktop and android)

bitjson:
@gabrielbazan7 In the screenshot above, the `0.001 BTC` now wraps to show `0.001
BTC` . Can we fix that?

Also, it seems like the font size just needs to come down on the title. (Isn't that font size just much larger than in other places?) I think it's still cleaner to have the time string right-aligned.

gabrielbazan7:
What to do think about this @bitjson ?

copay_-_copay_bitcoin_wallet

bitjson:
Awesome! Much better. 

Are the headings all the same size? (Might be an optical illusion, but they seem slightly larger.)

It might be a little cleaner if we make the Proposals and Recent Transactions icons smaller? (So they appear to be the same height as the wallet's name – probably the same size as the bitcoin icon.) 

And for the Recent Transactions card, should we try to render the wallet icon inside the box (instead of just the solid color)? As it is, it almost feels like the image didn't download properly and isn't being displayed. So the smaller icons would look like the wallet icons, just scaled down.

PR #8731, opened 2018-06-01
Feat: new version available message on android and ios

rastajpa:


cmgustavo:
I mark it as "requested changes" just to review it a little bit more.

src/pages/home/home.spec.ts
```diff
@@ -5,13 +5,29 @@ import { Subject } from 'rxjs';
...
+import { AddPage } from '../add/add';
+import { CopayersPage } from '../add/copayers/copayers';
+import { WalletDetailsPage } from '../wallet-details/wallet-details';
 import { ConfigProvider } from './../../providers/config/config';
+import { ActivityPage } from './activity/activity';
 import { HomePage } from './home';
+import { ProposalsPage } from './proposals/proposals';
 
-describe('HomePage', () => {
+fdescribe('HomePage', () => {
```
bitjson: Looks like an `fdescribe` was left in here

src/pages/home/home.ts
```diff
@@ -405,29 +408,66 @@ export class HomePage {
...
+          }
+          this.showAvailableVersionCard(version);
+        }
+      )
+      .catch(err => {
+        this.logger.error('Error getLatestAppVersion', err);
+      });
+  }
+
+  private showAvailableVersionCard(version) {
```
bitjson: Could you add the types of methods? `version` here is a `string` right? (Ideally we can get real types for everything, without using `any`)

bitjson:
(And agree with @cmgustavo 👍)

src/pages/home/home.spec.ts
rastajpa: fixed up

PR #8734, opened 2018-06-01
enable noUnusedParameters

bitjson:
Also remove logging during tests and DRY tests


bitjson:
@msalcala11 ready for review 🚀

PR #8743, opened 2018-06-05
Fix: items titles for Recent Txs and Txp on home view for small screens

Gamboster:


Gamboster:
iPhone SE:
iphone_se_-_11_3
-86c6-93b495deacdb.jpg)

PR #8752, opened 2018-06-07
Fix UI: BitPay Card Settings and Home (TX and TXP)

cmgustavo:
img_0782


bitjson:
Looks like this needs a `npm run fix:prettier`

Gamboster:
LGFM

PR #8753, opened 2018-06-07
REF: refactor receive page style

rastajpa:


src/pages/receive/receive.scss
```diff
@@ -36,36 +32,27 @@ page-receive {
...
         margin-right: 5px;
         width: 16px;
         display: inline-block;
       }
     }
-    img {
-      width: auto;
-      display: inherit;
+    .aclass {
+      display: flex;
```
msalcala11: We now have a mixin called `center` for these 3 lines in `mixins.scss`. It can be used via `@include center;`

rastajpa: fixed up

PR #8823, opened 2018-06-27
Fix builder

cmgustavo:
* Removes `service-worker`
* Fix FCM plugin post install


rastajpa:
ACK. Tested on Android

PR #8872, opened 2018-07-13
FIX: The scanner button on import-wallet and join-wallet works well now

rastajpa:


rastajpa:
The version of the scanner desktop plugin that fix the `naturalWidth` error is still in the dev version. Therefore, it is not a good idea to use an unstable version.

PR #8874, opened 2018-07-13
Feat/custom user agent

cmgustavo:
`cordova-plugin-useragent` still has an issue that requires manual intervention: https://github.com/ionic-team/ionic-native/issues/2287


msalcala11:
ACK

PR #8881, opened 2018-07-16
Bug/shapeshift int 04

cmgustavo:
* Hide wallets that needs backup
* Show error if Shapeshift returns error
* Fixes icon for unconfirmed funds


msalcala11:
ACK

PR #8883, opened 2018-07-16
Bump v4.6 beta. Fix builder for mobile

cmgustavo:


cmgustavo:
If this PR (https://github.com/ionic-team/ionic-native/pull/2607) is merged, this fix won't be necessary.

PR #8887, opened 2018-07-17
Refactor of all the messages of the modals

rastajpa:


src/components/info-sheet/info-sheet.html
```diff
@@ -1,17 +1,68 @@
...
+    <info-sheet-template *ngSwitchCase="'receiving-bitcoin'" type="info" option="false">
       <span sheet-title-text translate>Receiving {{params.coinName}} address</span>
       <span sheet-text translate>
         Your wallet address is similar to a bank account number. You can add funds to your wallet by withdrawing funds you may have
         stored on an exchange to your wallet address. Or other {{params.coinName}} users can send you money by scanning your
         QR code.</span>
       <span sheet-button-text translate>GOT IT</span>
     </info-sheet-template>
+
+    <info-sheet-template *ngSwitchCase="'paper-key-unverified'" type="warning" option="true">
```
msalcala11: seems like the option input is not necessary since the primary button can always emit true and the secondary button can always emit false.

rastajpa: You're right. At first, I thought it could be useful but I was wrong. Thanks.

src/pages/send/amount/amount.ts
```diff
@@ -466,7 +466,35 @@ export class AmountPage extends WalletTabsChild {
...
+
+    if (amount > maxAmountAvailable) {
+      const insufficientFundsInfoSheet = this.actionSheetProvider.createInfoSheet(
+        'insufficient-funds',
+        { amount: _amount, coin: unit.shortName }
+      );
+      insufficientFundsInfoSheet.present();
+      insufficientFundsInfoSheet.onDidDismiss(option => {
+        if (!option) {
+          this.sendMax();
```
cmgustavo: This function shouldn't be called before `finish()`, because there is a `zone.run` inside. The variable `expression` could be unpredictable

rastajpa: You're right. This line was removed. Thanks!

gabrielbazan7:
ACK !

PR #8892, opened 2018-07-19
WIP FEAT: Handle clipboard data

gabrielbazan7:


cmgustavo:
IMO, this feature should be transparent. If users have a new bitcoin URI or paypro URL in their clipboard, ask for continue to process the payment (or join, or import) or ignore it. This prevents to have another FABs in main view and works for desktop as well.

PR #8897, opened 2018-07-20
Fix: Available units for a selected wallet in Amount Page

Gamboster:


cmgustavo:
Test error: ``` TypeError: Cannot read property 'isFiat' of undefined```

Gamboster:
@cmgustavo, unit tests fixed. Thanks

PR #8907, opened 2018-07-24
Redesign of the Confirm Page

rastajpa:


vanhue:
image

Really nice work on implementing the miner fee field! Thanks for that.

Just a couple of things to point out. Could you implement the new dialogue component to replace the high fee modal. See example mock: 
image

Also, the grey text on the red miner fee field seems a little difficult to read. Perhaps change it to the same color as fee amount and the fee priority color. 

The cancel button text seems a bit large. I would recommend keeping the text size consistent with the header label font size. 

And lastly, there seems to be a bug in the bitcoin address where it shows "..." before and after the address, when it should only be shown after.

Thanks again.

vanhue:
Oh one last thing. Could you increase the size and right padding of the send icon in the transaction header? See mock: 

image

rastajpa:
@vanhue thanks for reviewing the PR. I will work on the things you asked me.

About this: "Could you implement the new dialogue component to replace the high fee modal". It was implemented in another PR of mine, you could see it here: https://github.com/bitpay/copay/pull/8887

rastajpa:
_Could you implement the new dialogue component to replace the high fee modal._ 
**Implemented on https://github.com/bitpay/copay/pull/8887**


_Also, the grey text on the red miner fee field seems a little difficult to read. Perhaps change it to the same color as fee amount and the fee priority color._ 
**Done**


_The cancel button text seems a bit large. I would recommend keeping the text size consistent with the header label font size._ 
**Done**


_And lastly, there seems to be a bug in the bitcoin address where it shows "..." before and after the address, when it should only be shown after._ 
**I couldn't reproduce it on my Android device or on desktop version**


_Oh one last thing. Could you increase the size and right padding of the send icon in the transaction header?_ 
**The size was increased but the right padding it's not good for Android devices because this kind of devices not showing the default right arrow on ion-buttons. I think we have to think a little bit more about this.**

Thank you very much for reviewing the PR :)

PR #8911, opened 2018-07-25
Ref/events wallet tabs 01

cmgustavo:
* Fix: refresh home view after send transaction
* Ref: improves event handler between the new wallet tabs
* Restore events from home view if wallet-tabs has been opened


jamesscookk:
Gustavo is on my jcstewzy account James m cook original cell 7632026237 but
he removed it and is taking all bits someone please report he's also fhcki
by others over I got so much proof

PR #8915, opened 2018-07-27
REF: txp and tx details style of views refactored

rastajpa:


src/pages/txp-details/txp-details.scss
```diff
@@ -60,6 +71,33 @@ page-txp-details {
   .container-from {
     .wallet,
     .payment-proposal-to {
+      justify-content: space-between;
+      .address-divider {
```
Gamboster: It is used in receive, confirm, tx-details and txp.. I think we can move it to app.scss, to prevent duplicate code

src/pages/txp-details/txp-details.ts
```diff
@@ -46,6 +47,8 @@ export class TxpDetailsPage {
   public loading: boolean;
   public contactName: string;
   public showMultiplesOutputs: boolean;
+  public amount;
```
Gamboster: string ?

rastajpa:
@Gamboster Thanks for reviewing the PR. All your comments were resolved in my last commit.

PR #8934, opened 2018-07-31
REF: Changed all backup mini modals for action sheets

rastajpa:
In addition:
- Unused code removed.
- Changed all fee warning modals from integrations
- Fixes for 'paper-key-unverified' on Receive Page


src/pages/settings/about/session-log/session-log.ts
```diff
@@ -146,12 +149,11 @@ export class SessionLogPage {
...
 
   private showWarningModal() {
-    const sessionLogWarningModal = this.popupProvider.createMiniModal(
-      'sensitive-info'
-    );
-    sessionLogWarningModal.present();
-    sessionLogWarningModal.onDidDismiss(response => {
-      if (response) this.isCordova ? this.sendLogs() : this.copyToClipboard();
+    let infoSheetType: InfoSheetType = 'sensitive-info';
+    const infoSheet = this.actionSheetProvider.createInfoSheet(infoSheetType);
```
msalcala11: can be simplified to `this.actionSheetProvider.createInfoSheet('sensitive-info');`. Same for all other info sheets in this diff.

rastajpa: you're right! thanks

PR #8937, opened 2018-08-01
Use label-tip component for warning, danger and info messages in the app

gabrielbazan7:
copay_-_copay_bitcoin_wallet

copay_-_copay_bitcoin_wallet

copay_-_copay_bitcoin_wallet

copay_-_copay_bitcoin_wallet

copay_-_copay_bitcoin_wallet

copay_-_copay_bitcoin_wallet

copay_-_copay_bitcoin_wallet

copay_-_copay_bitcoin_wallet

copay_-_copay_bitcoin_wallet


vanhue:
Also if we could try and use 'Title Case' for all labels, that would be ideal.

gabrielbazan7:
Ok ! thanks for the comments. I was working in some changes, now the page looks like this: 
copay_-_copay_bitcoin_wallet

Do you think I should use a danger style instead of a warning style?

vanhue:
@gabrielbazan7 I'm a fan of what you have. Maybe danger/caution color could be for the low fee, and the warning color for the amount too low to spend.

PR #8960, opened 2018-08-07
fix info sheet title alignment

willhay:


denettehay:
So  I can have my money back now

PR #8973, opened 2018-08-09
FEAT: show message when an user use a legacy BCH address

rastajpa:
bitpay_-_secure_bitcoin_wallet

### DIFF
--- a/src/components/info-sheet/info-sheet.html
+++ b/src/components/info-sheet/info-sheet.html
@@ -112,5 +112,15 @@
       <span sheet-text translate>{{params?.msg}}</span>
       <span sheet-button-text translate>GOT IT</span>
     </info-sheet-template>
+
+    <info-sheet-template *ngSwitchCase="'legacy-address-info'" type="info">
+      <span sheet-title-text translate>{{params.appName}} BCH wallets use the CashAddr format by default</span>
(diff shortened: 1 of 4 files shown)


Gamboster:
LGFM. Great idea!

src/components/info-sheet/info-sheet.html
```diff
@@ -112,5 +112,15 @@
...
     </info-sheet-template>
+
+    <info-sheet-template *ngSwitchCase="'legacy-address-info'" type="info">
+      <span sheet-title-text translate>{{params.appName}} BCH wallets use the CashAddr format by default</span>
+      <span sheet-text translate>
+        If you need to send to "old" addresses, and you are SURE those are BCH addresses, you can "translate" them using the BCH
+        Translator tool.
+      </span>
+      <span sheet-button-text translate>Translate address</span>
+      <span sheet-second-button-text translate>GOT IT</span>
```
cmgustavo: This action should clear the input, isn't it?

rastajpa: If you think that it's better for user experience, I will do it now!

rastajpa: Done!

PR #9001, opened 2018-08-15
Fix: improves in Send Feedback Page

Gamboster:
This PR fixes:
- If you click on a link, closes the feedback page (maybe you want to do some other action after that)
- When we click on the super happy face and then click on "write a review", the modal appears. But it takes you to the store anyway if you didn't click on "Continue to App store" 
- Removed unnecessary code
- Feedback Complete Page, renamed to Share Page
- Use back arrow when you go to Feedback page from settings


src/pages/feedback/send-feedback/send-feedback.ts
```diff
@@ -121,14 +121,12 @@ export class SendFeedbackPage {
...
+    infoSheet.onDidDismiss(async option => {
+      if (!option) return;
       if (this.launchReview.isRatingSupported()) {
-        this.launchReview.rating().then(val => {
-          if (val == 'dismissed') this.navCtrl.popToRoot({ animate: false });
-        });
+        this.launchReview.rating();
       } else {
         await this.launchReview.launch();
-        this.navCtrl.popToRoot({ animate: false });
```
cmgustavo: Why we should keep in that page? If users back to the app, they 'll see feedback page again. I 'm not sure if it 's a good UX.

PR #9005, opened 2018-08-15
Fix: Send Page Zero State

msalcala11:
Now that the send tab no longer shows the amount page, we need to move the zero state on the amount page (shown when the user has no funds in their wallet) to the send page.


Gamboster:
ACK!

PR #9008, opened 2018-08-16
Avoid user to change or use send max amount if paypro

gabrielbazan7:
. Avoid user to change or use send max amount if the previous view is not Amount page
. Fix transition when the user wants to change the amount
. If user click outside infosheet go to change amount


Gamboster:
ACK

package.json
```diff
@@ -279,7 +280,8 @@
       "cordova-plugin-spinner-dialog": {},
       "cordova-plugin-vibration": {},
       "im.ltdev.cordova.UserAgent": {},
-      "cordova-launch-review": {}
+      "cordova-launch-review": {},
+      "cordova-plugin-privacyscreen": {}
```
Gamboster: Remove this

Gamboster:
ACK

PR #9016, opened 2018-08-17
REF: change popup msg for action sheet when the backup failed

rastajpa:


cmgustavo:
```Error: src/pages/backup/backup-game/backup-game.ts(61,13): error TS6138: Property 'popupProvider' is declared but its value is never read.```

PR #9020, opened 2018-08-17
Home zero state

gabrielbazan7:
copay_-_copay_bitcoin_wallet


vanhue:
Hey @gabrielbazan7 is this card for when a user deletes all their wallets?

gabrielbazan7:
@vanhue yes! we will need your help with this one.

PR #9059, opened 2018-08-24
FEAT: animation to clipboard card

rastajpa:
ezgif com-video-to-gif


Gamboster:
Great work! I tested it in my Samsung S7

PR #9067, opened 2018-08-28
FIX: Paypro error handling

rastajpa:


corn91791:
I need to fix my whole account is it you who is messing up my account
getting my payments redirected other places

PR #9114, opened 2018-09-12
Ref: persistent logs

Gamboster:
- Call getLogs function once
- setLogs condition: all logs during 3 days
- Send logs as a .txt file on mobile


cmgustavo:
Needs rebase

PR #9135, opened 2018-09-19
FIX: fixes for gift cards that failed for some reason

rastajpa:
Also, a correct handling of the different states that a gift card can have with their respective messages.

### DIFF
--- a/src/pages/integrations/gift-cards/purchased-cards/card-list-item/card-list-item.ts
+++ b/src/pages/integrations/gift-cards/purchased-cards/card-list-item/card-list-item.ts
@@ -15,11 +15,6 @@ import {
     <ion-label>
       <div class="card-list-item__label">{{card.amount | formatCurrency:card.currency}}</div>
       <ion-note class="card-list-item__note">{{card.date | amTimeAgo}}</ion-note>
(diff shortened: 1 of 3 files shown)


src/pages/integrations/gift-cards/purchased-cards/card-list-item/card-list-item.ts
```diff
@@ -15,6 +15,11 @@ import {
     <ion-label>
       <div class="card-list-item__label">{{card.amount | formatCurrency:card.currency}}</div>
       <ion-note class="card-list-item__note">{{card.date | amTimeAgo}}</ion-note>
+      <ion-note class="assertive" *ngIf="card.status == 'FAILURE' || card.status == 'RESEND'" translate>Error</ion-note>
+      <ion-note class="assertive" *ngIf="card.status == 'expired'" translate>Expired</ion-note>
+      <ion-note class="assertive" *ngIf="card.status == 'invalid'" translate>Still waiting confirmation (Use higher fees setting to faster delivery)</ion-note>
```
msalcala11: I don't think this is necessary `(Use higher fees setting to faster delivery)` We use payment protocol for gift card purchases and the fee is locked

cmgustavo: I agree

PR #9139, opened 2018-09-20
Fix/Update translations

cmgustavo:


src/assets/i18n/po/template.pot
```diff
@@ -785,7 +772,7 @@ msgstr ""
 msgid "Error adding new entry"
 msgstr ""
 
-msgid "Error creating the invoice"
+msgid "Error adding new log"
```
gabrielbazan7: This translated string does not exist anymore

PR #9150, opened 2018-09-24
FEAT: hide the sliding button when the keyboard is shown

rastajpa:


src/pages/integrations/gift-cards/confirm-card-purchase/confirm-card-purchase.html
```diff
@@ -88,7 +88,7 @@
 
 <ion-footer>
   <page-slide-to-accept #slideButton *ngIf="isCordova" [disabled]="!(wallet && totalAmountStr)" buttonText="{{'Slide to confirm' | translate}}"
-    (slideDone)="buyConfirm()"></page-slide-to-accept>
+    (slideDone)="buyConfirm()" [ngClass]="{'slide-confirm-fast': hideSlideButton, 'slide-confirm-slow': !hideSlideButton, 'slide-confirm-down': hideSlideButton}"></page-slide-to-accept>
```
Gamboster: I think you could use something like this:

```[ngClass]="{'class1 class2' : expression1}">```

rastajpa: you're right

PR #9305, opened 2018-11-15
Handle gift card purchase interruptions

msalcala11:
This PR gracefully handles the case when a user attempts to purchase a gift card and then closes the app before the gift card purchase completes (users might do this when experiencing slow api calls or shoddy network connections), and ensures that the user is able to fetch their claim code when they relaunch the app.

Here's the approach:

1. When the user confirms the purchase, before broadcasting the payment, store the gift card and its `accessKey` with status `UNREDEEMED`. If the user does not close the app during the broadcast step, and the purchase succeeds like usual, then update the `UNREDEEMED` status to `SUCCESS`.

3. If the user does close the app mid-purchase, then on app launch, check for any `UNREDEEMED` gift cards and attempt to redeem them. If the redemption attempt succeeds, save the claim code. If the redemption attempt indicates that the payment was never actually broadcasted before the user closed the app mid-purchase, delete the `UNREDEEMED` gift card.

### DIFF
--- a/src/pages/integrations/gift-cards/card-details/card-details.html
+++ b/src/pages/integrations/gift-cards/card-details/card-details.html
@@ -49,7 +49,7 @@
           </div>
         </div>
 
-        <div *ngIf="!card.claimCode">
+        <div *ngIf="!card.claimCode" class="card-info__status-message">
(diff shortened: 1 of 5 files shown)


gabrielbazan7:
ACK

PR #9353, opened 2018-11-27
Ref: change cached balance key name

Gamboster:


src/providers/persistence/persistence.ts
```diff
@@ -30,7 +30,7 @@ const Keys = {
   GIFT_CARD_USER_INFO: 'amazonUserInfo', // keeps legacy key for backwards compatibility
   APP_IDENTITY: network => 'appIdentity-' + network,
   BACKUP: walletId => 'backup-' + walletId,
-  BALANCE_CACHE: cardId => 'balanceCache-' + cardId,
+  BALANCE: id => 'balance-' + id, // Cached balance
```
matiu: there is no need to change `BALANCE_CACHE` to `BALANCE`.

add `BALANCE_CACHE_OLD`

PR #9376, opened 2018-11-30
Ref/cordova wkwebview 01

cmgustavo:


app-template/config-template.xml
```diff
@@ -51,14 +47,17 @@
...
-    <plugin name="cordova-plugin-qrscanner" spec="2.5.0" />
+    <plugin name="cordova-plugin-qrscanner" spec="2.6.0" />
+    <plugin name="cordova-plugin-add-swift-support" spec="1.7.2" />
     <plugin name="cordova-plugin-splashscreen" spec="5.0.2" />
     <plugin name="cordova-plugin-statusbar" spec="2.4.2" />
-    <plugin name="cordova-plugin-inappbrowser" spec="2.0.2" />
-    <plugin name="cordova-plugin-touch-id" spec="3.3.1" />
+    <plugin name="cordova-plugin-inappbrowser" spec="3.0.0" />
+    <plugin name="cordova-plugin-touch-id" spec="3.4.0">
+      <variable name="FACEID_USAGE_DESCRIPTION" value=" " />
```
matiu: should we add an id here?

app-template/config-template.xml
```diff
@@ -51,14 +47,17 @@
     <plugin name="cordova-plugin-android-fingerprint-auth" spec="1.4.3" />
     <plugin name="cordova-plugin-device" spec="2.0.1" />
     <plugin name="cordova-plugin-globalization" spec="1.0.9" />
-    <plugin name="cordova-plugin-qrscanner" spec="2.5.0" />
+    <plugin name="cordova-plugin-qrscanner" spec="2.6.0" />
+    <plugin name="cordova-plugin-add-swift-support" spec="1.7.2" />
```
matiu: why we need this?

app-template/config-template.xml
```diff
@@ -51,14 +47,17 @@
...
-    <plugin name="cordova-plugin-inappbrowser" spec="2.0.2" />
-    <plugin name="cordova-plugin-touch-id" spec="3.3.1" />
+    <plugin name="cordova-plugin-inappbrowser" spec="3.0.0" />
+    <plugin name="cordova-plugin-touch-id" spec="3.4.0">
+      <variable name="FACEID_USAGE_DESCRIPTION" value=" " />
+    </plugin>
     <plugin name="cordova-plugin-whitelist" spec="1.3.3" />
-    <plugin name="cordova-plugin-x-socialsharing" spec="5.4.1" />
-    <plugin name="cordova-plugin-x-toast" spec="2.6.0" />
+    <plugin name="cordova-plugin-x-socialsharing" spec="5.4.3" />
```
matiu: lets not upgrade plugin ins unless there is a security issue.

x-socialsharing // x-toast // qr-scanner

app-template/config-template.xml
```diff
@@ -119,6 +116,14 @@
     </platform>
 
     <platform name="android">
+      <access origin="https://bws.bitpay.com" />
```
matiu: do we need this even we use CSP?

app-template/config-template.xml
matiu: + inappbrowser

app-template/config-template.xml
cmgustavo: No, the value is " " (empty). Description for FaceId is optional for `NSFaceIDUsageDescription` in plist file.

app-template/config-template.xml
cmgustavo: `cordova-plugin-qrscanner` requires `cordova-plugin-add-swift-support`, but it 's not fixed version. Check this line:
```

cmgustavo: Install a fixed version of `cordova-plugin-add-swift-support` in our side, would avoid to install any update of this.

matiu: ack

app-template/config-template.xml
matiu: ack

app-template/config-template.xml
matiu: ack

app-template/config-template.xml
matiu: ok

cmgustavo: qr-scanner is ok. Version number on package.json was 2.6.0 for long time (maybe we forget to change it on config.xml). Previous version may break the building process with cordova 8: https://github.com/bitpay/cordova-plugin-qrscanner/issues/144#issuecomment-389976578

PR #9383, opened 2018-12-04
Fix: errors/warnings related 'use strict' mode

Gamboster:
Correction of some errors and warnings that I could observe, by running the code under "strict mode".

### DIFF
--- a/src/providers/config/config.ts
+++ b/src/providers/config/config.ts
@@ -106,117 +106,118 @@ export interface Config {
   };
 }
 
-const configDefault: Config = {
-  // wallet limits
-  limits: {
-    totalCopayers: 6,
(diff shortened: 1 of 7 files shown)


Gamboster:
@cmgustavo Fixed with configDefault as "readonly"

src/providers/config/config.ts
```diff
@@ -228,7 +229,7 @@ export class ConfigProvider {
             this.configCache = _.clone(config);
             this.backwardCompatibility();
```
matiu: function name:
setOldConfig?

Gamboster: We could change the name of that function in another pull request

PR #9387, opened 2018-12-06
 Fix: Derivation path when try to join using specific phrase

Gamboster:


src/pages/add/join-wallet/join-wallet.ts
```diff
@@ -126,7 +133,31 @@ export class JoinWalletPage {
...
+  public processInvitation(invitation: string): void {
+    if (this.regex.test(invitation)) {
+      this.logger.info('Processing invitation code...');
+      let walletData;
+      try {
+        walletData = this.bwcProvider.parseSecret(invitation);
+        this.setDerivationPath(walletData.network);
+        this.logger.info('Correct invitation code for ' + walletData.network);
+      } catch (ex) {
+        this.logger.warn(ex);
```
cmgustavo: This output of "logger" is hard to identify in the "Log" section if has not a prefix.

Gamboster: Fixed

PR #9394, opened 2018-12-10
Chore/packages 01

cmgustavo:


src/pages/send/send.spec.ts
```diff
@@ -605,10 +610,15 @@ describe('SendPage', () => {
...
+      instance.navParams = {
+        data: {
+          amount: '1.00000',
+          coin: Coin.BCH
+        }
+      };
       const amount = '1.00000';
       const coin = Coin.BCH;
-      spyOn(navParams, 'get').and.returnValues(amount, coin);
+      spyOn(instance.navParams, 'data').and.returnValues(amount, coin);
```
Gamboster: I think this line is not necessary. You could remove it

cmgustavo: +1 Thanks!

Issue #9396, opened 2018-12-10
refactor address.ts validateAddress


matiu:
validateAddress should be named `isValid`
and should return a boolean

PR #9399, opened 2018-12-11
Ref: addressProvider to remove unnecessary validateAddress function

Gamboster:
This PR Fixes: https://github.com/bitpay/copay/issues/9396

### DIFF
--- a/src/pages/send/multi-send/multi-send.ts
+++ b/src/pages/send/multi-send/multi-send.ts
@@ -240,9 +240,7 @@ export class MultiSendPage extends WalletTabsChild {
       return true;
     } else {
       this.invalidAddress = true;
-      const network = this.addressProvider.getNetwork(
-        data
-      );
+      const network = this.addressProvider.getNetwork(data);
 
       if (this.wallet.coin === 'bch' && this.wallet.network === network) {
         const isLegacy = this.checkIfLegacy();
--- a/src/pages/send/multi-send/multi-send.ts
+++ b/src/pages/send/multi-send/multi-send.ts
@@ -240,13 +240,9 @@ export class MultiSendPage extends WalletTabsChild {
       return true;
     } else {
       this.invalidAddress = true;
-      let network;
-
-      const extractedAddress = this.addressProvider.extractAddress(data);
-      const addressData = this.addressProvider.validateAddress(
-        extractedAddress
+      const network = this.addressProvider.getNetwork(
+        data
       );
-      network = addressData.network;
 
       if (this.wallet.coin === 'bch' && this.wallet.network === network) {
         const isLegacy = this.checkIfLegacy();
--- a/src/providers/address/address.spec.ts
+++ b/src/providers/address/address.spec.ts
@@ -9,128 +9,73 @@ describe('AddressProvider', () => {
     addressProvider = testBed.get(AddressProvider);
   });
 
-  describe('getCoin', () => {
-    const BTCLivenetAddresses = [
-      '15qT4RJTjs7GSTReEmgXr4LbMjTVQ51LZA',
-      'bitcoin:1CVuVALD6Zo7ms24n3iUXv162kvUzsHr69',
-      'bitcoin:1CVuVALD6Zo7ms24n3iUXv162kvUzsHr69?amount=0.00090000'
+  describe('getCoin and getNetwork', () => {
+    const testVectors: any[] = [
+      // BTCLivenetAddresses
+      ['15qT4RJTjs7GSTReEmgXr4LbMjTVQ51LZA', 'btc', 'livenet'],
+      ['bitcoin:1CVuVALD6Zo7ms24n3iUXv162kvUzsHr69', 'btc', 'livenet'],
+      [
+        'bitcoin:1CVuVALD6Zo7ms24n3iUXv162kvUzsHr69?amount=0.00090000',
+        'btc',
+        'livenet'
+      ],
+      // BTCTestnetAddresses
+      ['mscoRRUxbicZdUms3EqDr9jwtCmbbPgfcY', 'btc', 'testnet'],
+      ['bitcoin:n3LHh1WTFSpSVKXNFQo4U5eLAqowCadFHY', 'btc', 'testnet'],
+      [
+        'bitcoin:n3LHh1WTFSpSVKXNFQo4U5eLAqowCadFHY?amount=0.00090000',
+        'btc',
+        'testnet'
+      ],
+      // BCHLivenetAddresses
+      ['CHUwf57Maceqk5jhMX6NJkuLEbHEfqqgwx', 'bch', 'livenet'], // BCH Livenet Bitpay Style
+      ['qq9jk8jskjsmgqnzygwjsghp3knautm2dcnc5e4k7n', 'bch', 'livenet'],
+      ['bitcoincash:CHUwf57Maceqk5jhMX6NJkuLEbHEfqqgwx', 'bch', 'livenet'],
+      [
+        'bitcoincash:qz8ds306px5n65gffn8u69vvnksfw6huwyjczrvkh3',
+        'bch',
+        'livenet'
+      ],
+      [
+        'BITCOINCASH:qz8ds306px5n65gffn8u69vvnksfw6huwyjczrvkh3',
+        'bch',
+        'livenet'
+      ],
+      [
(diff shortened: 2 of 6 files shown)


src/pages/send/multi-send/multi-send.ts
```diff
@@ -240,13 +240,7 @@ export class MultiSendPage extends WalletTabsChild {
...
     } else {
       this.invalidAddress = true;
-      let network;
-
-      const extractedAddress = this.addressProvider.extractAddress(data);
-      const addressData = this.addressProvider.validateAddress(
-        extractedAddress
-      );
-      network = addressData.network;
+      const network = this.addressProvider.getNetwork(data);
```
matiu: beautiful!

src/providers/address/address.spec.ts
```diff
@@ -57,135 +61,78 @@ describe('AddressProvider', () => {
...
+      });
     });
 
-    it('should return undefined if we send an invalid address', () => {
-      let address = 'invalidAddress';
-      let result = addressProvider.getNetwork(address);
+    it('should return null if we send an invalid address', () => {
+      const address = 'invalidAddress';
+      const result = addressProvider.getNetwork(address);
       expect(result).toBeUndefined();
```
matiu: should return null or undefined?

src/providers/address/address.spec.ts
```diff
@@ -57,135 +61,78 @@ describe('AddressProvider', () => {
...
+      'BITCOINCASH:qz8ds306px5n65gffn8u69vvnksfw6huwyjczrvkh3',
+      'bitcoincash:qz8ds306px5n65gffn8u69vvnksfw6huwyjczrvkh3?amount=0.00090000'
+    ];
 
-      BTCTestnetAddress = 'bitcoin:n3LHh1WTFSpSVKXNFQo4U5eLAqowCadFHY';
-      result = addressProvider.getNetwork(BTCTestnetAddress);
-      expect(result).toEqual('testnet');
-    });
+    const BCHTestnetAddresses = [
+      // 'mg6PLV5yyUS6Gy55fJ7f994dQ7RpfJNYC9', // TODO: BCH Testnet Bitpay Style
```
matiu: add: 
`BCHTEST: mg6PLV5yyUS6Gy55fJ7f994dQ7RpfJNYC9`

src/providers/address/address.spec.ts
```diff
@@ -57,135 +61,78 @@ describe('AddressProvider', () => {
...
+    ];
 
-    it('should return correct network for BTC testnet address', () => {
-      let BTCTestnetAddress = 'mscoRRUxbicZdUms3EqDr9jwtCmbbPgfcY';
-      let result = addressProvider.getNetwork(BTCTestnetAddress);
-      expect(result).toEqual('testnet');
+    const BCHLivenetAddresses = [
+      'CHUwf57Maceqk5jhMX6NJkuLEbHEfqqgwx', // BCH Livenet Bitpay Style
+      'qq9jk8jskjsmgqnzygwjsghp3knautm2dcnc5e4k7n',
+      'bitcoincash:qz8ds306px5n65gffn8u69vvnksfw6huwyjczrvkh3',
```
matiu: add
`bitcoincash: CHUwf57Maceqk5jhMX6NJkuLEbHEfqqgwx`

src/providers/address/address.spec.ts
```diff
@@ -57,135 +61,78 @@ describe('AddressProvider', () => {
...
-        network: 'livenet',
-        coin: null,
-        translation: undefined
-      });
+    it('should return null if we send an invalid address', () => {
+      const address = 'invalidAddress';
+      const result = addressProvider.getCoin(address);
+      expect(result).toBeNull();
     });
   });
```
matiu: maybe it would be easy to defined the test vectors:

testVectors:
[
[<string>,<expectedCoin>,<expectedNetwork>],
[<string>,<expectedCoin>,<expectedNetwork>],
...
];

and then iterate

testVectors.forEach( (v)=> {
expect(addressProvider.getNetwork(v[0])).toEqual(v[2]);
expect(addressProvider.getCoin(v[0])).toEqual(v[1]);
});

src/providers/address/address.ts
```diff
@@ -24,80 +24,45 @@ export class AddressProvider {
     };
   }
 
-  public getCoin(address: string) {
-    address = address.replace(/^(bitcoincash:|bchtest:|bitcoin:)/i, '');
+  public getCoin(address: string): string {
+    let extractedAddress = this.extractAddress(address);
```
matiu: it is difficult to understand the function like it is written:


in `extractAddress(address)` is not obvious what is happending. Why extracting an address from something that is already `address`?

matiu: maybe `removePrefix`?

src/providers/address/address.ts
```diff
@@ -116,22 +81,22 @@ export class AddressProvider {
   }
 
   public extractAddress(address: string): string {
```
matiu: see above regarding this fn name

src/providers/address/address.ts
```diff
@@ -116,22 +81,22 @@ export class AddressProvider {
...
-    let extractedAddress = address
+    const extractedAddress = address
       .replace(/^(bitcoincash:|bchtest:|bitcoin:)/i, '')
       .replace(/\?.*/, '');
     return extractedAddress || address;
   }
 
   public isValid(address: string): boolean {
```
matiu: isValid's argument's name should be 'str' or similar. not address, to avoid confusion.

src/providers/address/address.ts
```diff
@@ -165,10 +130,10 @@ export class AddressProvider {
     }
 
     // Regular Address: try Bitcoin and Bitcoin Cash
-    let regularAddressLivenet = Address.isValid(address, 'livenet');
-    let regularAddressTestnet = Address.isValid(address, 'testnet');
-    let regularAddressCashLivenet = AddressCash.isValid(address, 'livenet');
-    let regularAddressCashTestnet = AddressCash.isValid(address, 'testnet');
+    const regularAddressLivenet = Address.isValid(address, 'livenet');
```
matiu: We should refactor this to:
// first part, that check BIP21:
 if (URI.isValid(str)) return true;
...
/// then... 
 if (Address.isValid(str, 'livenet')) return true;
 if (Address.isValid(str, 'testnet')) return true;
 if ... return true;
 if ... return true;

 return false;

src/providers/address/address.spec.ts
Gamboster: The getNetwork function currently returns undefined. That test works correctly

src/providers/address/address.spec.ts
Gamboster: we have to update BWS with bitcore-lib-cash version 0.18.1 to get correct values for BCHTEST prefix

src/providers/address/address.spec.ts
Gamboster: we have to update BWS with bitcore-lib-cash version 0.18.1 to get correct values for BCHTEST prefix

src/providers/address/address.ts
Gamboster: That function also removes suffixes from uris, for example: 
`'bitcoin:1CVuVALD6Zo7ms24n3iUXv162kvUzsHr69?amount=0.00090000'`

The function returns: ```1CVuVALD6Zo7ms24n3iUXv162kvUzsHr69```
Maybe I can use ```cleanAddress``` ?

src/providers/address/address.ts
Gamboster: Done! Thanks

src/providers/address/address.ts
Gamboster: Done

src/providers/address/address.ts
Gamboster: Done

src/providers/address/address.spec.ts
Gamboster: Great idea! Done

src/providers/address/address.spec.ts
Gamboster: Done

src/pages/send/multi-send/multi-send.ts
Gamboster: ^_^

PR #9422, opened 2018-12-22
show available balance while refreshing

matiu:
as per title.

also remove twoStep, since it is not longer used on BWC.

### DIFF
--- a/src/pages/wallet-details/wallet-details.html
+++ b/src/pages/wallet-details/wallet-details.html
@@ -44,7 +44,7 @@
           </div>
         </div>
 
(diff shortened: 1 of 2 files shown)


src/pages/wallet-details/wallet-details.html
```diff
@@ -44,7 +44,7 @@
           </div>
         </div>
 
-        <div class="balance-spendable" (click)="openBalanceDetails()" *ngIf="!wallet.balanceHidden && !wallet.scanning && showBalanceButton && wallet.status">
```
Gamboster: If you delete showBalanceButton, probably you have to remove that variable from the controller file also

PR #9393, opened 2018-12-09
#9392 Added possibility to print qr code

MrTschi:

Issue #9392, opened 2018-12-09
Possibility to print QR-Code


MrTschi:
**Fetaure request**
As an app-user I'd like to have the possibility to print the QR code directly within the app so that I can export/print it without taking a screenshot.

Gamboster:
Great idea! I saw your PR, so I will be reviewing it. Thank you for your collaboration :)

MrTschi:
Feature merged. Thank you!

chchzsyzchgg3393:
Al code Expired

PR #9425, opened 2018-12-24
Revert "#9392 Added possibility to print qr code"

cmgustavo:
Reverts bitpay/copay#9393

NOTE: Add new modules is not allowed until version 6 of Copay


matiu:
@MrTschi  Hey!

I am sorry about the confusion.  Your PR will be included on the upcoming Copay v6 version.

We have decided to look all dependencies within minor version releases, and only make changes on major version. Since the feature includes a new pluggin, it needs a new major version to be released.

Do you agree on releasing just the desktop printing support (`window.print`) for now, and the mobile support later?

MrTschi:
Hi @matiu,

ah okay. You should add this to Contributing.md file ^^

I think this feature is mainly useful for mobile devices because you can‘t really create a ‚bullet-proof‘ backup. If you made a mistake in manually writing down the seed it would end in hell. 
Printing the wallet recovery QR Code is much better in my opinion. 

Where can I find the release roadmap for 2019?

jesskie:
Thank you

jess

PR #9427, opened 2018-12-26
[FEAT] Single Passphrase

gabrielbazan7:


src/pages/onboarding/tour/tour.scss
```diff
@@ -14,6 +14,13 @@ page-tour {
       background-color: transparent;
     }
   }
+  .checkbox-list {
```
Gamboster: Is this class still necessary?

src/pages/home/home.html
```diff
@@ -127,8 +125,60 @@
           </div>
         </ion-item-divider>
         <ion-list>
-          <page-wallet-activity *ngFor="let notification of notifications" (click)="openNotificationModal(notification)"
-            [notification]="notification"></page-wallet-activity>
+          <page-wallet-activity *ngFor="let notification of notifications" (click)="openNotificationModal(notification)" [notification]="notification"></page-wallet-activity>
+        </ion-list>
+      </div>
+
+      <div class="card" *ngIf="vaultWallets && vaultWallets[0]">
```
Gamboster: We have repeated 3 times the code to list wallets in the same view:
We should optimize that in some way. For the moment, adding a 'TODO' would be enough, and we leave it for another PR

src/pages/onboarding/tour/tour.scss
gabrielbazan7: no longer

src/pages/home/home.html
gabrielbazan7: done

src/pages/onboarding/tour/tour.scss
gabrielbazan7: removed

PR #9436, opened 2019-01-03
Delete package-lock.json

serderovsh:


cmgustavo:
We are not going to remove package-lock. This is recommended by NPM Docs:

```
package-lock.json is automatically generated for any operations where npm modifies either the node_modules tree, or package.json . It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates
```

PR #9462, opened 2019-01-11
[FEAT] import vaults wallets

gabrielbazan7:


src/pages/add/import-wallet/import-wallet.html
```diff
@@ -57,14 +57,27 @@
...
+
+    <ion-item *ngIf="selectedTab != 'file' && !importForm.value.importVault">
       <ion-label stacked>{{'Coin' | translate}}</ion-label>
       <ion-select okText="{{okText}}" cancelText="{{cancelText}}" placeholder="{{'Select a coin' | translate}}" formControlName="coin">
         <ion-option value="btc">Bitcoin (BTC)</ion-option>
         <ion-option value="bch">Bitcoin Cash (BCH)</ion-option>
       </ion-select>
     </ion-item>
 
+    <ion-item [ngClass]="{'with-label': importForm.value.importVault}">
```
Gamboster: It is rare that the option to import a vault, removes one of the options above (Select coin). The user may get confused

gabrielbazan7: Done

PR #9473, opened 2019-01-16
[FIX] adding untilItChanges variable to Local/TxAction event 

gabrielbazan7:


jesskie:
Please let me used my key.  Vord

jess

jesskie:
My doughter is in hospital

jess

jesskie:
I need too. Col

jess

jesskie:
I will try sir

jess

PR #9491, opened 2019-01-25
[REF] Remove recent tx and txps card, show badge for pending txps

gabrielbazan7:
bitpay

bitpay


vanhue:
Nice work! 
Does the notification ticker disappear when there aren't any proposals? 
What action is required in order to clear the notifications?

gabrielbazan7:
> Does the notification ticker disappear when there aren't any proposals?
Yes
> What action is required in order to clear the notifications?
The transactions need to be broadcasted. (no pending transaction proposal -> no ticker)

PR #9514, opened 2019-02-09
Server-side gift card catalog management

msalcala11:
1. Moves offered gift cards and assets to the server
2. Caches gift card images locally on the device on 1st load for fast rendering and for the ability to view previously purchased gift cards properly in offline mode

This will allow adding new gift cards to the catalog without having to release a new version of the app.


src/utils.ts
```diff
@@ -0,0 +1,9 @@
+export function promiseSerial(tasks: Array<() => Promise<any>>) {
```
cmgustavo: What 's the purpose of this file?

msalcala11: This is where commonly-used, generic functions will go.

msalcala11: The first function I've added (`promiseSerial`) executes a lists of async tasks in series.

cmgustavo: ACK!

gabrielbazan7:
CODE ACK

PR #9522, opened 2019-02-14
Fix: Request image with null src for gift cards

Gamboster:
Image of the error:
bitpay


Gamboster:
I close this PR. The problem will be fixed in this other one: https://github.com/bitpay/copay/pull/9514

PR #9556, opened 2019-03-03
Add base64-js to the dependencies

jguthrie7:
On a fresh install, I had to also install `base64-js` to get copay to run


cmgustavo:
Closed in favor of #9583 . Thanks @jguthrie7 !

PR #9574, opened 2019-03-14
better handling of paypro err messages

matiu:
This is minor PR to handle better errors from Paypro. Tested with:

https://github.com/bitpay/bitcore/pull/2031

But it should not affect the current paypro error reporting.

### DIFF
--- a/src/providers/bwc-error/bwc-error.ts
+++ b/src/providers/bwc-error/bwc-error.ts
@@ -30,10 +30,6 @@ export class BwcErrorProvider {
         case 'INVALID_BACKUP':
           body = this.translate.instant('Wallet Recovery Phrase is invalid');
           break;
- 
-        case 'INVALID_BACKUP':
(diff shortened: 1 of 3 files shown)


src/providers/bwc-error/bwc-error.ts
```diff
@@ -19,6 +19,18 @@ export class BwcErrorProvider {
...
       switch (name) {
+        case 'UNCONFIRMED_INPUTS_NOT_ACCEPTED':
+          body = this.translate.instant('This invoice does not accept unconfirmed inputs.');
+          break;
+ 
+        case 'INVOICE_EXPIRED':
+          body = this.translate.instant('This invoice is not longer accepting payments');
+          break;
+ 
+        case 'INVALID_BACKUP':
```
Gamboster: This case already exists

matiu: thanks,. updated.

PR #9603, opened 2019-03-26
Use case-insensitive sort for gift cards

msalcala11:
Also ensures gift cards are properly displayed in offline mode.


orangecstsne04:
?

PR #9627, opened 2019-04-04
[FIX] receive view when getting an error

gabrielbazan7:


src/components/info-sheet/info-sheet.html
```diff
@@ -219,5 +219,12 @@
       <span sheet-button-text translate>Get help</span>
       <span sheet-second-button-text translate>Report issue</span>
     </info-sheet-template>
+
+    <info-sheet-template *ngSwitchCase="'address-unavailable'" type="danger">
+      <span sheet-title-text translate>Cannot create new address</span>
+      <span sheet-text translate>Check your internet connection and try again.</span>
```
cmgustavo: Another issue if you are trying to create a new address is when GAP reach to maximum value

PR #9648, opened 2019-04-10
Fix multisig gift card purchases

msalcala11:
Also:

1. Automatically deletes payment proposal if a gift card purchase returns a PayPro error (so that the user doesn't have to manually delete it to re-try the purchase).
2. Displays a `Print Required` alert for gift card brands that require a printed gift card for redemption.


gabrielbazan7:
ACK !

PR #9649, opened 2019-04-11
 Ref/last balance

matiu:
- clean lastKnown balance code for wallets and cards
- better offline balance and history


src/providers/persistence/persistence.ts
```diff
@@ -105,10 +105,6 @@ export class PersistenceProvider {
     });
   }
 
-  deleteProfile() {
```
gabrielbazan7: "should be able to delete a profile" spec from persistence.spec will fail

PR #9667, opened 2019-04-16
Ref/packages 01

cmgustavo:
* Remove unused packages. (avj, archive, connect, mkdirp, etc)
* Removes commitizen config.
* Removes junitReporter


matiu:
👏

PR #9738, opened 2019-05-21
[FEAT] PayPro Provider tests

gabrielbazan7:


src/providers/paypro/paypro.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    it('should return paypro details', () => {
+      walletFixture.id.fetchPayPro = (_payProUrl, cb) => {
+        const paypro = {
+          verified: true
+        };
+        return cb(null, paypro);
+      };
+      profileProvider.wallet = walletFixture;
+      payproProvider
+        .getPayProDetails('https://bitpay.com/i/BAikNA9QFMFiEs8HtNnQVt', 'btc')
```
cmgustavo: This should be local data. No external connection need to be realize with Unit Testing.

PR #9749, opened 2019-05-22
Pre-Single Seed 

msalcala11:
Updates the styling of the app to be simpler and more minimal in preparation for single-seed functionality.


gabrielbazan7:
Nice!!! is looking great. Some things that I note is that the "Balance Hidden" feature is not working in the home view and the headers are not white using Copay template. I will review it more in-depth today

src/pages/txp-details/txp-details.html
```diff
@@ -1,4 +1,4 @@
-<ion-header class="wallet-details-header">
+<ion-header class="wallet-details-heade">
```
gabrielbazan7: typo

gabrielbazan7:
There is any reason why you use `this.statusBar.styleDefault();` in home view ? The status bar looks like this: 

Screenshot_20190523-152052_Copay

PR #9962, opened 2019-07-17
Activation fees

msalcala11:


cmgustavo:
Already merged into v6.0

PR #9974, opened 2019-07-19
Ref: Separate read only wallets in home page

Gamboster:
BitPay


bknght0911:
I do not subscribe to you , thank you,

PR #9977, opened 2019-07-19
Mnemonic recover tool

bsnowden3:
PR allows users to recover old wallet mnemonic phrases after migration to single seed.
Jul-19-2019 14-09-42

### DIFF
--- a/src/pages/settings/advanced/advanced.html
+++ b/src/pages/settings/advanced/advanced.html
@@ -16,13 +16,11 @@
           If enabled, wallets will also try to spend unconfirmed funds.
         </div>
         <div *ngIf="!isCopay" translate>
-          However, unconfirmed funds are not allowed for spending with merchants, BitPay Card loads, or BitPay in-app gift card purchases.
+          However, unconfirmed funds are not allowed for spending with
+          merchants, BitPay Card loads, or BitPay in-app gift card purchases.
         </div>
       </div>
     </label-tip>
-    <button ion-item *ngIf="oldProfileAvailable" class="with-label no-border" (click)="openWalletRecoveryPage()">
-      Recover old wallets
-    </button>
   </ion-list>
   <ion-list class="settings-list">
     <ion-item class="with-label no-border">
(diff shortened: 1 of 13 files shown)


bsnowden3:
Title of warning has been fixed to fit on one line.

src/pages/settings/advanced/wallet-recover-page/wallet-recover-page.html
```diff
@@ -0,0 +1,18 @@
+<ion-header>
+  <ion-navbar>
+    <ion-title>{{'Old Wallet Recovery Tool'}}</ion-title>
```
gabrielbazan7: this should be: <ion-title>Old Wallet Recovery Tool</ion-title>

src/pages/settings/advanced/advanced.html
```diff
@@ -16,6 +16,8 @@
         <div *ngIf="!isCopay" translate>However, unconfirmed funds are not allowed for spending with merchants, BitPay Card loads, or BitPay in-app gift card purchases.</div>
       </div>
     </label-tip>
+    <button ion-item class="with-label no-border" (click)="openWalletRecoveryPage()">{{'Recover old wallets' | translate}}</button>
```
gabrielbazan7: This need a condition that checks if profile old exist since we have plans to delete the old profile in the future and we don't want to continue showing this option

src/pages/settings/advanced/wallet-recover-page/wallet-mnemonic-recover-page/wallet-mnemonic-recover-page.html
```diff
@@ -0,0 +1,31 @@
+<ion-header>
+  <ion-navbar>
+    <ion-title>{{ title }}</ion-title>
+  </ion-navbar>
+</ion-header>
+
+<ion-content>
+  <div class="container">
+    <label-tip class="no-arrowhead" type="warn" margin-top>
+      <span label-tip-title translate>Mnemonic Recovery Warning</span>
```
gabrielbazan7: remove translates completely from this feature

src/pages/settings/advanced/advanced.html
gabrielbazan7: Or maybe at the time, we remove it we can remove the code too. So maybe is ok this way. What do you think ?

bsnowden3:
Jul-22-2019 15-23-19

matiu:
there seems to be a problem when there is no profile old:
BitPay

matiu:
CodeACK.
great work!

On a app with profileOld worked as expected.

ToDo:
- [ ] please handle with there is no profileOld
- [ ] Could you please remove the "copy to clipboard" button. it could be insecure.
- [ ] Put the option "recover old wallets"  at the end of advanced option (last item). That item should not appear if there is no `profileOld`.
-

bsnowden3:
Thanks! I've now addressed those requested changes

PR #10005, opened 2019-07-23
Fix: Specify derivation path in create and import

Gamboster:


src/pages/home/home.scss
```diff
@@ -514,7 +514,7 @@ page-home {
 
     &.wallets {
       button.item.item-block {
-        border-left: 8px solid #EDEEF3;
+        border-left: 8px solid #edeef3;
```
cmgustavo: It's been fixed in another PR 👍

Gamboster: done

PR #10048, opened 2019-07-29
fix(Fingerprint): Check Face ID if it has been disabled by many retry…

cmgustavo:
…. Uses Biometric ID


src/providers/wallet/wallet.ts
```diff
@@ -1611,6 +1611,8 @@ export class WalletProvider {
     return promise.then(() => {
       this.configProvider.set(opts);
       return Promise.resolve();
+    }).catch(e => {
```
gabrielbazan7: this catch is not necessary

cmgustavo: +1

PR #10078, opened 2019-08-06
test accounts when joinning a new wallet

matiu:
same logic of "create wallet" but for "join wallet": Test until a free account number is found.


Gamboster:
The tests fail. I think it needs a `fix: prettier`

Gamboster:
ACK

PR #10107, opened 2019-08-15
Fix: scanner button overlapping

Gamboster:


src/pages/home/home.scss
```diff
@@ -5,6 +5,13 @@ page-home {
       background: $feather;
     }
 
+    &.content-md {
```
cmgustavo: `*-md` mean it only works for Android (or web for devs). Is that ok?

Gamboster: Yes, this fix is just for android devices

PR #10108, opened 2019-08-19
ref(fcm): keep fcm plugin. Fix builder. Minimum API level 27

cmgustavo:


cmgustavo:
Ready for review. Tested on iOS and Android as well.

PR #10112, opened 2019-08-21
Chore/ fix compile errors

justinkook:
- fix tsc compile errors

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "copay",
   "description": "A Secure Bitcoin Wallet",
   "author": "BitPay",
-  "version": "6.2.0",
+  "version": "6.2.1",
   "keywords": [
     "bitcoin",
     "wallet",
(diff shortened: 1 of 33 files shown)


Gamboster:
Great! I totally agree with increasing typescript version
Just as a reminder and in case you need to do it, we can use the own filter of js arrays, to avoid problems of types that may exist with _.filter

For example:
```this.service = this.homeIntegrationsProvider.get().filter(homeIntegration => {
      homeIntegration.name == this.serviceName;
    });```

instead of 

```this.service = _.filter(this.homeIntegrationsProvider.get(), {
      name: this.serviceName
    });```

File: package.json [hunk 1]
```diff
@@ -208,7 +208,7 @@
     "ts-node": "5.0.0",
     "tslint": "5.8.0",
     "tslint-config-prettier": "1.7.0",
-    "typescript": "2.6.2",
+    "typescript": "3.5.3",
```
cmgustavo: I think it's danger to upgrade a major version, because it's incompatible with current `webpack` 3.12.

PR #10116, opened 2019-08-22
feat(paypro): adds header, app name and version

cmgustavo:


matiu:
this is a temporary script until the change is merged in bwc

PR #9966, opened 2019-07-18
Feature(Eth) - Target ETH branch

justinkook:
# Requirements:
https://github.com/bitpay/bitcore/pull/2281

# Copay Set Up

**Bitcore**

1. `npm i` in `bitcore/` root directory.

2. `rm -rf node_modules package-lock.json` in `bitcore/packages/bitcore-wallet-client/` directory.

> This is to remove symlinks / lerna bootstrap npm packages inside bitcore-wallet-client.

3. `npm i` inside `bitcore/packages/bitcore-wallet-client/` to install the NPM Registry versions of packages.

**Copay**

1. `rm -rf node_modules package-lock.json` inside `copay/` to remove old versions of bitcore-wallet-client.

2. `bash pack-all.sh` inside `copay/` to install local bitcore-wallet-client.

## Status:
- FIN

## Steps to set up eth:

`pack-all.sh` assumes copay and bitcore folders to be a sibling and in the same folder.
ex:
```
dev/
   copay/
      src/
   bitcore/
      packages/
```

1. Run `pack-all.sh` bash script to npm install your local version of bitcore-wallet-client into Copay. This is done using `npm pack` command. Update package versions as needed.

Your `package.json` should include this:
```
    "bitcore-wallet-client": "file:../bitcore/packages/bitcore-wallet-client/bitcore-wallet-client-8.6.0.tgz",
...
```
2. Comment / Delete Content Security Policy in `index-template.html`:
```
 <!-- <meta http-equiv="Content-Security-Policy" content="default-src ... data:"> -->
```

2. Point BWS url to `http://localhost:3232/bws/api` in `bws.ts`.

3. When creating an ethereum wallet set BWS url to:
`http://localhost:3232/bws/api`

### Common Errors:
`Uncaught TypeError: exists is not a function`
`Version guard duplicate version of bitcore-lib` in `bws.logs`

Fix:
`rm -rf node_modules/ && npm i`

PR #10128, opened 2019-08-29
Erc20

justinkook:
Based on ETH branch
https://github.com/bitpay/copay/pull/9966

## Status

- Working Balance / Tx History / Send Tokens
- Balance sometimes doesnt work and gets overrided.


Derick7:
This copay support contact is really not helping their users at all, I wrote here long time ago, that I have some bitcoin inside my wallet I mistakenly save 11 Recovery phrase instead of 12, no response ever since, is this app not really the best app for saving bitcoin, you don’t really have customer in your heart, ever sis Ec I wrote my message they keep ignoring me here

PR #10197, opened 2019-10-03
ref(keys): rename group by key. A few UI improvements

cmgustavo:


src/pages/settings/settings.html
```diff
@@ -32,17 +32,14 @@
         </ion-icon>
         <ion-label>
           <div *ngIf="walletsGroup[0].canSign">{{profileProvider.getWalletGroup(walletsGroup[0].keyId).name}}</div>
-          <div *ngIf="!walletsGroup[0].canSign" translate>Read Only Wallets</div>
+          <div *ngIf="!walletsGroup[0].canSign" translate>Read Only Key</div>
```
Gamboster: Are we sure to rename Read Only wallets to Read only keys? Even if that kind of wallets doesn't have private key?

src/pages/settings/wallet-group-settings/wallet-group-settings.html
```diff
@@ -1,26 +1,45 @@
...
+
+<ion-content #scrollArea [fixed-scroll-bg-color]="headerColor" [ion-content-background-color]="headerColor">
+  <div class="wrapper">
+    <expandable-header [scrollArea]="scrollArea" [fadeFactor]="5" [disableFade]="true" #expandableHeader>
+      <ion-toolbar [navbar-bg]="headerColor" class="wide-header__title">
+        <expandable-header-primary>
+          <span *ngIf="canSign">{{'Key Settings' | translate}}</span>
+          <span *ngIf="!canSign">{{'Read Only Key' | translate}}</span>
+          <div class="subtitle" *ngIf="wallets">
+            This key contains {{wallets.length}} wallets.
```
Gamboster: We should not show this message to read only. Since they can come from multiple different keys

Gamboster:
There are still places with "Read only wallets" text:
home.html
add-wallet.html
profile.ts

src/pages/settings/wallet-group-settings/wallet-group-settings.html
cmgustavo: Done. Thanks!

PR #10213, opened 2019-10-09
Update BitPay Invoice URI flow

justinkook:
- Deprecated`/invoiceData/` api route, does not provide `CORS` headers anymore, so we cannot use this route in Copay. Uses alternative `/invoices/` route.

The difference is `{ org, buyer }` is not returned from so we do not know if an email is set and we cannot set the `buyerProvidedEmail`. 

Also the merchant name is not known since `org` is not returned.

`confirm-invoice.ts`

- Removed `invoiceName`, anything related to `email`, `merchantEmail`, `buyerProvidedEmail`.

- Removed email field (will be replaced with BitPay ID soon).

### DIFF
--- a/src/pages/integrations/invoice/confirm-invoice/confirm-invoice.ts
+++ b/src/pages/integrations/invoice/confirm-invoice/confirm-invoice.ts
@@ -50,17 +50,13 @@ import { ConfirmPage } from '../../../send/confirm/confirm';
 })
 export class ConfirmInvoicePage extends ConfirmPage {
   public invoiceData: any;
-  public invoiceName: string;
   public invoiceId: string;
   public invoiceUrl: string;
-  public email: string;
   public onlyIntegers: boolean;
   public currency: string;
   public invoiceFeeSat: number;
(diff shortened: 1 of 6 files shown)


src/pages/integrations/invoice/confirm-invoice/confirm-invoice.ts
```diff
@@ -217,14 +202,6 @@ export class ConfirmInvoicePage extends ConfirmPage {
...
   }
 
-  ionViewWillLeave() {
-    if (!this.invoicePaid) {
-      this.clipboardProvider.copy(this.invoiceUrl);
-    } else {
-      this.externalLinkProvider.open(this.browserUrl);
-    }
-  }
-
```
justinkook: This fixes the `https://test.bitpay.com/invoice/jasdfasdf9879asdf` copied in `home.ts` page instead of `https://test.bitpay.com/i/jasdfasdf9879asdf`

gabrielbazan7:
ACK!

PR #10264, opened 2019-10-24
Feat: QR code Key export

Gamboster:
1) The option to export the key as a QR code will appear in:
`Settings -> Key Settings -> Advanced -> Export Key with QR`
2) This option will only be seen if:
- The key backup is done
- The key has mnemonics
3) When scanning the QR, from the general scanner of the app -> It takes you to import and autocomplete the words
4) When scanning the QR from the import scanner -> that autocompletes the words

Slack___Mustafa_Inamullah___BitPay

### DIFF
--- a/src/pages/add/import-wallet/import-wallet.ts
+++ b/src/pages/add/import-wallet/import-wallet.ts
@@ -186,12 +186,11 @@ export class ImportWalletPage {
       return undefined;
     }
     if (info.type == '1' && info.hasPassphrase) {
-      const title = this.translate.instant('Error');
+      const title = this.translate.instant('Warning');
       const subtitle = this.translate.instant(
         'Password required. Make sure to enter your password in advanced options'
       );
       this.popupProvider.ionicAlert(title, subtitle);
-      return undefined;
     }
     return info;
   }
--- a/src/pages/settings/wallet-group-settings/wallet-group-qr-export/wallet-group-qr-export.html
+++ b/src/pages/settings/wallet-group-settings/wallet-group-qr-export/wallet-group-qr-export.html
@@ -1,6 +1,6 @@
 <ion-header>
   <ion-navbar>
-    <ion-title>{{'QR Key Export' | translate}}</ion-title>
+    <ion-title>{{'Export key' | translate}}</ion-title>
   </ion-navbar>
 </ion-header>
 <ion-content>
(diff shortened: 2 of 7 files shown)


src/pages/add/import-wallet/import-wallet.ts
```diff
@@ -186,12 +186,11 @@ export class ImportWalletPage {
       return undefined;
     }
     if (info.type == '1' && info.hasPassphrase) {
-      const title = this.translate.instant('Error');
+      const title = this.translate.instant('Warning');
       const subtitle = this.translate.instant(
         'Password required. Make sure to enter your password in advanced options'
       );
       this.popupProvider.ionicAlert(title, subtitle);
-      return undefined;
```
cmgustavo: What happen in this case if not return `undefined`? Will it return `info` anyway?

src/pages/settings/wallet-group-settings/wallet-group-settings.html
```diff
@@ -115,6 +115,10 @@
 
       <ion-item-divider *ngIf="canSign">{{'Advanced' | translate}}</ion-item-divider>
 
+      <button *ngIf="!needsBackup && canSign && !isDeletedSeed" ion-item (click)="openQrExport()">
+        <span translate>Export Key with QR</span>
```
cmgustavo: I prefer: `Export Key`.

src/pages/settings/wallet-group-settings/wallet-group-qr-export/wallet-group-qr-export.html
```diff
@@ -0,0 +1,22 @@
+<ion-header>
+  <ion-navbar>
+    <ion-title>{{'QR Key Export' | translate}}</ion-title>
```
cmgustavo: Title: `Export Key`

src/pages/add/import-wallet/import-wallet.ts
Gamboster: Yes, so the words will be completed in mnemonics field.
If you are importing a wallet with mnemonics with passphrase, the import page will show you a message saying that don't forget to put the password.

src/pages/settings/wallet-group-settings/wallet-group-settings.html
Gamboster: Done

src/pages/settings/wallet-group-settings/wallet-group-qr-export/wallet-group-qr-export.html
Gamboster: Done

PR #10271, opened 2019-10-28
Feat: unit tests for eth in incomingDataProvider

Gamboster:
Also:
- Fix `handleEthereumAddress` call with redirParams
- Missing tests added to incomingDataProvider


justinkook:
Thanks for this! 👍

PR #10290, opened 2019-11-04
Add bitpay id pairing and card dashboard in app browser/functionality.

JohnathanWhite:
This PR adds bitpay ID as well as the wallet card dashboard in app browser functionality.

- Added the base bitpayID template to settings
- Created utils/iab-patch.js that runs post plugin install and sets the status bar height on ios to 0. This is needed to achieve full screen on notched phones (iphoneX etc).
- Added the iab provider that allows for generating/accessing iab instances.
- Added method generatePairingToken in BitPayID provider that takes the pairing code passed from the iab and creates/persist a token and user info into persistenceStorage. 
- Added iab card provider to handle messages sent from card iab
- Added a generic graph query handler - this takes a postMessage payload from the iab with a query and query variables and signs the request. It then passes it back to the iab where the call is made. 
- Added the new card home page tab and gated it behind the hidden feature flag.
- Added load error logic to the iab - if a load failure occurs, the show method of the iab gets overwritten to execute a default error action sheet.


couby410:
https://codecov.io/gh/bitpay/copay/pull/10290?src=pr&el=continue

justinkook:
Whoa I'll check this out. Pretty exciting!

cmgustavo:
Try to avoid pushing the `package-lock`. It will save a lot of headaches on rebase `master` branch

package.json
```diff
@@ -140,7 +141,7 @@
     "cordova-plugin-fcm-ng": "5.1.1",
     "cordova-plugin-file": "6.0.2",
     "cordova-plugin-globalization": "1.0.9",
-    "cordova-plugin-inappbrowser": "2.0.2",
+    "cordova-plugin-inappbrowser": "git+https://github.com/apache/cordova-plugin-inappbrowser.git",
```
cmgustavo: We should work with stable version of the plugin. it gets from `master` branch (that could be unstable).

JohnathanWhite: This was intentional, the latest stable does not include the postMessage feature we need for inappbrowser to send messages to the parent. It is only available in the master branch(3.1.1-dev).

cmgustavo: That's ok for now, but we can't deploy unstable plugins. I hope they release soon the new version that include it.

bsnowden3:
On iOS I see a couple of UI quirks with the iab browser rendering certain pages. Will send screenshots.

rastajpa:
Hey Johnathan! I was reviewing the code and testing the PR in my browser and on my android device and here are some things I found so far:

**Browser:**
- Every time I refresh the browser, a new window opens to log in to BitPay.
- When I click on home settings get `ERROR TypeError: Cannot read property 'subscribe' of undefined`
- In settings when I click on Connect to my BitPay ID I get `SettingsPage.html:164 ERROR TypeError: this.bitpayIdRef.show is not a function` and nothing happens

**Android:**
- If I go to Settings->Connect to my BitPay ID and then Loggin to BitPay ID I can't return to BitPay App, so I have to close the browser. After close the browser If I click again in Connect to my BitPay ID the app closes, weird.

I don't know why is a hidden feature, maybe because it is in WIP?

Anyway great job so far!

package.json
JohnathanWhite: I raised a request to have that merged into a stable release. Worst case scenario could we fork the branch?

JohnathanWhite: Good news! it is now available in latest stable 3.1.0. I just pushed up the change.

tsconfig.json
```diff
@@ -22,5 +22,6 @@
   "atom": {
     "rewriteTsconfig": false
   },
+  "files": ["./node_modules/cordova-plugin-inappbrowser/types/index.d.ts"],
```
cmgustavo: Maybe we should add `@ionic-native/in-app-browser` that includes the types. So, we don't need to add this to tsconfig

src/pages/settings/bitpay-id/bitpay-id.ts
```diff
@@ -0,0 +1,68 @@
...
+    private navCtrl: NavController,
+    private popupProvider: PopupProvider,
+    private actionSheetProvider: ActionSheetProvider,
+    private changeDetectorRef: ChangeDetectorRef,
+    private translate: TranslateService
+  ) {}
+
+  ionViewDidLoad() {
+    this.userBasicInfo = this.navParams.data;
+    this.changeDetectorRef.detectChanges();
```
cmgustavo: Nice!

src/pages/settings/settings.ts
```diff
@@ -123,6 +200,21 @@ export class SettingsPage {
     });
   }
 
+  ionViewDidLeave() {
+    if (this.bitpayIdRef) {
+      this.bitpayIdRefSubscription.unsubscribe();
```
cmgustavo: What happen if I close the app in this view? The subscription will remain active. Adds "Platform.pause" to catch that event when device enter on background.

cmgustavo:
On which platforms it will work? I think that in-app-browser works different on Electron

rastajpa:
Hey Johnathan! I was reviewing the code and testing the PR in my browser and in my android device and here are some things I found so far:

*Browser:*
- Every time I refresh the browser, a new window opens to log in to BitPay.
- When I click on home settings sometimes I get `ERROR TypeError: Cannot read property 'subscribe' of undefined`
- In settings when I click on Connect to my BitPay ID I get `SettingsPage.html:164 ERROR TypeError: this.bitpayIdRef.show is not a function` and nothing happens

*Android:*
- If I go to Settings->Connect to my BitPay ID and then Loggin to BitPay ID I can't return to BitPay App, so I have to close the browser. After close the browser If I click again in Connect to my BitPay ID the app closes, weird.

We can not merge code that introduce some errors to the app. So, if this PR needs an specific config, please added the steps to the description

Anyway great job!

src/providers/in-app-browser/card.ts
```diff
@@ -0,0 +1,172 @@
...
+    // private actionSheetProvider: ActionSheetProvider,
+    private bitpayIdProvider: BitPayIdProvider,
+    private appIdentityProvider: AppIdentityProvider,
+    private persistanceProvider: PersistenceProvider,
+    private iab: InAppBrowserProvider // private translate: TranslateService
+  ) {
+    this.NETWORK = 'livenet';
+    this.BITPAY_API_URL =
+      this.NETWORK == 'livenet'
+        ? 'https://10.10.10.189:8088'
```
cmgustavo: When this service will be available in production?

tsconfig.json
cmgustavo: Is the plugin `@ionic-native/in-app-browser` an option?

cmgustavo:
Please, run `npm run fix:prettier` in order to fix code issues.

JohnathanWhite:
@cmgustavo Making a new PR that is based off of Gabriel's feat/newDesign (new card tabs ui)

PR #10300, opened 2019-11-10
Ref/create multiple wallets 01

cmgustavo:


src/providers/push-notifications/push-notifications.ts
```diff
@@ -91,7 +91,10 @@ export class PushNotificationsProvider {
       );
       return;
     }
-    this._subscribe(walletClient);
+    if (!_.isArray(walletClient)) walletClient = [walletClient];
```
Gamboster: We usually use `[].concat(walletClient)`

src/providers/profile/profile.ts
```diff
@@ -1389,31 +1388,81 @@ export class ProfileProvider {
...
       bwsurl: defaults.bws.url,
-      singleAddress: opts.singleAddress || false,
-      coin: opts.coin
+      singleAddress: this.currencyProvider.isSingleAddress(coin) || false,
+      coin
     };
+  }
 
-    return this.createWallet(addingNewWallet, defaultOpts);
+  public createDefaultWallet(coins): Promise<any> {
```
Gamboster: Great!

src/providers/push-notifications/push-notifications.ts
cmgustavo: +1!

PR #10304, opened 2019-11-11
FIX: fixes for create a new key flow

rastajpa:


src/pages/settings/settings.html
```diff
@@ -73,7 +73,7 @@ <h4>Learn where you can spend your crypto today.</h4>
         </ion-note>
       </button>
       <button ion-item class="create-button" *ngIf="walletsGroups && walletsGroups[0]" (click)="goToAddView()" translate>
-        Create or Import a Key
```
cmgustavo: I'm not sure on removing "Import"

rastajpa: in the ticket was specified to delete the import option of `configuration -> create or import a key`... if it  won't possible to import, so the label must change

rastajpa: Fixed! Thanks for reviewing it

couby1master:
https://t.me/couby1master

couby1master:
https://t.me/couby1master

PR #10316, opened 2019-11-12
Styles for eth wallet name in token details

Gamboster:
Also fixes `translate`
Copay


couby1master:
https://t.me/couby1master

PR #10331, opened 2019-11-13
Fix(fee) - Use chain for tx.feeStr and feeRatePerStr if Utxo

justinkook:
All miner fees will be based on `chain` so:
- any ERC20 token will have fees in ETH.
- any SLP tokens will have fees in BCH.

Only if utxo based coin:
- `tx.feeRatePerStr`

Also fixes `txp-details.html` & `txp.html`
- Token txp.amount is `0` USDC. Uses `tx.amountStr` instead of `tx.amount` so it displays:

~~0.00 USDC~~ => `0.00 ETH`

### DIFF
--- a/src/pages/send/confirm/confirm.ts
+++ b/src/pages/send/confirm/confirm.ts
@@ -464,7 +464,12 @@ export class ConfirmPage extends WalletTabsChild {
                 tx.feeRate
               } vs. referent level (5 * feeRate) ${maxAllowedFee}`
             );
-            if (tx.network != 'testnet' && tx.feeRate > maxAllowedFee) {
+            const isUtxo = this.currencyProvider.isUtxoCoin(wallet.coin);
+            if (
+              tx.network != 'testnet' &&
+              tx.feeRate > maxAllowedFee &&
+              isUtxo
+            ) {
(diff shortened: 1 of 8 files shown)


src/pages/send/confirm/confirm.ts
```diff
@@ -573,9 +578,11 @@ export class ConfirmPage extends WalletTabsChild {
     return new Promise((resolve, reject) => {
       this.getTxp(_.clone(tx), wallet, opts.dryRun)
         .then(txp => {
-          const per = this.getFeeRate(txp.amount, txp.fee);
-          txp.feeRatePerStr = per.toFixed(2) + '%';
-          txp.feeTooHigh = this.isHighFee(txp.amount, txp.fee);
+          if (this.currencyProvider.isUtxoCoin(tx.coin)) {
```
matiu: ACK

matiu:
Code ACK

src/pages/send/confirm/confirm.ts
couby1master: 🏴

couby1master:
🏴

couby1master:
> LGTM
🏴

PR #10336, opened 2019-11-15
Ref/erc20 create

matiu:
- refactor Token creation
- needs https://github.com/bitpay/bitcore/pull/2560

### DIFF
--- a/src/pages/add/select-currency/select-currency.ts
+++ b/src/pages/add/select-currency/select-currency.ts
@@ -147,14 +147,10 @@ export class SelectCurrencyPage {
     const selectedTokens = _.keys(_.pickBy(this.tokensSelected));
     this.onGoingProcessProvider.set('creatingWallet');
     this.profileProvider
-      .createMultipleWallets(coins)
+      .createMultipleWallets(coins, selectedTokens)
       .then(wallets => {
         this.walletProvider.updateRemotePreferences(wallets);
         this.pushNotificationsProvider.updateSubscription(wallets);
-        const newEthWallet = wallets.find(wallet => wallet.coin === 'eth');
-        if (newEthWallet && selectedTokens.length > 0) {
-          this.createTokenWallet(newEthWallet, selectedTokens);
-        }
         this.endProcess();
       })
       .catch(e => {
@@ -209,25 +205,6 @@ export class SelectCurrencyPage {
       this.endProcess();
     });
   }
-
-  private async addTokenWallet(pairedWallet, token) {
-    const { name, symbol } = token;
-    const { credentials } = _.cloneDeep(pairedWallet);
-    credentials.walletName = name;
-    credentials.coin = symbol.toLowerCase();
-    credentials.token = token;
-    await this.profileProvider.loadAndBindProfile(credentials);
-  }
-
-  private createTokenWallet(wallet, selectedTokens) {
-    const addTokens = this.availableTokens.filter(token =>
-      selectedTokens.includes(token.symbol)
-    );
-    for (const token of addTokens) {
-      this.addTokenWallet(wallet, token);
-    }
-  }
-
   public setTokens(coin?: string): void {
     if (coin === 'eth' || !coin)
       for (const token of this.availableTokens) {
--- a/src/pages/add/select-currency/select-currency.ts
+++ b/src/pages/add/select-currency/select-currency.ts
@@ -111,7 +111,7 @@ export class SelectCurrencyPage {
     });
     modal.present();
     modal.onDidDismiss(() => {
-      this._createWallet(coins);
+      this._createWallets(coins);
     });
     this.persistenceProvider.setKeyOnboardingFlag();
   }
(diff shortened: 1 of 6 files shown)


src/providers/profile/profile.ts
```diff
@@ -1438,60 +1426,83 @@ export class ProfileProvider {
...
+                datas.unshift(firstWalletData);
+                let walletClients = _.map(datas,'walletClient');
+
+                // Handle tokens
+                const ethWalletClient = walletClients.find(wallet => wallet.credentials.coin === 'eth');
+                let tokenCreate  = [];
+                if (ethWalletClient) {
+                  tokenCreate = tokens.map(token => this.createTokenWallet(ethWalletClient, token));
+                }
+                Promise.all(tokenCreate).then( tokenClients => {
```
cmgustavo: Since `createTokenWallet` is not an async function, I think that "Promise.all" is not necessary

src/pages/add/select-currency/select-currency.ts
```diff
@@ -184,61 +177,31 @@ export class SelectCurrencyPage {
...
     walletSelector.present();
-    walletSelector.onDidDismiss(wallet => {
-      if (!_.isEmpty(wallet)) {
-        this.addTokenWallet(wallet, token);
-        this.endProcess();
+    walletSelector.onDidDismiss(pairedWallet => {
+      if (!_.isEmpty(pairedWallet)) {
+        this.addTokenWallet(pairedWallet, token);
       }
+      this.endProcess();
```
justinkook: `this.endProcess()` needs to be inside the if statement or the app will go back to the home screen on `addTokenWallet` info-sheet dismiss even if you dont select an ETH wallet.

src/providers/profile/profile.ts
matiu: thanks! fixing.

src/providers/profile/profile.ts
```diff
@@ -1438,60 +1428,74 @@ export class ProfileProvider {
...
+      bp_partner: ethWallet.bp_partner,
+      bp_partner_version: ethWallet.bp_partner_version
+    });
+    walletClient.fromObj(tokenCredentials);
+    return walletClient;
+  }
+
+  public createMultipleWallets(coins: string[], tokens = []): Promise<any> {
     return new Promise((resolve, reject) => {
+      if (tokens && tokens.length && coins.indexOf('eth') <= 0) {
```
justinkook: if only eth is selected then:
```
coins = ['eth']` then index will be `0`. 
```

Screen Shot 2019-11-15 at 1 26 54 PM

justinkook: should be `tokens.length = 0 && && coins.indexOf('eth') < 0`

src/pages/add/select-currency/select-currency.ts
```diff
@@ -187,66 +182,64 @@ export class SelectCurrencyPage {
...
       }
     );
     walletSelector.present();
-    walletSelector.onDidDismiss(wallet => {
-      if (!_.isEmpty(wallet)) {
-        this.addTokenWallet(wallet, token);
-        this.endProcess();
-      }
+    walletSelector.onDidDismiss(pairedWallet => {
+      return this.createAndBindTokenWallet(pairedWallet, token);
```
justinkook: Need for dismissal if wallet is not selected:
```
 if (!_.isEmpty(wallet)) { ... }
```

src/pages/add/select-currency/select-currency.ts
```diff
@@ -187,66 +182,64 @@ export class SelectCurrencyPage {
...
-      .filter(wallet => wallet.coin === token.symbol.toLowerCase())
-      .map(wallet => wallet.id);
-    const availableWallets = this.wallets.filter(
-      wallet =>
-        !tokenWalletIds.includes(`${wallet.id}-${token.address}`) &&
-        wallet.coin === 'eth'
-    );
+    if (eligibleWallets.length == 1) {
+      return this.createAndBindTokenWallet(eligibleWallets[0], token);
+    }
```
justinkook: Remove this `if block`
Need to show wallet selector even if there is only 1 eth wallet available to confirm that the user is creating a token wallet and to confirm that token wallet is being linked to ETH wallet.

src/pages/add/select-currency/select-currency.ts
```diff
@@ -187,66 +182,64 @@ export class SelectCurrencyPage {
...
+        token
+      );
+      this.profileProvider
+        .addAndBindWalletClient(tokenWalletClient)
+        .then(() => {
+          this.endProcess();
+        });
+    } else {
+      this.endProcess();
+    }
```
justinkook: remove else statement

Causes UI bug on info-sheet dismiss.

src/providers/profile/profile.ts
cmgustavo: I also got that popup too, but the wallets (ETH and USDC selected) were created anyway

matiu: thanks!

      if (tokens && tokens.length && coins.indexOf('eth') < 0) {

if should fail if no `eth` is given a parameter and one or more token is given.

PR #10350, opened 2019-11-20
[FEAT] Handle bip21 like format (bitpay-uri)

gabrielbazan7:


src/providers/incoming-data/incoming-data.spec.ts
```diff
@@ -902,6 +902,25 @@ describe('Provider: Incoming Data Provider', () => {
...
+          expectedType: 'BitPayUri'
+        },
+        {
+          data:
+            'bitpay:mrNYDWy8ZgmgXVKDb4MM71LmfZWBwGztUK?coin=btc&amount=0.0002&message=message',
+          expectedType: 'BitPayUri'
+        },
+        {
+          data:
+            'bitpay:0xDF5C0dd7656bB976aD7285a3Fb80C0F6B9604576?amount=1543000000000000000?gasPrice=0000400000000000000',
```
matiu: in this case... coin is undef... what would happen?

Mvfaassen: Wallet will open but will not redirect to payflow. OR, wallet will pick the last coin provided in the sendflow.

gabrielbazan7: Just add some tests for that. If coin or address are missing show it as invalid on SendPage and show it as plain-text on ScanPage.

cmgustavo:
Needs rebase

PR #10360, opened 2019-11-22
Log screen events

bsnowden3:


src/providers/analytics/analytics.ts
```diff
@@ -17,4 +17,8 @@ export class AnalyticsProvider {
     if (this.platformProvider.isCordova)
       this.FCMPlugin.setUserProperty(name, value);
   }
+
+  setScreenName(screenName: string) {
```
cmgustavo: Just an idea: instead to create many `logEvent` with the name of the screen, why don't we do only 1 event and send the name of the screen by params? Something like this: 
```
this.FCMPlugin.logEvent('screenName', { page: screenName });
``` 
On this way, we're going to track only one event but different params. I think it's easier than track many eventName.

PR #10378, opened 2019-11-27
fix(profile): set true to store profile in addAndBindWalletClient

cmgustavo:


src/providers/profile/profile.ts
```diff
@@ -753,12 +753,13 @@ export class ProfileProvider {
   // Adds and bind a new client to the profile
   private async addAndBindWalletClient(
     wallet,
-    opts: WalletBindTypeOpts = { bwsurl: null, store: true }
+    opts?: WalletBindTypeOpts
```
justinkook: I think this should be changed to `opts: WalletBindTypeOpts = {}` so it doesnt throw an error when we try to destructure `opts.bwsUrl / opts.store`

The error would be `cannot read property store of undefined`

cmgustavo: Done. Thanks!

PR #10417, opened 2019-12-10
Disable openWalletName click if ERC20Token wallet

justinkook:
related MR:

https://github.com/bitpay/bitcore/pull/2629

### DIFF
--- a/src/pages/settings/wallet-settings/wallet-settings.html
+++ b/src/pages/settings/wallet-settings/wallet-settings.html
@@ -6,12 +6,20 @@
 
 <ion-content>
   <ion-list class="settings-list bp-list">
-    <button ion-item (click)="openWalletName()">
+    <button ion-item *ngIf="!isERCToken; else disabledClick" (click)="openWalletName()">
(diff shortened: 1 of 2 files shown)


src/pages/settings/wallet-settings/wallet-settings.html
```diff
@@ -6,7 +6,7 @@
 
 <ion-content>
   <ion-list class="settings-list bp-list">
-    <button ion-item (click)="openWalletName()">
+    <button ion-item (click)="isERCToken ? '' : openWalletName()">
```
gabrielbazan7: maybe this is not a good solution since still a button ( with all the buttons effects ) that do nothing. What do you think ?

justinkook: yes I agree, what about this:

```
<div ion-item *ngIf="isERCToken ? ''>
      <div translate float-left>Name</div>
      <ion-note item-end>
        {{wallet && wallet.name}}
      </ion-note>
</div>
```

PR #10419, opened 2019-12-10
Fix token walletName automatically updates

justinkook:
- Get correct token Wallet using token walletId to update walletName


rastajpa:
Tests needs to be fixed

PR #10420, opened 2019-12-11
Remove single ETH wallet condition in addTokenWallet

justinkook:
Remove single ETH wallet case in addTokenWallet sheet


matiu:
why?

justinkook:
> why?

Its so users can confirm which ETH wallet they are linking with and check balance. Some wallet names are the same as the default ethereum walletName is `Ethereum`

PR #10423, opened 2019-12-12
[FIX] show error if no wallets with funds on confirm view

gabrielbazan7:


src/pages/send/confirm/confirm.ts
```diff
@@ -294,31 +294,30 @@ export class ConfirmPage extends WalletTabsChild {
...
-        network,
-        coin
-      });
 
-      return resolve();
+    this.wallets = this.profileProvider.getWallets({
+      onlyComplete: true,
+      hasFunds: true,
+      network,
+      coin
```
justinkook: shouldn't we be adding `minAmount` here as one of the opts, so the wallet selector will not show any non eligible wallets?

justinkook:
```
this.wallets = this.profileProvider.getWallets({
      onlyComplete: true,
      hasFunds: true,
      network,
      coin,
      minAmount <----
```

gabrielbazan7: mmm personally I don't like that solution, I want to see all my wallets and check balances of each of them. Not showing the wallets might cause confusion on users that doesn't know why they can't see their wallets in case the first wallet have enough funds but the second no ( the second one will not be shown with no explanation )

src/pages/send/confirm/confirm.ts
```diff
@@ -337,6 +336,7 @@ export class ConfirmPage extends WalletTabsChild {
 
     if (this.tx.paypro) this.paymentTimeControl(this.tx.paypro.expires);
 
+    const goBack = this.wallets.length > 1 ? false : true;
```
justinkook: `const goBack = this.wallets.length > 0`

gabrielbazan7: The idea here is that if you have more than 1 wallet, you will have the opportunity to choose another if you do not have enough to pay

src/pages/send/confirm/confirm.ts
```diff
@@ -294,31 +294,30 @@ export class ConfirmPage extends WalletTabsChild {
...
+      network,
+      coin
     });
+
+    if (_.isEmpty(this.wallets)) {
+      const msg = this.translate.instant(
+        'You are trying to send more funds than you have available. Make sure you do not have funds locked by pending transaction proposals.'
+      );
+      const title = this.translate.instant('No wallets available');
+      return Promise.reject({ msg, title });
```
justinkook: the only time you will ever reach this condition is when the wallet balance is `0` due to the flag:

`hasFunds: true`

You can have a transaction with amount 10000 ETH with only 1 ETH wallet balance, but `this.wallets` will not be empty.

justinkook:
LGTM 👍 

Screen Shot 2019-12-13 at 1 28 07 PM

PR #10431, opened 2019-12-18
[REF] do not request encryption if wallets were already created without it

gabrielbazan7:


src/providers/profile/profile.ts
```diff
@@ -674,8 +674,14 @@ export class ProfileProvider {
 
   private askToEncryptKey(key): Promise<any> {
     if (!key) return Promise.resolve();
+    // if the key is already encrypted, keep it that way for new wallets
     if (key.isPrivKeyEncrypted()) return Promise.resolve();
 
+    // do not request encryption if wallets were already created without it
+    const wallets = this.getWallets({ keyId: key.id });
```
cmgustavo: At this point, you have already created the wallets.

cmgustavo: So, it will be true

gabrielbazan7: bindWalletClient function is called after the key encryption

cmgustavo: Ok, you're right.

PR #10471, opened 2020-01-10
Updated UI for XRP minimum balance design

justinkook:
### A/B testing version 2

- wallet-details shows total balance instead of spendable/available balance
- add info icon to open breakdown of balance details.

Screen Shot 2020-01-14 at 11 58 50 AM

- changed wording and help text on XRP Minimum Balance flow:

Screen Shot 2020-01-10 at 4 33 24 PM

### DIFF
--- a/src/pages/wallet-details/wallet-details.html
+++ b/src/pages/wallet-details/wallet-details.html
@@ -33,22 +33,11 @@
 
         <div (longPress)="toggleBalance()">
           <div (tap)="updateAll(true)" *ngIf="!wallet.balanceHidden && !wallet.scanning && !walletNotRegistered && (wallet.cachedStatus || wallet.lastKnownBalance)">
-            <div *ngIf="wallet.coin !== 'xrp'; else xrpBalance">
-              <div class="balance-str">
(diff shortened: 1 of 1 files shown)


src/pages/wallet-details/wallet-details.html
```diff
@@ -33,22 +33,13 @@
 
         <div (longPress)="toggleBalance()">
           <div (tap)="updateAll(true)" *ngIf="!wallet.balanceHidden && !wallet.scanning && !walletNotRegistered && (wallet.cachedStatus || wallet.lastKnownBalance)">
-            <div *ngIf="wallet.coin !== 'xrp'; else xrpBalance">
+            <div (click)="wallet.coin === 'xrp' ? openBalanceDetails() : ''">
```
cmgustavo: I think we can't have `(tap)` and `(click)` into the same scope. It could have a weird behavior. Maybe the function should be relocated as an (i)nfo icon below.

justinkook: added info icon for click event. Thanks!

vanhue:
@justinkook Could we also include the updated icon?
Screenshot 2020-01-14 12 23 10

PR #10481, opened 2020-01-17
Fix gift card purchases with fiat-pegged blockchain tokens

msalcala11:
Use invoice rates in `ConfirmCardPurchasePage` calculations, and fix non-USD gift card purchases with fiat-pegged blockchain tokens.


cmgustavo:
Needs rebase

PR #10483, opened 2020-01-20
Feat/add card eu

JohnathanWhite:
This PR adds bitpay ID as well as the in app browser functionality which will point to the new card onboarding/dashboard.

- Added the base bitpayID template to settings
- Created utils/iab-patch.js that runs post plugin install and sets the status bar height on ios to 0. This is needed to achieve full screen on notched phones (iphoneX etc). Also, added a java override to allow for onfido camera access.
- Added the iab provider that allows for generating/accessing iab instances.
- Added method generatePairingToken in BitPayID provider that takes the pairing code passed from the iab and creates/persist a token and user info into persistenceStorage.
- Added iab card provider to handle messages sent from card iab
- Added a generic graph query handler - this takes a postMessage payload from the iab with a query and query variables and signs the request. It then passes it back to the iab where the call is made.
- Added load error logic to the iab - if a load failure occurs, the show method of the iab gets overwritten to execute a default error action sheet.
- Added a card specific callback hook to the payment handler. If context is card then the callback reopens the iab and executes a postMessage to update upon completion.


package.json
```diff
@@ -142,7 +143,7 @@
     "cordova-plugin-fcm-ng": "5.1.1",
     "cordova-plugin-file": "6.0.2",
     "cordova-plugin-globalization": "1.0.9",
-    "cordova-plugin-inappbrowser": "2.0.2",
+    "cordova-plugin-inappbrowser": "3.1.0",
```
cmgustavo: Maybe you're missing to add this version into `app-template` also

JohnathanWhite: good catch just updated

cmgustavo:
Code looks good! Just run `npm run fix` to fix ts lint and prettier issues

PR #10535, opened 2020-01-24
Feature/litecoin

justinkook:
- Need to fix `getCoinAndNetwork` for `address.ts`

BTC testnet address are same as LTC testnet address.


losh11:
any reason why this was closed? just needs a rebase right?

PR #10536, opened 2020-01-27
Fix UI confirm view, simplex, txp-details and proposals.

cmgustavo:


src/pages/wallets/proposals-notifications/proposals-notifications.html
```diff
@@ -34,7 +34,7 @@
   </div>
 </wide-header-page>
 <ion-footer>
-  <page-slide-to-accept #slideButton *ngIf="isCordova" [disabled]="!txpsToSign[0]" [buttonText]="buttonText" (slideDone)="sign()"></page-slide-to-accept>
+  <page-slide-to-accept #slideButton *ngIf="isCordova && txpsToSign[0]" [disabled]="!txpsToSign[0]" [buttonText]="buttonText" (slideDone)="sign()"></page-slide-to-accept>
```
Gamboster: I think we should remove one of this controls: *ngIf="txpsToSign[0]" or [disabled]="!txpsToSign[0]"

cmgustavo: I prefer to remove `disabled` but it's necessary inside "ngIf". Will remove from [disabled] property. Thanks

PR #10564, opened 2020-01-30
feat(address-book): adds coin-icon to contact list. Ref option-sheet.…

cmgustavo:
… Increase size to gravatar

Screen Shot 2020-01-30 at 14 56 17

Screen Shot 2020-01-30 at 14 58 19


vanhue:
Nice work

PR #10579, opened 2020-02-03
FIX: fixes for ios devices

rastajpa:


src/components/action-sheet/action-sheet.html
```diff
@@ -1,4 +1,4 @@
 <div class="backdrop" [ngClass]="{'backdrop-fade-in': slideIn}" (click)="dismiss()"></div>
-<div class="action-sheet" [ngClass]="{'opened': slideIn, 'closed-top': !slideIn && fromTop, 'closed-bottom': !slideIn && !fromTop, 'action-sheet-top': fromTop, 'action-sheet-bottom': !fromTop }">
+<div class="action-sheet" [ngClass]="{'opened': slideIn, 'closed-top': !slideIn && fromTop, 'closed-bottom': !slideIn && !fromTop, 'action-sheet-top': fromTop, 'action-sheet-bottom': !fromTop, 'safe-top-area': fromTop }">
```
cmgustavo: it's too complex to read. Why not group class with same condition?

src/components/action-sheet/action-sheet.scss
```diff
@@ -61,4 +61,8 @@ action-sheet {
       transform: translateY(calc(-100% + 8px));
     }
   }
+  .safe-top-area {
+    padding-top: calc(10px + env(safe-area-inset-top));
```
cmgustavo: 10px is applied to desktop too. Is that the idea?

src/components/action-sheet/action-sheet.html
rastajpa: you're right!

src/components/action-sheet/action-sheet.scss
rastajpa: it is on purpose. It looks more similar to the design

PR #10592, opened 2020-02-06
[FEAT] bp qr code

gabrielbazan7:


package.json
```diff
@@ -212,7 +213,7 @@
     "ts-node": "5.0.0",
     "tslint": "5.8.0",
     "tslint-config-prettier": "1.7.0",
-    "typescript": "2.6.2",
+    "typescript": "2.9.2",
```
cmgustavo: Be careful with upgrading typescript. It has a very close dependency with webpack for compilation (in prod mode).

src/components/bp-qr-code-component/bp-qr-code-component.ts
```diff
@@ -0,0 +1,19 @@
+import { Component, Input } from '@angular/core';
+import { CurrencyProvider } from '../../providers/currency/currency';
+
+@Component({
+  selector: 'bp-qr-code-component',
+  templateUrl: 'bp-qr-code-component.html'
+})
+export class BpQrCodeComponent {
```
cmgustavo: Since we're installing a plugin from npm, it shouldn't be necessary. The plugin should have all the component elements to work. Try to isolate these functions into the plugin.

gabrielbazan7: Maybe more than a component I should add this to the includes folder.

gabrielbazan7: My main problem is that for some reason I cannot use a ts variable for module-color, position-ring-color, position-center-color properties

justinkook:
👍  LGTM, all qr codes scanned!

BTC qr code has some delays but is able to scan.

PR #10599, opened 2020-02-09
Feature - Price charts using Apex Charts

justinkook:
relies on

https://github.com/bitpay/copay/pull/10641


cmgustavo:
Please, don't add `package-lock.json`. We'll do it on a new bump (next version).

justinkook:
> There is an error regarding currencies supported by Simplex.
> This is the correct listing.
> `AUD CAD CHF CZK DKK EUR HUF ILS JPY KRW NOK NZD PLN RUB SEK TRY USD ZAR`
> You can bring it from simplexProvider

fixed!

gabrielbazan7:
ACK!!

Gamboster:
Tested on my android. It looks good and the error with the currencies of Simplex was fixed.

Just to know... Why do we change from library chart.js to apexchart?

justinkook:
You need a plugin to perform the same touch actions/animations as apex charts.

https://chartjs-plugin-datalabels.netlify.com/guide/

Also animations were better for apexcharts

cmgustavo:
I'd prefer to disable this feature (select range). Is that possible?

Screen Shot 2020-03-16 at 14 45 00

rastajpa:
needs rebase

PR #10625, opened 2020-02-14
[FEAT]: tag scan option for XRP

rastajpa:


justinkook:
Tested confirm screen + scan

PR #10639, opened 2020-02-19
[REF] translate to cashaddr if legacy

gabrielbazan7:


src/providers/address/address.ts
```diff
@@ -19,6 +19,11 @@ export class AddressProvider {
     this.core = this.bwcProvider.getCore();
   }
 
+  public translateToCashAddress(addressToTranslate: string): string {
```
cmgustavo: Could you add a log like this: Converted: "legacyAddress" -> "newCashAddress"

gabrielbazan7: yes!

PR #10647, opened 2020-02-20
remove button from ion toogle

gabrielbazan7:


src/theme/lists.scss
```diff
@@ -12,9 +12,6 @@
   &.top-padding {
     margin-top: 40px;
   }
-  ion-item {
-    padding-left: 19px;
```
Gamboster: I suggest leaving this and putting `padding-left: 16px;` for those specific items in the settings css

PR #10667, opened 2020-02-25
Addr Book: Scanner option to detination tag

rastajpa:


rastajpa:
> Can we add Validator.regex `const regex = /^[0-9]{9,}$/;` to:
> 
> https://github.com/bitpay/copay/pull/10667/files#diff-bedfadb7f4f0a5b0b131680f43c59f85R56

Done in the last commit

PR #10687, opened 2020-03-02
Fix/revert to wk

JohnathanWhite:
- reverts UIWebview to WK
- adds check for electron to prevent IAB from opening if on desktop version


config.xml
```diff
@@ -0,0 +1,163 @@
+<?xml version='1.0' encoding='utf-8'?>
```
cmgustavo: I don't know why this file isn't in the gitignore list. You shouldn't add this file due to it's overwritten on every `apply:xx` by `app-template/config-template.xml.

msalcala11:
ACK

config.xml
JohnathanWhite: That was my mistake! I just removed it.

PR #10691, opened 2020-03-03
[FEAT]: Dark Mode

rastajpa:


src/theme/dark.scss
```diff
@@ -0,0 +1,934 @@
+.dark-theme {
+  // App background color
+  $background-color-dark-theme: #000000;
```
cmgustavo: Just a recommendation: avoid pure black color. See Material Design recommends: https://material.io/design/color/dark-theme.html

src/theme/dark.scss
```diff
@@ -0,0 +1,934 @@
...
+      }
+      .item-subtitle {
+        color: color($colors, light-grey);
+      }
+    }
+  }
+  page-import-wallet,
+  page-create-wallet,
+  page-bitpay-phase-one-card-intro {
+    ion-item {
```
cmgustavo: We should cover all `ion-item` elements, not for specific page. Example of missing dark format (list of item from shapeshift): 

Screen Shot 2020-03-10 at 16 47 24

src/theme/variables.scss
```diff
@@ -129,6 +129,9 @@ $button-md-clear-background-color-hover: transparent;
 // so that the above variables are used and Ionic's
 // default are overridden.
 @import 'ionic.theme.default';
+@import 'dark';
+@import 'light';
+
```
cmgustavo: Must define standard ion-button color=dark -> (invert color)

Screen Shot 2020-03-10 at 16 49 57

src/app/app.component.ts
```diff
@@ -217,7 +220,7 @@ export class CopayApp {
       // Only overlay for iOS
       if (this.platform.is('ios')) {
         this.statusBar.overlaysWebView(true);
-        this.statusBar.styleDefault();
+        this.themeProvider.useDefaultStatusBar(true);
```
cmgustavo: `cordova-plugin-statusbar` detect automatically dark mode. I think it's not necessary to make changes. Also the app is forcing that to use "light" mode: https://github.com/bitpay/copay/blob/master/app-template/config-template.xml#L111

src/theme/dark.scss
rastajpa: I agree!

src/theme/dark.scss
rastajpa: It would be enough to add the 'bp-list' class to <ion-list>. Good catch!

gabrielbazan7:
ACK!! great job !!

src/providers/theme/theme.ts
```diff
@@ -0,0 +1,123 @@
+import { Injectable } from '@angular/core';
+import { StatusBar } from '@ionic-native/status-bar';
+import { BehaviorSubject } from 'rxjs/BehaviorSubject';
```
cmgustavo: Maybe we could remove it, since the value doesn't need a history. What do you think?

src/providers/theme/theme.ts
```diff
@@ -0,0 +1,123 @@
...
+    private logger: Logger,
+    private persistenceProvider: PersistenceProvider,
+    private statusBar: StatusBar,
+    private platformProvider: PlatformProvider
+  ) {
+    this.logger.debug('ThemeProvider initialized');
+  }
+
+  public init() {
+    return this.persistenceProvider.getAppTheme().then(theme => {
```
cmgustavo: `Return`  is not needed here.

src/providers/theme/theme.ts
```diff
@@ -0,0 +1,123 @@
...
+  constructor(
+    private logger: Logger,
+    private persistenceProvider: PersistenceProvider,
+    private statusBar: StatusBar,
+    private platformProvider: PlatformProvider
+  ) {
+    this.logger.debug('ThemeProvider initialized');
+  }
+
+  public init() {
```
cmgustavo: I think it's not necessary since we the `persistentProvider` is loaded before of everything . Maybe it's a good practice to set the variable in the constructor.

Gamboster: We can’t move the init function to the constructor, since the themeProvider is loaded at the beginning of the app, being in the app component. Moving it to the constructor doesn’t give time to fully prepare persistanceProvider, so we get a storage related error.

src/providers/theme/theme.ts
Gamboster: done

src/providers/theme/theme.ts
Gamboster: done. Thanks!

src/providers/theme/theme.ts
cmgustavo: Got it. Thanks for explanation.

PR #10696, opened 2020-03-04
[FEAT] Clear encrypt password option

gabrielbazan7:


src/providers/key/key.ts
```diff
@@ -276,4 +278,11 @@ export class KeyProvider {
 
     return key.sign(rootPath, txp, password);
   }
+
+  public match(key1, key2) {
```
cmgustavo: It should be a private function? I'd change the name by another non-reserved word

gabrielbazan7: Is public I use it here: https://github.com/bitpay/copay/pull/10696/files#diff-a287f4baee716e25e5027889f0b17d93R903

But, I change the function name thanks!

PR #10705, opened 2020-03-06
added optional code; added error handling if failed pairing

JohnathanWhite:
This PR adds an optional 2fa code if present and also handles failed pairings more gracefully by closing the IAB and showing an error-default bottom sheet.

### DIFF
--- a/src/providers/in-app-browser/card.ts
+++ b/src/providers/in-app-browser/card.ts
@@ -207,17 +207,26 @@ export class IABCardProvider {
          * */
 
         case 'pairing':
-          const { secret } = event.data.params;
           // generates pairing token and also fetches user basic info and caches both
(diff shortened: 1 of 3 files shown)


src/providers/in-app-browser/card.ts
```diff
@@ -207,17 +207,26 @@ export class IABCardProvider {
...
-            secret,
+            event.data.params,
             (user: User) => {
               if (user) {
                 this.getCards();
                 this.user.next(user);
               }
             },
-            err => this.logger.debug(err)
+            () => {
```
cmgustavo: Sometimes is a good idea to add the `e` variable here and show in console.log, in spite of we show another message to the user

JohnathanWhite: added a log.error!

PR #10713, opened 2020-03-10
[FEAT] new release msg

gabrielbazan7:
BitPay


vanhue:
@gabrielbazan7 will there be a way to dismiss that message?

gabrielbazan7:
> @gabrielbazan7 will there be a way to dismiss that message?

No, the user will need to update the app to make the message disappear. But I'm not sure, what do you think ?

vanhue:
If it's only for desktop then it should be fine. 👍

cmgustavo:
It should be possible to close.... so waiting for it

PR #10717, opened 2020-03-12
Fix UI for email-component

justinkook:
Screen Shot 2020-03-12 at 12 20 14 PM


vanhue:
Nice work @justinkook!

PR #10722, opened 2020-03-16
[FIX] fee level for giftcards and bpcards

gabrielbazan7:


src/pages/integrations/bitpay-card/bitpay-card-topup/bitpay-card-topup.ts
```diff
@@ -515,10 +515,6 @@ export class BitPayCardTopUpPage {
                 );
                 let newAmountSat = maxAmountSat - invoiceFeeSat;
 
-                // Set expiration time for this invoice
```
cmgustavo: Why it's removed?

gabrielbazan7: That's the expiration time of the first invoice created that it's not needed

cmgustavo: Good!

cmgustavo:
Rebase is needed

PR #10723, opened 2020-03-16
[FIX] Incoming-data: Btc/Bch URI - paypro parameter

gabrielbazan7:


src/providers/incoming-data/incoming-data.ts
```diff
@@ -329,8 +329,10 @@ export class IncomingDataProvider {
     let address = parsed.address ? parsed.address.toString() : '';
     let message = parsed.message;
     let amount = parsed.amount || amountFromRedirParams;
-    if (parsed.r) this.goToPayPro(data, coin);
-    else this.goSend(address, amount, message, coin);
+    if (parsed.r) {
+      const payProUrl = this.getPayProUrl(parsed.r);
```
cmgustavo: We should do the same for Bitcoin Cash, shouldn't we?

gabrielbazan7: Done. Thanks

PR #10740, opened 2020-03-20
[FEAT] misc updates

JohnathanWhite:
misc updates 

- selecting a card in the settings tab opens the iab and deep links to the cards settings
- if in the middle of a topup and back is pressed from the amount page the iab repopens to previous state
- updated card intro page
- added badge for card tab
- added hiding of add bitpay card based on condition


utils/iab-patch.js
```diff
@@ -22,7 +22,7 @@ try {
   const content = fs
     .readFileSync(file, 'utf8')
     .split('#define    LOCATIONBAR_HEIGHT 21.0');
-  if (content.includes('20.0')) {
+  if (content[0].includes('20.0')) {
```
cmgustavo: Good catch!

src/theme/icons.scss
```diff
@@ -77,6 +77,10 @@
   color: #464d5f;
 }
 
+.tab-badge {
```
cmgustavo: It's already fixed here (not sure if it applies): https://github.com/bitpay/copay/pull/10738/files#diff-3c303face59d40ba5031ed6c51449334R793

src/pages/integrations/bitpay-card/bitpay-card-home/bitpay-card-home.ts
```diff
@@ -16,18 +20,29 @@ import { PhaseOneCardIntro } from '../bitpay-card-phases/phase-one/phase-one-int
...
     private navCtrl: NavController,
-    private iabCardProvider: IABCardProvider
+    private iabCardProvider: IABCardProvider,
+    private events: Events,
+    private persistenceProvider: PersistenceProvider
   ) {}
 
   ngOnInit() {
     this.appName = this.appProvider.info.userVisibleName;
+    this.events.subscribe('reachedCardLimit', () => {
```
cmgustavo: I'm not sure this event was triggered. The `publish` action could be triggered before the subscriber (since it's in a provider). Would you mind to explain the purpose of the `events` here?

Also, any events should be unsubscribed to avoid duplicated subscribers.

JohnathanWhite: So the publish comes from card.ts and cards.ts when cards are added. This component is a singleton its on init is only ever called once. I added the event to update it in case you disconnected your bitpay id then re connected it. This way it has a default state from the persistence provider and any changes that happen after it has a subscribe thats ready to update it if changes occur. The alternative was to create an observable off of the card items but this seemed easier.

cmgustavo: Agreed. I prefer an event than observable too. Thanks!

PR #10751, opened 2020-03-26
[FEAT]: Show legacy addr format for BCH wallets

rastajpa:


matiu:
Huge NAK.

matiu:
Huge ACK :)

PR #10753, opened 2020-03-27
Updating the condition

maektwain:
This integration has no effect from the gift card settings in the app-template


gabrielbazan7:
done here: https://github.com/bitpay/copay/pull/10759

PR #10759, opened 2020-03-30
[REF] set integrations

gabrielbazan7:


src/pages/home/home.ts
```diff
@@ -212,6 +213,7 @@ export class HomePage {
 
     this.homeIntegrations = _.remove(integrations, x => {
       this.showBuyCryptoOption = x.name == 'simplex' && x.show == true;
+      this.showShoppingOption = x.name == 'giftcards' && x.show == true;
```
cmgustavo: This function is iterative, so the value could be false if the latest element is not `giftcards`

gabrielbazan7:
updated

PR #10779, opened 2020-04-06
[FIX]: Delete txp if speed up tx fails

rastajpa:


bitschmidty:
@cmgustavo @rastajpa Is this regarding CPFP functionality?

PR #10786, opened 2020-04-07
[FEAT] Handle BitPay redir link

gabrielbazan7:


src/providers/incoming-data/incoming-data.ts
```diff
@@ -609,6 +614,25 @@ export class IncomingDataProvider {
     }
   }
 
+  private goToBitPayRedir(data: string): void {
+    this.logger.debug('Incoming-data (redirect): BitPay Redir');
+    const redir = data.replace('bitpay://landingt/', '');
```
cmgustavo, line 619: Is `landingt` ok? or just `landing`

gabrielbazan7: just landing

PR #10800, opened 2020-04-13
Adding BUSD

micahriggan:
Needs a package.json update after BWC changes


src/providers/currency/token.ts
```diff
@@ -23,5 +23,11 @@ export const TokenOpts = {
     symbol: 'GUSD',
     decimal: 2,
     address: '0x056fd409e1d7a124bd7017459dfea2f387b6d5cd'
+  },
+  '0x4Fabb145d64652a948d72533023f6E7A623C7C53': {
```
micahriggan, line 27: should be lowercased

src/providers/currency/token.ts
```diff
@@ -23,5 +23,11 @@ export const TokenOpts = {
...
+    name: 'Binance USD Coin',
+    symbol: 'BUSD',
+    decimal: 18,
+    address: '0x4Fabb145d64652a948d72533023f6E7A623C7C53'
```
micahriggan, line 31: ^^

PR #10807, opened 2020-04-15
fix(android): adds timeout for createInvoice if 403 error occurs

cmgustavo:


src/providers/gift-card/gift-card.ts
```diff
@@ -144,16 +146,20 @@ export class GiftCardProvider extends InvoiceProvider {
     const promise = shouldSync
```
matiu, line 146: Please replace this with an `if`, for clarity.
```
let promise;
if (shouldSync) {
 promise=...
} else {
 //
}

```

PR #10832, opened 2020-04-21
ref(content): Extend swipe down to refresh. Disable home refresh. Dis…

cmgustavo:
…able bounce in all pages

### DIFF
--- a/src/pages/settings/key-settings/key-settings.html
+++ b/src/pages/settings/key-settings/key-settings.html
@@ -8,7 +8,7 @@
   </ion-navbar>
 </ion-header>
 
-<ion-content #scrollArea>
+<ion-content #scrollArea no-bounce class="add-bottom-safe-area">
   <div class="wrapper">
     <expandable-header [scrollArea]="scrollArea" [fadeFactor]="5" [disableFade]="true" #expandableHeader>
       <ion-toolbar class="wide-header__title">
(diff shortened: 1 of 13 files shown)


src/pages/settings/key-settings/key-settings.html
```diff
@@ -1,22 +1,5 @@
...
-<ion-content #scrollArea>
-  <div class="wrapper">
-    <expandable-header [scrollArea]="scrollArea" [fadeFactor]="5" [disableFade]="true" #expandableHeader>
-      <ion-toolbar class="wide-header__title">
-        <expandable-header-primary>
-          {{canSign ? ('Key Settings' | translate) : ('Read Only Wallets' | translate) }}
-        </expandable-header-primary>
-      </ion-toolbar>
-    </expandable-header>
+<wide-header-page title="{{canSign ? ('Key Settings' | translate) : ('Read Only Wallets' | translate) }}">
```
Gamboster, line 1: Did you test if the wallet reorder still works correctly by making this change?

cmgustavo: Uhm.. nop. Fixing that now. Thanks!

cmgustavo: Updated. Reverted changes and added safe-bottom-area manually

cmgustavo:
Updated.

PR #10833, opened 2020-04-22
fix(qr-code): Removes shake when creating new address. Removes border…

cmgustavo:
…-line toolbar iOS. Align top icons

TODO:

- [x] Fix dark-mode on copayers-view


cmgustavo:
Updated.

PR #10837, opened 2020-04-23
[FIX]: Import wallets for an existing key

rastajpa:


src/providers/key/key.ts
```diff
@@ -285,4 +285,8 @@ export class KeyProvider {
       return key1.fingerPrint === key2.fingerPrint;
     else return key1.id === key2.id;
   }
+
+  public getKeyByFingerprint(key) {
```
matiu, line 289: Given "isMatch" will check first fingerPrint and then id, 
maybe a better will be "getMatchedKey"

src/providers/profile/profile.ts
```diff
@@ -926,6 +926,14 @@ export class ProfileProvider {
     this.logger.info('Importing Wallet xPrivKey');
     opts.xPrivKey = xPrivKey;
     return this.serverAssistedImport(opts).then(data => {
+      const key = this.keyProvider.getKeyByFingerprint(data.key);
+      if (key) {
```
matiu, line 930: a comment would help help. // If we have the key already, bind the new wallets to the current key.

matiu:
LGTM, left some minor comments.

src/providers/key/key.ts
rastajpa: getMatchedKey looks better

PR #10856, opened 2020-04-29
[REF] avoid get coins request if no utxo coin

gabrielbazan7:


src/providers/wallet/wallet.ts
```diff
@@ -886,7 +886,7 @@ export class WalletProvider {
       // no future transactions...
       if (tx.time > now) tx.time = now;
 
-      if (tx.confirmations === 0) {
+      if (tx.confirmations === 0 && wallet.coin === 'btc') {
         const coins = await this.getCoinsForTx(wallet, tx.txid);
```
cmgustavo, line 890: Seems that you don't need `coins`

2018jbsober: I do I don’t no how to put them in a wallet

PR #10880, opened 2020-05-07
[FIX]: refresh total cash value on incoming tx

rastajpa:


src/pages/tabs/tabs.ts
```diff
@@ -104,12 +104,13 @@ export class TabsPage {
...
         'NewOutgoingTx',
-        'UpdateTx'
+        'UpdateTx',
+        'NewIncomingTx'
       ],
       (eventName: string) => {
         if (walletId && type == eventName) {
           setTimeout(() => {
-            this.updateTxps();
+            this.fetchAllWalletsStatus();
```
cmgustavo, line 113: I wouldn't like to fetch all wallets when happen any of those events (performance issue). If it's only for new incoming tx I'd rather to a new function

PR #10886, opened 2020-05-11
Fix: default values when comes from charts

Gamboster:
Bug fixed: If you go to Simplex through charts and then change the selected currency, default values are not updated


Gamboster:
> LGTM. But please, add a comment there.

Done

PR #10892, opened 2020-05-14
[FIX]: get wallet balances after importing a key

rastajpa:


src/pages/add/import-wallet/import-wallet.ts
```diff
@@ -242,6 +242,7 @@ export class ImportWalletPage {
       this.walletProvider.updateRemotePreferences(wallet);
       this.pushNotificationsProvider.updateSubscription(wallet);
       this.profileProvider.setWalletBackup(wallet.credentials.walletId);
+      this.walletProvider.fetchStatus(wallet, {});
```
cmgustavo, line 245: Maybe we could call an existent event that updates all wallets and total balance: `Local/FetchWallets`

PR #10910, opened 2020-05-19
ref(wallets-tab): cache ordered wallets by group to speed up the app

cmgustavo:


src/pages/tabs/tabs.ts
```diff
@@ -236,7 +239,7 @@ export class TabsPage {
 
   private async fetchAllWalletsStatus() {
     this.logger.debug('Fetching All Wallets and calculate Total Amount');
-    const wallets = this.profileProvider.getWallets();
+    const wallets = this.profileProvider.wallet;
```
gabrielbazan7, line 242: Not sure. But that way I think you are getting the hidden wallets too.

src/pages/tabs/tabs.ts
```diff
@@ -280,7 +283,7 @@ export class TabsPage {
 
     const promises = [];
 
-    _.each(this.profileProvider.wallet, wallet => {
+    _.each(wallets, wallet => {
```
gabrielbazan7, line 286: +1

src/pages/tabs/tabs.ts
cmgustavo: Ok, I'll add a filter. Thanks!

PR #10912, opened 2020-05-20
[FIX] Navigation stack and dark mode fixes for shared wallets

rastajpa:


cmgustavo:
Added here: https://github.com/bitpay/copay/pull/10910/commits/36ab842e3794dbd834694546cc5465501ad22d51 

Thanks!

PR #10917, opened 2020-05-21
[FIX] Show finish modal after creating a txp from read only wallets

rastajpa:


src/pages/send/confirm/confirm.ts
```diff
@@ -1181,7 +1181,7 @@ export class ConfirmPage {
     onlyPublish?: boolean,
     redirectionParam?: { redir: string }
```
cmgustavo, line 1182: What about to add default value here? 
```
redirectionParam?: { redir: string = '' }
```
Does it work?

cmgustavo: Forget this. I see the error.

PR #10920, opened 2020-05-22
Feat: BUSD added to Simplex integration

Gamboster:


src/providers/rate/rate.ts
```diff
@@ -61,7 +61,25 @@ export class RateProvider {
...
+      const newRate = +(
+        (this.rates[chain]['USD'] * this.rates['btc'][code]) /
+        this.rates['btc']['USD']
+      ).toFixed(2);
+      return newRate;
+    }
+    this.logger.warn(
+      'There are no rates for chain: ' + chain + ' - code: ' + code
+    );
+    return undefined;
```
cmgustavo, line 82: Why returning undefined is ok since the function should return a number?

src/providers/rate/rate.ts
```diff
@@ -61,7 +61,25 @@ export class RateProvider {
   public getRate(code: string, chain?: string, opts?: { rates? }): number {
     const customRate =
       opts && opts.rates && opts.rates[chain] && opts.rates[chain][code];
-    return customRate || this.rates[chain][code];
+    if (customRate) return customRate;
+    if (this.rates[chain][code]) return this.rates[chain][code];
+    if (
```
cmgustavo, line 66: IMO the logic is a bit complicated. I'd rather to hide rate if it isn't returned from the source.

src/providers/rate/rate.ts
Gamboster: undefined is a universal value and can be returned by functions of any type. We cannot return a number that could not be defined.

src/providers/rate/rate.ts
Gamboster: I think it isn't so complicated: it simply calculates the rate based on the BTC values. The `"anyOtherFiat"/USD` rate should be the same for all crypto we support.
With this fix, some errors are fixed in specific cases such as the calculation of maximums, minimums and default values to buy tokens with alternative fiat currencies that don't come with a rate from our servers.

src/providers/rate/rate.ts
cmgustavo: +1 thanks for the explanation!

src/providers/rate/rate.ts
cmgustavo: Since it's a Simplex issue only, why not to add that logic into the simplex view? `rate.js` is a generic provider that is wide used in the whole app. That's my concern, that we're adding something so specific for only one case, aren't we?

cmgustavo: Should add a comment there. I'll merge it for now

PR #10928, opened 2020-05-26
[FEAT]: handling of advertising push notifications

rastajpa:


src/providers/push-notifications/push-notifications.ts
```diff
@@ -224,4 +232,28 @@ export class PushNotificationsProvider {
...
     this.FCMPlugin.clearAllNotifications();
   }
+
+  private verifySignature(data): boolean {
+    const b = BWC.Bitcore;
+    const ECDSA = b.crypto.ECDSA;
+    const Hash = b.crypto.Hash;
+    const SEP = '::';
+    const pubKey =
+      '02463827f64d47a9adaeddbeb860c1ea83eeb19024f9056de1772604c8085340a5';
```
matiu, line 242: please move this to https://github.com/bitpay/copay/blob/master/app-template/bitpay/appConfig.json
marketingPublicKey

src/providers/push-notifications/push-notifications.ts
```diff
@@ -12,6 +12,7 @@ import { ConfigProvider } from '../config/config';
 import { PlatformProvider } from '../platform/platform';
 import { ProfileProvider } from '../profile/profile';
 
+import BWC from 'bitcore-wallet-client';
```
cmgustavo, line 15: Why don't you use provider `BwcProvider` instead to includes all bwc library?

src/pages/home/home.ts
```diff
@@ -179,6 +181,12 @@ export class HomePage {
     this.events.subscribe('Local/FetchCards', bpCards => {
       if (!bpCards) this.addBitPayCard();
     });
+    this.events.subscribe('ShowAdvertising', data => {
```
cmgustavo, line 184: Ads will be shown only if selected view is home. What happen if users are in another view? Won't they ever seen the ads?

matiu: this ads will only show if the user clicks on the push notification. That is the intended behaivor.

cmgustavo: +1

PR #10989, opened 2020-07-01
[REF] moved -> sent to self - remove this.wallet from send wallet list

gabrielbazan7:


src/pages/tx-details/tx-details.ts
```diff
@@ -237,7 +237,7 @@ export class TxDetailsModal {
           if (this.btx.action == 'received')
             this.title = this.translate.instant('Received Funds');
           if (this.btx.action == 'moved')
-            this.title = this.translate.instant('Moved Funds');
+            this.title = this.translate.instant('Sent to self Funds');
```
cmgustavo, line 240: Maybe it should be: `Sent funds to self`

PR #11064, opened 2020-08-04
[FEAT]: Coinbase accounts as payment option for invoices

rastajpa:
DONE:
- Coinbase accounts added to wallet selector. 
- Gift card purchases through Coinbase Accounts.
- Invoice payments with Coinbase Accounts.
- BitPay card loads through Coinbase Accounts.

### DIFF
--- a/src/pages/send/confirm/confirm.ts
+++ b/src/pages/send/confirm/confirm.ts
@@ -94,6 +94,7 @@ export class ConfirmPage {
   // Coinbase
   public fromCoinbase;
   public coinbaseAccount;
+  public coinbaseAccounts;
   public showCoinbase;
 
   public mainTitle: string;
@@ -325,15 +326,6 @@ export class ConfirmPage {
         this.currencyProvider.getPrecision(this.coin).unitToSatoshi,
         '1.2-6'
       );
-    } else if (wallet && wallet.accounts) {
-      // for payments with Coinbase accounts
-      this.tx.minerFee = this.navParams.data.minerFee;
-      this.totalAmount = tx.amount - this.tx.minerFee;
-      this.totalAmountStr = this.decimalPipe.transform(
-        this.totalAmount /
-        this.currencyProvider.getPrecision(this.coin).unitToSatoshi,
-        '1.2-6'
-      );
     }
   }
 
@@ -375,7 +367,11 @@ export class ConfirmPage {
       coin
     });
 
-    if (_.isEmpty(this.wallets) && !this.showCoinbase) {
+    this.coinbaseAccounts = this.showCoinbase
+      ? this.coinbaseProvider.getAvailableAccounts(coin)
+      : [];
+
+    if (_.isEmpty(this.wallets) && _.isEmpty(this.coinbaseAccounts)) {
       const msg = this.translate.instant(
         'You are trying to send more funds than you have available. Make sure you do not have funds locked by pending transaction proposals.'
       );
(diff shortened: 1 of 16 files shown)


src/providers/coinbase/coinbase.ts
```diff
@@ -761,4 +767,53 @@ export class CoinbaseProvider {
       });
     });
   }
+
+  public payInvoice(invoiceId, currency, twoFactorCode?) {
+    return new Promise((resolve, reject) => {
+      const url = 'https://bitpay.com/oauth/coinbase/pay/' + invoiceId;
+      const data = {
+        currency,
+        token: this.accessToken,
```
cmgustavo, line 776: Before send `token` you should check that it isn't expired.

src/providers/coinbase/coinbase.ts
```diff
@@ -761,4 +767,53 @@ export class CoinbaseProvider {
...
+      });
+
+      this.logger.debug('Coinbase: Paying invoice...');
+      this.http.post(url, data, { headers }).subscribe(
+        data => {
+          this.logger.info('Coinbase: Pay Invoce SUCCESS ' + data);
+          return resolve();
+        },
+        data => {
+          if (this.isExpiredTokenError(data.error.errors)) {
```
cmgustavo, line 791: It's not necessary here

src/providers/coinbase/coinbase.ts
```diff
@@ -761,4 +767,53 @@ export class CoinbaseProvider {
...
+            data.error.errors &&
+            data.error.errors[0].id == 'two_factor_required'
+          ) {
+            this.logger.error('Coinbase: 2FA is required ' + data.status);
+            return reject('2fa'); // return string to identify
+          } else {
+            this.logger.error(
+              'Coinbase: Send Transaction ERROR ' + data.status
+            );
+            return reject(this.parseErrorsAsString(data.error));
```
cmgustavo, line 813: Error handler on BitPay side is different than Coinbase API

src/pages/send/confirm/confirm.ts
```diff
@@ -318,6 +325,15 @@ export class ConfirmPage {
           this.currencyProvider.getPrecision(this.coin).unitToSatoshi,
         '1.2-6'
       );
+    } else if (wallet && wallet.accounts) {
+      // for payments with Coinbase accounts
+      this.tx.minerFee = this.navParams.data.minerFee;
```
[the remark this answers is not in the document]
rastajpa: It is necessary to avoid inconsistency between the amount of an invoice and what the user will pay. That value is used to show a message.

src/pages/wallets/wallets.ts
```diff
@@ -120,7 +123,11 @@ export class WalletsPage {
     this.coinbaseLinked = this.coinbaseProvider.isLinked();
     if (this.coinbaseLinked) {
       if (force || !this.coinbaseData) {
-        this.coinbaseProvider.updateExchangeRates();
+        const config = this.configProvider.get();
+        this.coinbaseAltCurrency = force
+          ? config.wallet.settings.alternativeIsoCode
```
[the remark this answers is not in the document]
rastajpa: Coinbase provides exchange rate for multiple currencies. But at the moment it is not necessary to apply this change. It was already removed

src/providers/coinbase/coinbase.ts
rastajpa: the only way to know if the token expired is with the error. So as a solution now the token is refreshed before trying to make the payment

src/providers/coinbase/coinbase.ts
rastajpa: in fact it's the same :)

src/providers/coinbase/coinbase.ts
```diff
@@ -761,4 +763,80 @@ export class CoinbaseProvider {
...
+    });
+    return coinbaseAccounts;
+  }
+
+  public payInvoice(
+    invoiceId: string,
+    currency: string,
+    twoFactorCode?: string
+  ): Promise<any> {
+    return this.doRefreshToken().then(_ => {
```
cmgustavo, line 798: If 2fa is required, this function will be called twice. I think that we could avoid that when 2fa was not null.

cmgustavo:
Needs rebase

cmgustavo:
Try to identify the source of funds (it's a wallet BCH from my Coinbase account)

Screen Shot 2020-08-20 at 09 38 55

cmgustavo:
I'm not sure what that error means. It happens after click on confirm button.

Screen Shot 2020-08-20 at 09 42 00

rastajpa:
> I'm not sure what that error means. It happens after click on confirm button.
> 
> Screen Shot 2020-08-20 at 09 42 00

this is because CORS...

PR #11093, opened 2020-08-21
[FEAT] ERC20 Multisig

gabrielbazan7:
To try this PR on testnet. Add to `TokenOpts` at `token.ts` file: 
```
  '0x44d69d16C711BF966E3d00A46f96e02D16BDdf1f': {
    name: 'USD Coin Testnet',
    symbol: 'USDC',
    decimal: 6,
    address: '0x44d69d16C711BF966E3d00A46f96e02D16BDdf1f'
}
``` 
remove line 211 and 237 from `selectCurrency.ts`

When adding the token select USDC testnet and select the multisig testnet wallet.

### DIFF
--- a/src/pages/add/create-wallet/create-wallet.ts
+++ b/src/pages/add/create-wallet/create-wallet.ts
@@ -430,9 +430,7 @@ export class CreateWalletPage implements OnInit {
       ? this.profileProvider.getWalletsFromGroup({
           keyId: this.keyId,
           hasFunds: true,
-          coin: 'eth',
-          m: 1,
-          n: 1
+          coin: 'eth'
         })
       : [];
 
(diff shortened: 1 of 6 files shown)


rowlands69 (approved):

Gamboster:
Great job. I was testing the PR and these are some things I saw:

I think that the creation of the ERC20 multisig wallets is not very clear through "Add Simple Wallet". It seems better to me to move that option to "Add Shared Wallet".

The history presented a problem in the amount of txs. They are displayed with too many extra digits than it should.

Some wallets were shown with "Error 500", but apparently it happens when trying to create a testnet wallet linked to a livenet or vice versa.

The import doesn't bring the multisig tokens.

gabrielbazan7:
Thanks for testing.
I moved the multisig tokens creation to the shared option. 
The problems you mention are probably due to internal transactions not being tracked on bitcore-node ( it works fine on my own bitcore node testnet instance ). 
Regarding the import of multisig wallets I couldn't find a way to make it work.

PR #11119, opened 2020-09-10
Update scan icon

vanhue:
- Scan icon should appear more clear with better colors


cmgustavo (changes requested):
LGTM, but we shouldn't remove package-lock. Please revert only this file and update PR. 
Thanks!

vanhue:
> LGTM, but we shouldn't remove package-lock. Please revert only this file and update PR.
> Thanks!

Whoops. How did that get in there? Just fixed

PR #11131, opened 2020-09-16
Fix: Wyre limits and fees

Gamboster:
(Limits and fees were updated in Wyre's docs.)
Also prevents cards from expanding if there is no price loaded


cmgustavo (approved):
LGTM

PR #11137, opened 2020-09-17
[FIX] IAB_INIT_CLEAR_LISTENER

JohnathanWhite:
removing event listener after IAB init to ensure one time execution


cmgustavo (approved):
LGTM

PR #11146, opened 2020-09-19
fix(connection): smooth app loading if server is down

cmgustavo:
* Pre-load total balance from localStorage (Home)
* Display "Network error" in Home
* Show cached balance for all wallets


gabrielbazan7 (approved):
LGTM

PR #11163, opened 2020-09-30
Feat: SEPA bank transfer payment method added

Gamboster:
This new payment method is accepted by Simplex, only for countries belonging to the EU.


rastajpa (approved):
Nice fee with this new payment method. Great work here!
LGTM

gabrielbazan7 (approved):

PR #11173, opened 2020-10-05
[REF] hide opts logs

gabrielbazan7:


src/providers/profile/profile.ts
```diff
@@ -1453,6 +1453,7 @@ export class ProfileProvider {
       if (showOpts.extendedPrivateKey) showOpts.extendedPrivateKey = '[hidden]';
       if (showOpts.mnemonic) showOpts.mnemonic = '[hidden]';
       if (showOpts.password) showOpts.password = '[hidden]';
+      if (showOpts.key) showOpts.key = '[hidden]';
```
cmgustavo, line 1456: Isn't it `keyId` instead `key`?

cmgustavo (approved):
ACK

PR #11174, opened 2020-10-05
fix(topup): show Coinbase accounts after enter amount in cryptocurrency

cmgustavo:


rastajpa:
BitPay
I can't see the alt currency amount in the wallet selector

cmgustavo:
> BitPay
> I can't see the alt currency amount in the wallet selector

Good catch. I'll take a look, thanks!

cmgustavo:
Weird, after testing a few times, I always see the fiat value.
Maybe it's a connection issue. I'll add a conditional to avoid that

Thanks!

rastajpa (approved):
ACK!

gabrielbazan7 (approved):

PR #11187, opened 2020-10-15
Bug/fix duplicate bch

matiu:


cmgustavo (approved):
ACK!

troowala:
Hello Does this bug fix address the issue of spendability of BCH in v11.0.4 where upon trying to spend funds from a BTC duplicated BCH wallet one gets the error "Your wallet is in a corrupt state, please contact support and share the logs provided". 

Also the wallet was imported from a phrase and not a file so there is definitely no file level corruption.

I have tried sending with different amounts all less than the max balance as well as the max amount but the issue continues to happen.

The logs dump the private key in them which i was suprised about hence not pasting them here but the BCH wallet upon duplication shows under (legacy) and when trying to send any funds i get the above error. I have tested on both Android & Ubuntu same result.

Thanks

PR #11193, opened 2020-10-20
[FEAT] Encrypt storage keys

gabrielbazan7:
To test add some encrypting keys to STORAGE_ENCRYPTING_KEYS array
```
STORAGE_ENCRYPTING_KEYS: [
        'asdfghjklpoiuytrewqazxcvbnjskawq',
        'poiqwerlkhjkasdfgiuwerhjabsdfgks',
        'agksdfkjg234587asdjkhfdsakhjg283'
      ]

### CODE REVIEW

File: src/constants.ts [hunk 1]
```diff
@@ -1,2 +1,9 @@
 export const CARD_IAB_CONFIG =
   'directories=no,titlebar=no,toolbar=no,location=no,status=no,menubar=no,scrollbars=no,resizable=no,hidden=yes,clearcache=yes,hidespinner=yes,disallowoverscroll=yes,zoom=no,transitionstyle=crossdissolve';
+
+// new ky first
+export const ENC_KYS = [
```
matiu, line 5: for security, the keys should be #private var
src/providers/key-encrypt/key-encrypt.ts 
 and should not be available outside that class.

src/providers/key-encrypt/key-encrypt.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  init() {
+    return new Promise<void>(resolve => {
+      setTimeout(async () => {
+        const storage = this.platform.is('cordova')
+          ? new FileStorage(this.file, this.logger)
+          : new LocalStorage(this.logger);
+
+        const keys = await storage.get('keys'); // get key
```
matiu, line 27: keys dont need to be stored.

src/providers/key-encrypt/key-encrypt.spec.ts
```diff
@@ -0,0 +1,205 @@
+// ADD SETTER AND GETTER TO key-encrypt.ts TO RUN THIS TESTS
```
matiu, line 1: that is this file? why all commented out?

PR #11198, opened 2020-10-22
Remove some integrations

msalcala11:


cmgustavo:
I think that we need to disable it from here: https://github.com/bitpay/copay/blob/master/src/app/app.component.ts#L488 

If someone already has linked their account, it will continue connecting to the external service.

cmgustavo (approved):
👍

PR #11203, opened 2020-10-26
ref(webpack): no obfuscate function for dev env

cmgustavo:
Only for env.DEV on browsers:

* Refresh data on browser after code has changed


bsnowden3 (approved):
ACK!

gabrielbazan7 (approved):

PR #11205, opened 2020-10-27
fix(lock app): reset timer if it failed on resume the event

cmgustavo:
Error (negative coundown): 

Screen Shot 2020-10-26 at 16 23 38


gabrielbazan7 (approved):
ACK

PR #11212, opened 2020-10-29
[FIX] erc20 & eth multisig paypro when linked wallet has no funds

gabrielbazan7:
needs https://github.com/bitpay/bitcore/pull/3000


cmgustavo (approved):
ACK!!

cmgustavo (approved):
ACK!

PR #11231, opened 2020-11-03
[FEAT] added fb analytics bridge to wallet events

JohnathanWhite:


src/app/app.component.ts
```diff
@@ -382,6 +384,7 @@ export class CopayApp {
           });
       });
     }
+    this.analyticsProvider.logEvent('adsupport_test_event', {});
```
cmgustavo, line 387: I would add it into that conditional: `this.platformProvider.isCordova && this.appProvider.info.name === 'bitpay'`

cmgustavo (approved):
ACK

PR #11236, opened 2020-11-04
fix(apple-wallet): use custom plugin to force encryption scheme

cmgustavo:
See https://github.com/cmgustavo/cordova-apple-wallet


JohnathanWhite (approved):

gabrielbazan7 (approved):
ACK

PR #11248, opened 2020-11-09
ref(coinbase): increase limit for accounts

cmgustavo:


src/providers/coinbase/coinbase.ts
```diff
@@ -411,7 +413,7 @@ export class CoinbaseProvider {
   }
 
   private _getAccounts(): Promise<any> {
-    const url = this.credentials.API + '/v2' + '/accounts';
+    const url = this.credentials.API + '/v2' + '/accounts?&limit=' + LIMIT;
```
Gamboster, line 416: Is the `&` here necessary?

rastajpa (approved):
ACK

Gamboster (approved):
ACK

PR #11256, opened 2020-11-11
[FEAT] bitpay Id unlock shopper

JohnathanWhite:
This PR allows for unlocking/claiming of invoices with the bitpay Id userShopper token. To test use the following format for the intent:

`bitpay://unlock?r=https://<host>/i/<invoiceId>`


cmgustavo (approved):
ACK

PR #11269, opened 2020-11-17
fix(ui): improves dark/light mode on some user actions

cmgustavo:
### A white background is showing when opening keyboard

RPReplay_Final1605647985

### Fixed: dark/light background

RPReplay_Final1605647942


gabrielbazan7 (approved):
ACK

PR #11270, opened 2020-11-17
[FEAT]: handling data messages

rastajpa:
Needs https://github.com/bitpay/bitcore/pull/3012


tuanggo (approved):

src/providers/push-notifications/push-notifications.ts
```diff
@@ -70,10 +70,27 @@ export class PushNotificationsProvider {
           config.productsUpdates.enabled
         )
           this.subscribeToTopic('productsupdates');
+
+        setInterval(() => {
```
cmgustavo, line 74: If users decide to turn off notifications, we should kill setInterval when unsubscribe.

cmgustavo (approved):
ACK!

PR #11273, opened 2020-11-19
[FIX]: Update backup flag when create new keys

rastajpa:


Gamboster:
There are some tests failing

cmgustavo (approved):
ACK

src/providers/key/key.ts
```diff
@@ -244,7 +244,8 @@ export class KeyProvider {
 
   public isDeletedSeed(keyId: string): boolean {
     if (!keyId) return true;
-    const key = this.getKey(keyId);
+    let key = this.getKey(keyId);
```
matiu, line 247:
``` 
let key = this.getKey(keyId);
if (!key) return true;

let keyObj = key.toObj();
return ( !keyObj.mnemonic && !keyObj.mnemonicEncrypted ) ;
}

File: src/providers/profile/profile.ts [hunk 2]
```diff
@@ -207,13 +207,15 @@ export class ProfileProvider {
   private requiresGroupBackup(keyId: string) {
     let k = this.keyProvider.getKey(keyId);
     if (!k) return false;
+    else k = k.get();
```
matiu, line 210: replace:
```else k = k.get();```

with:
```
const keyObj = k.toObj();
if (!keyObj.mnemonic && !keyObj.mnemonicEncrypted) return false;
return true;
```

File: src/providers/profile/profile.ts [hunk 3]
```diff
@@ -207,13 +207,15 @@ export class ProfileProvider {
...
     if (!k.mnemonic && !k.mnemonicEncrypted) return false;
     return true;
   }
 
   private requiresBackup(wallet) {
     let k = this.keyProvider.getKey(wallet.credentials.keyId);
     if (!k) return false;
+    else k = k.get();
```
matiu, line 218: see above.

PR #11285, opened 2020-11-23
Revert "[FIX]: Update backup flag when create new keys"

cmgustavo:
Reverts bitpay/copay#11273

It breaks the app: 
error-profile


Gamboster (approved):
ACK

PR #11286, opened 2020-11-23
Disable SEPA bank transfer payment method

Gamboster:
This feature will be disabled until Simplex ensures a stable version from its website


cmgustavo (approved):
ACK

PR #11288, opened 2020-11-23
[FIX]: Fixes for Add Funds page

rastajpa:


src/providers/persistence/persistence.ts
```diff
@@ -883,6 +883,14 @@ export class PersistenceProvider {
   getCardFastTrackEnabled() {
     return this.storage.get(Keys.CARD_FAST_TRACK_ENABLED);
   }
+
+  setOnboardingFlowFlag(value: string) {
```
cmgustavo, line 887: Why it's needed another Onboarding flag?

rastajpa: The other flag is to show the keyOnboardingSlides that are different from the "feature education" of the new onboarding. Those slides can appear in case the user deletes all their keys and has already completed onboarding.

cmgustavo (approved):
LGTM

PR #11298, opened 2020-11-25
fix(dynamic-links): adds handler to redirect any valid view

cmgustavo:
Still having issue if app is closed. It works on background: https://github.com/chemerisuk/cordova-plugin-firebase-dynamiclinks/issues/85


Gamboster (approved):
LGTM

PR #11302, opened 2020-11-26
fix(android): splash screen doesn't work if App is asking camera permission

cmgustavo:
We should remove this patch since it could affect loading the app on first installations. 
It doesn't show the splash screen.


Gamboster (approved):
ACK

PR #11312, opened 2020-11-30
fix(bp-card-intro): ask camera permission when open bp card intro view

cmgustavo:
Screen Shot 2020-11-30 at 16 42 45


Gamboster (approved):
ACK

PR #11328, opened 2020-12-03
fix(bp-card-intro): release camera resources after initializing scanner

cnguyen812:
This PR pauses the scanner preview after initializing it from the card intro. The purpose of calling the scanner initialization is solely to invoke a camera permission request, so it should be safe to pause it immediately after. On Android, the selfie camera was not available in the InAppBrowser until the scanner is paused. 

Tested against Pixel 3 XL / Android 11.


cmgustavo (approved):
LGTM! Tested on iOS

PR #11348, opened 2020-12-10
fix(dynamic-links): open deepLink on Android if App is non-installed, closed or in background

cmgustavo:
Uses latest version of FCM plugin: https://github.com/cmgustavo/cordova-plugin-fcm/releases/tag/v8.2.0


Gamboster (approved):
ACK

PR #11350, opened 2020-12-11
Fix: mnemonicHasPassphrase

Gamboster:


src/providers/wallet/wallet.ts
```diff
@@ -1767,7 +1767,8 @@ export class WalletProvider {
       }
 
       const mnemonicHasPassphrase = this.keyProvider.mnemonicHasPassphrase(
-        wallet.credentials.keyId
+        wallet.credentials.keyId,
+        password
```
matiu, line 1771: should not be necessary: 
https://github.com/bitpay/bitcore/blob/master/packages/bitcore-wallet-client/src/lib/key.ts#L313

matiu:
Closed due to https://github.com/bitpay/copay/pull/11352

PR #11359, opened 2020-12-15
fix(performance): downgrade typescript version

cmgustavo:
* Removes unused plugin on `webpack.config`
* Removes private function from BWC
* Removes walletConnect
* Removes `electron-builder`
* Remove unused CSS classes
* Fix tslint issues (card and giftcard)
* Downgrades `@types/node` and `typescript`
* AOT true


gabrielbazan7 (approved):
ACK

PR #11390, opened 2020-12-23
Fix: creation of wallets

Gamboster:
- Can create new keys by joining a wallet
- Fixed error when trying to create an ETH wallet in some cases (It was shown disabled as if it were a multisig)
- Create wallets from "Buy crypto", now difference between singleseed and multiseeds
- Fix: strings with missing translations


gabrielbazan7 (approved):
ACK

PR #11419, opened 2021-01-04
Feat: add DAI and WBTC support to Wyre

Gamboster:
- Adds DAI and WBTC support to Wyre

I also tested USDC, DAI (crypto) and IDR, SAR (Fiat) that are listed as supported on the Simplex website. But they don't seem to work. That's why I didn't add them.

- Wyre's support url updated. At their request, I have removed their contact email.
- This PR also fixes an error that disable continue button in amount page when select an uncommon fiat currency
- Adds a log for certain errors in requests to Simplex


gabrielbazan7 (approved):
ACK

PR #11422, opened 2021-01-05
[REF] wallet connect handle url - move events to provider

gabrielbazan7:


src/providers/wallet-connect/wallet-connect.ts
```diff
@@ -37,6 +80,160 @@ export class WalletConnectProvider {
     });
   }
 
+  public async retrieveWalletConnector(walletConnectData) {
```
cmgustavo, line 83: Why you need async here?

cmgustavo (approved):
LGTM! 🚀

PR #11431, opened 2021-01-08
Ref: exchange and buy crypto improvements

Gamboster:
[DONE] I should not be allowed to enter more than what I own. Perhaps an error that says "You do not have enough...." This prevents having to see the same error much later in the flow.

[DONE] Need back button in top left exchange page

[DONE] More margin between title and body area (40 px)

[DONE] Can we have a toggle between fiat and crypto in the amount screen?
when you go into the choose amount view.. you switch to fiat to enter a dollar amount and when you come back to Exchange page, it would auto convert to crypto amount

[DONE] The icons on the wallet activity view for "send, receive, exchange, and buy" seem a little small. Houston also wants to make them like the crypto icons (rounded boxes). 50px x 50px with 12px border-radius

[DONE] Remove the $ icon for buy crypto

[DONE] When I tap on exchange from within a wallet, it should automatically select the wallet you in.
when you're in the wallet you want to exchange from, that wallet gets prefilled in "source"..
If the wallet has 0, that wallet is set for destination
if it doesn't support it, we don't show the exchange button
if a coin only supports send/receive, then we just show those two buttons

[DONE] Fixed: when i tap cancel it goes to wallets page instead of wallet detail page


cmgustavo:
Trying to send more funds that I have, I get this:

Screen Shot 2021-01-11 at 10 17 34

Then, select Max Amount:

Screen Shot 2021-01-11 at 10 17 43

Max Amount takes less than max available.

cmgustavo (approved):
ACK

PR #11438, opened 2021-01-13
Add v12 new features slides and fix check of major version

gonzaloleon:
Apply
- add v12 new features slides (with images)
- add dark mode styles
- new features slides only for _major version_
- implement **Version** interface


cmgustavo (approved):
ACK! 🚀

PR #11444, opened 2021-01-15
fix(wallets-tab): update balance for new wallets when receiving funds

cmgustavo:
Issues

* If wallet has been created recently, balance 0. After received funds, I had to restart the app to see new balance in the list of wallets.
* `hasZeroBalance` lock the value in zero, due to cached info is empty (for recently created wallet).


Gamboster (approved):
LGTM

PR #11447, opened 2021-01-18
Minor fix - sendPage

gonzaloleon:
- Move events subscribers to ionViewWillEnter
- Fix this.search assignment
- Move **_setDataFromClipboard_** to async method

### DIFF
--- a/src/pages/send/send.spec.ts
+++ b/src/pages/send/send.spec.ts
@@ -14,7 +14,7 @@ import { PlatformProvider } from '../../providers/platform/platform';
 // pages
 import { SendPage } from './send';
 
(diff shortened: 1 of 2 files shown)


src/pages/send/send.spec.ts
```diff
@@ -14,7 +14,7 @@ import { PlatformProvider } from '../../providers/platform/platform';
 // pages
 import { SendPage } from './send';
 
-describe('SendPage', () => {
+fdescribe('SendPage', () => {
```
Gamboster, line 17: fdescribe here

cmgustavo (approved):
LGTM

PR #11454, opened 2021-01-20
fix(coinbase): Ignore coin without exchange rate

cmgustavo:
* Speedup Wallets tab. Show Coinbase section until coinbaseData is available


Gamboster (approved):
ACK

gonzaloleon (approved):
LGTM!

PR #11458, opened 2021-01-22
Debounce redeem calls

msalcala11:
Currently NewBlock bws events trigger a redeem request, but sometimes a whole bunch of blocks come in at the same time, each triggering a redeem request. This PR ensures redeem requests are properly debounced in such a scenario.


cmgustavo (approved):
LGTM

PR #11459, opened 2021-01-25
fix(topup): adds new param to getSendMaxInfo to be more precise

cmgustavo:


matiu (changes requested):
obsolete, please close.

PR #11469, opened 2021-01-29
[REF]: check alt currency

rastajpa:
It needs https://github.com/bitpay/bitcore/pull/3060

### DIFF
--- a/src/pages/home/home.ts
+++ b/src/pages/home/home.ts
@@ -22,6 +22,7 @@ import {
   PlatformProvider,
   PopupProvider,
   ProfileProvider,
+  RateProvider,
   ReleaseProvider
 } from '../../providers';
 import { ActionSheetProvider } from '../../providers/action-sheet/action-sheet';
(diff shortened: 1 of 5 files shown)


src/pages/tabs/tabs.ts
```diff
@@ -121,6 +121,7 @@ export class TabsPage {
 
     await this.checkCardEnabled();
     await this.tabProvider.prefetchCards();
+    this.checkAltCurrency(); // Check if the alternative currency setted is no longer supported
```
cmgustavo, line 124: I'm not completely sure about this line. tabPage is loaded first, then homePage. So, if those `await` don't exist, the action won't be called due to the event from home was not subscribed.

We should avoid that async issue to mitigate any random behavior when app is loading.

src/pages/home/home.ts
```diff
@@ -874,16 +874,16 @@ export class HomePage {
     });
   }
 
-  private showInfoSheet(params): void {
+  private showInfoSheet(altCurrency): void {
```
cmgustavo, line 877: I'd rename the function's name to be more specifics for "altCurrency". It's a generic name.

cmgustavo (approved):
LGTM! 👍🏼

PR #11473, opened 2021-02-02
Fixes and improvements for buy crypto

Gamboster:
This PR needs: https://github.com/bitpay/bitcore/pull/3064/files
Anyway, it can be tested by sending the amount parameter with null

Fixes:
- Remove "Buy XRP" from charts page
- Wyre's redirectUrl has been fixed to return to the app after a purchase
- Crypto purchases not supported by Wyre will no longer set Apple pay by default. It will set debit card
- Now you can select the same payment method that was previously selected
- Error messages in buy crypto are now red
- The amount page button now says "Continue" (instead of "Select payment method")

Improvements:
- New analytics events to Buy / Exchange crypto with more data
- Refactored Wyre integration to work with the latest changes to its documentation.
- Consistency between the offer shown by Wyre and Simplex => Refactor in Wyre's getOffer function
- Improved the explanation about buy crypto fees

### DIFF
--- a/src/pages/buy-crypto/crypto-offers/crypto-offers.html
+++ b/src/pages/buy-crypto/crypto-offers/crypto-offers.html
@@ -36,7 +36,12 @@
         </div>
         <div class="content" translate *ngIf="offer.value.fiatMoney || offer.value.errorMsg">
           <div *ngIf="offer.value.errorMsg">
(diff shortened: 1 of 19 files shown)


gabrielbazan7 (approved):
ACK !

PR #11478, opened 2021-02-04
Ref: remove Wyre for EU countries and Canada if promotion is active

Gamboster:
This PR needs: https://github.com/bitpay/bitcore/pull/3070

Removes Wyre option from EU Countries and Canada.
Sets Credit card as default option


gabrielbazan7 (approved):
ACK

PR #11484, opened 2021-02-05
APP-61 [WIP] [Improve] SendPage - AddressBook

gonzaloleon:
### Improve

- Add async pipe to pages
- Switch to async functions
- other minor improves

### Fix
- A wallet or contact (to send) can be touch more than once


cmgustavo (changes requested):
* Send page seems to be fine, but I don't see any loading indicator (blank screen for a few seconds, then wallets appear).

* Address book seems to be slower than before. It takes a second to show your contacts (I only have 5)

PR #11494, opened 2021-02-12
[FIX] iab pause lock disabled

JohnathanWhite:
- removed unused
- prevents the in app browser from pausing when lockMethod is disabled


cmgustavo (approved):
LGTM

PR #11503, opened 2021-02-16
APP-66 ref(biometric): replaces old biometric cordova plugin by ionic native (2)

cmgustavo:
* Also fix Android that execute `pause` when system actions are called


gabrielbazan7 (approved):
ACK!! Nice Work !!

PR #11507, opened 2021-02-18
[FEAT] dynamic network setting

JohnathanWhite:
- tapping on 'Home' toggles between networks
- tied feature-data provider and tab component to globally set network
- fixed the card advertisement not being set on init

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -364,13 +364,6 @@ export class CopayApp {
       this.platformProvider.isCordova &&
       this.appProvider.info.name === 'bitpay'
     ) {
-      if (cards) {
-        this.events.publish('CardAdvertisementUpdate', {
-          status: 'connected',
-          cards
-        });
-      }
-
       const host =
         this.NETWORK === 'testnet' ? 'test.bitpay.com' : 'bitpay.com';
       this.logger.log(`IAB host -> ${host}`);
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -21,6 +21,7 @@ import {
   GiftCardProvider,
   IABCardProvider,
   InAppBrowserProvider,
+  NewFeatureData,
   PersistenceProvider
 } from '../providers';
 import { AppProvider } from '../providers/app/app';
@@ -151,7 +152,8 @@ export class CopayApp {
     private bitpayIdProvider: BitPayIdProvider,
     private themeProvider: ThemeProvider,
     private logsProvider: LogsProvider,
-    private dynamicLinksProvider: DynamicLinksProvider
+    private dynamicLinksProvider: DynamicLinksProvider,
+    private newFeatureDataProvider: NewFeatureData
   ) {
     this.imageLoaderConfig.setFileNameCachedWithExtension(true);
     this.imageLoaderConfig.useImageTag(true);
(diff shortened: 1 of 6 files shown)


src/pages/home/home.ts
```diff
@@ -832,20 +833,20 @@ export class HomePage {
...
-      });
+      this.persistenceProvider
+        .getNetwork()
+        .then((currentNetwork: Network | undefined) => {
+          const newNetwork =
+            !currentNetwork || currentNetwork === Network.livenet
+              ? Network.testnet
+              : Network.livenet;
+          this.persistenceProvider.setNetwork(newNetwork);
+          alert(`Network changed to ${newNetwork}. Please restart the app.`);
```
cmgustavo, line 847: I'd avoid to use `alert` function. Maybe you could restart the app automatically. Take a look: https://github.com/bitpay/wallet/blob/master/src/pages/settings/wallet-settings/wallet-settings-advanced/wallet-service-url/wallet-service-url.ts#L119

src/app/app.component.ts
```diff
@@ -151,7 +152,8 @@ export class CopayApp {
     private bitpayIdProvider: BitPayIdProvider,
     private themeProvider: ThemeProvider,
     private logsProvider: LogsProvider,
-    private dynamicLinksProvider: DynamicLinksProvider
+    private dynamicLinksProvider: DynamicLinksProvider,
+    private newFeatureDataProvider: NewFeatureData
```
cmgustavo, line 156: I'm not completely sure about this. NewFeatureDataProvider is not used in app.component

JohnathanWhite: I was using it in my setNetwork method 

ln 233
 ```
[
      this.bitpayProvider,
      this.bitpayIdProvider,
      this.iabCardProvider,
      this.giftCardProvider
    ].forEach(provider => provider.setNetwork(this.NETWORK));
```

 but its causing a dependency injection issue. I have removed that provider from the app component and just changed it to pull network from local storage.

File: src/pages/home/home.ts (hunk 1 above)
Comment (JohnathanWhite) [reply src/pages/home/home.ts:847]: Ok i added the automatic reload but changed the alert to use an info sheet. Just to avoid confusion for testers, I wanted to have an indication of what network they are using.

File: src/app/app.component.ts [hunk 3]
```diff
@@ -221,7 +221,14 @@ export class CopayApp {
...
     );
 
-    this.logger.debug('BitPay: setting network');
+    const network = await this.persistenceProvider.getNetwork();
+
+    if (network) {
+      this.NETWORK = network;
+    }
+
+    this.logger.debug('BitPay: setting network', this.NETWORK);
```
cmgustavo, line 230: +1

src/app/app.component.ts
```diff
@@ -357,6 +364,13 @@ export class CopayApp {
       this.platformProvider.isCordova &&
       this.appProvider.info.name === 'bitpay'
     ) {
+      if (cards) {
+        this.events.publish('CardAdvertisementUpdate', {
```
cmgustavo, line 368: Be careful sending an event here. This is the first component the app will load before tabs.ts and home.ts. So, if subscription to this event is in home.ts, it won't be called (I guess)

src/pages/home/home.ts
```diff
@@ -832,20 +835,32 @@ export class HomePage {
...
+            'in-app-notification',
+            {
+              title: 'Network Changed',
+              body: `Network changed to ${newNetwork}. Restarting app.`
+            }
+          );
+          infoSheet.present();
+          infoSheet.onDidDismiss(() => {
+            window.location.reload();
+            if (this.platformProvider.isCordova) this.splashScreen.show();
```
cmgustavo, line 859: Good idea.

cmgustavo (dismissed):
ACK!

cmgustavo (approved):
LGTM 👍🏼

PR #11508, opened 2021-02-19
ref(location): optimize http request (using cache)

cmgustavo:
* Removed unused css classes
* Use Cached country instead to call same request


src/providers/tab/tab.ts
```diff
@@ -9,26 +13,25 @@ export class TabProvider {
   activeGiftCardsPromise: Promise<GiftCard[]>;
```
JohnathanWhite, line 13: we can also remove the bitpayCardItemsPromise declaration here

cmgustavo: +1

JohnathanWhite:
ACK!

gabrielbazan7 (approved):
ACK

PR #11516, opened 2021-02-24
Fix perpetual spinner when buying a gift card in some cases

msalcala11:
A few users have reported a bug that results in a perpetual spinner being shown when they attempt to purchase a specific gift card brand in the BitPay app which makes it impossible for them to purchase any gift cards for that particular brand. This PR fixes that issue.


cmgustavo (approved):
LGTM

PR #11517, opened 2021-02-24
APP-120 Clean before recreate and clean after keydelete

gonzaloleon:


src/providers/wallet/wallet.ts
```diff
@@ -1424,6 +1424,7 @@ export class WalletProvider {
   public recreate(wallet): Promise<any> {
     return new Promise((resolve, reject) => {
       this.logger.info('Recreating wallet:', wallet.id);
+      this.clearTxHistory(wallet);
```
cmgustavo, line 1427: I'd prefer to remove it. Why do we need to clear txs history?

src/pages/settings/key-settings/key-delete/key-delete.ts
```diff
@@ -62,13 +64,13 @@ export class KeyDeletePage {
...
 
-        wallets.forEach(wallet => {
-          this.pushNotificationsProvider.unsubscribe(wallet);
-        });
-
         const keyInUse = this.profileProvider.isKeyInUse(this.keyId);
 
         if (!keyInUse) {
+          wallets.forEach(wallet => {
+            this.pushNotificationsProvider.unsubscribe(wallet);
```
cmgustavo, line 71: +1

cmgustavo (approved):
ACK!

PR #11521, opened 2021-02-25
APP-72 [Feat] Call clearCache endpoint

gonzaloleon:
Add call to clearCache endpoint from WalletTransactionHistory to clear cache from server

Need to deploy: [bitpay/bitcore#3066](https://github.com/bitpay/bitcore/pull/3066)

Have to wait to update of bitcore-wallet-client with the clearCache API function.

### DIFF
--- a/src/providers/wallet/wallet.ts
+++ b/src/providers/wallet/wallet.ts
@@ -731,7 +731,7 @@ export class WalletProvider {
     });
   }
 
-  public clearCache(wallet): Promise<any> {
+  clearWalletCache(wallet): Promise<boolean> {
     return new Promise(resolve => {
       const config = this.configProvider.get();
       const defaults = this.configProvider.getDefaults();
@@ -1471,7 +1471,7 @@ export class WalletProvider {
   public clearTxHistory(wallet): void {
     this.invalidateCache(wallet);
     this.persistenceProvider.removeTxHistory(wallet.id);
-    this.clearCache(wallet)
+    this.clearWalletCache(wallet)
       .then(() => {
         this.logger.info(
           `TxHistory cache cleared from server for: ${wallet.id}`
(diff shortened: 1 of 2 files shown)


src/providers/wallet/wallet.ts
```diff
@@ -731,6 +731,23 @@ export class WalletProvider {
     });
   }
 
+  public clearCache(wallet): Promise<any> {
```
cmgustavo, line 734: Is this function accesible from external?

src/providers/wallet/wallet.ts
```diff
@@ -731,6 +731,23 @@ export class WalletProvider {
...
+      const defaults = this.configProvider.getDefaults();
+      const bws_url =
+        (config.bwsFor && config.bwsFor[wallet.id]) || defaults.bws.url;
+      this.bwcProvider
+        .getClient(JSON.stringify(wallet.credentials), {
+          bwsurl: bws_url
+        })
+        .clearCache(err => {
+          if (err) resolve(false);
+          return resolve(true);
```
cmgustavo, line 746: This function returns a boolean, but in definition it says <any>

src/providers/wallet/wallet.ts
```diff
@@ -1454,6 +1471,15 @@ export class WalletProvider {
   public clearTxHistory(wallet): void {
     this.invalidateCache(wallet);
     this.persistenceProvider.removeTxHistory(wallet.id);
+    this.clearCache(wallet)
```
cmgustavo, line 1474: I'd suggest a more specific name, like `clearHistoryCache` or something. It only clears txs history cache, right?

cmgustavo (changes requested):
LGTM, just suggested a few changes

cmgustavo (approved):
ACK

PR #11552, opened 2021-03-08
[FEAT]: rocket animation for doge

rastajpa:
rocket


src/pages/finish/finish.ts
```diff
@@ -32,9 +113,23 @@ export class FinishModalPage {
         this.viewCtrl.dismiss();
       }, 4000);
     }
+    if (this.navParams.data.coin === 'doge' && this.cssClass === 'success') {
```
cmgustavo, line 116: I'd add a control to check if `coin` is undefined. What do you think?

cmgustavo (dismissed):
LGTM! Just add a control to check if `coin` param doesn't exist

src/pages/finish/finish.ts
rastajpa: yes, that make sense...

rastajpa:
> LGTM! Just add a control to check if `coin` param doesn't exist

Done, ty!

cmgustavo (approved):
ACK 🚀

PR #11553, opened 2021-03-08
APP-141 Total for all coins and tokens tx

gonzaloleon:
- Show total for all coins, not only UTXOs
- Show total for ERCTokens in fiat


cmgustavo (dismissed):
LGTM

PR #11554, opened 2021-03-09
APP-62 Adjust feeRate calculation and UI message

gonzaloleon:
- Change fee percentage calculation over the amount (not amount+fee)
- Remove possible confuse message on UI


cmgustavo (approved):
ACK

PR #11565, opened 2021-03-17
[FIX] sign typed data method - wallet connect

gabrielbazan7:


package.json
```diff
@@ -162,7 +162,9 @@
     "cordova-plugin-x-socialsharing": "5.6.8",
     "cordova-plugin-x-toast": "2.6.0",
     "countries-list": "2.5.5",
+    "secp256k1": "4.0.2",
```
cmgustavo, line 165: I don't see that package in the app. Is it required?

cmgustavo (approved):
LGTM. Worked 👍🏼

PR #11588, opened 2021-03-29
[APP-149] Fix WalletConnect UI. Adds to Home tab (optional)

cmgustavo:
* (optional) Show in Home tab

### DIFF
--- a/src/providers/config/config.ts
+++ b/src/providers/config/config.ts
@@ -222,7 +222,7 @@ export class ConfigProvider {
         buycrypto: true,
         exchangecrypto: true,
         giftcards: true,
-        walletConnect: true
+        walletConnect: false
       },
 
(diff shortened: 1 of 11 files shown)


src/providers/config/config.ts
```diff
@@ -222,7 +222,7 @@ export class ConfigProvider {
         buycrypto: true,
         exchangecrypto: true,
         giftcards: true,
-        walletConnect: true
+        walletConnect: false
```
Gamboster, line 225: Is it okay for walletConnect to be turned off by default?

cmgustavo: Yes. We don't know that it shows in Home by default.

gabrielbazan7 (approved):
ACK

PR #11603, opened 2021-04-12
Ref: format Amount

Gamboster:
This PR is to prevent functions that sometimes return numbers and sometimes return strings.
Improves types and unit tests


gabrielbazan7 (approved):
ACK

PR #11607, opened 2021-04-14
Bump v12.4.3

cmgustavo:
# Release notes - BitPay Wallet - Version 12.4.3

### Bug

APP-196 ETH big number

APP-163 Return fee needed with insufficient funds for fee error


gabrielbazan7 (approved):
ACK

PR #11610, opened 2021-04-14
Catch removeTx errors during gift card purchases

msalcala11:
Fix for the perpetual spinner issue that occurs when some users try to buy a gift card but enter their encrypt password incorrectly.

Also simplifies the `publishAndSign` method in wallet provider.


cmgustavo (approved):
LGTM! 👍🏼

PR #11617, opened 2021-04-19
Fix(Coinbase): pay invoices and refresh token if it has expired

cmgustavo:
* APP-193
* APP-207


rastajpa (approved):
LGTM! nice work

gabrielbazan7 (approved):
ACK

PR #11621, opened 2021-04-22
APP-198 Reuse handleError with insufficient funds message improved

gonzaloleon:
Add complete insufficient funds message when buy giftcard
- MIner fee
- Network fee

And show both in UI.


gabrielbazan7 (approved):
ACK

PR #11643, opened 2021-05-06
[FIX] hiding switcher if cardTop

JohnathanWhite:
Hiding the switcher during card tops to prevent non-USD currency rate issues.


cmgustavo (approved):
LGTM 👍🏼  Tested on iOS.

PR #11649, opened 2021-05-11
APP-231: fix update icon after login. Use same icon inside bitpay-id …

cmgustavo:
…page. Force caps on the initials.


gabrielbazan7 (approved):
ACK

PR #11650, opened 2021-05-11
APP-115 Replace Toast Notification with Custom

gonzaloleon:


cmgustavo (changes requested):
Looks great! Just a few comments:

* Notification should hide in ~5 seconds automatically
* Below corners should be rounded

Screen Shot 2021-06-18 at 11 02 34

cmgustavo:
👍🏼 

Screen Shot 2021-07-05 at 17 42 13

I think It only needs more padding left-right like iOS: 

IMG_2696

cmgustavo:
Use `safe-area-inset-top` to skip the notch. 

IMG_2698

cmgustavo (approved):
Great work! 💯  LGTM

PR #11652, opened 2021-05-11
[FIX] globalTab indexes for card and settings

JohnathanWhite:
incremented card and settings globalTab.select index to account for transact being added


cmgustavo (approved):
ACK! 👍🏼

PR #11654, opened 2021-05-12
fix full precision for export; electron

bsnowden3:
Fixes precision display for file exports


src/providers/wallet/wallet.ts
```diff
@@ -769,14 +769,17 @@ export class WalletProvider {
 
         if (cacheCoin == 'bits') {
           this.logger.debug('Fixing Tx Cache Unit to: ' + wallet.coin);
+          const isFullPrecision = this.platformProvider.isElectron ? true : false;
```
JohnathanWhite, line 772: could remove the ternary as this.platformProvider.isElectron is a boolean

PR #11663, opened 2021-05-18
Fix APP-232: bp icon login, scan button and close button

cmgustavo:
* Fix login icon alignment (from Home, Settings and BitPayId views)
* Do not apply background when pressing "Scan" button (Active)
* Adds margin bottom to the "Close" button

### DIFF
--- a/src/pages/home/home.ts
+++ b/src/pages/home/home.ts
@@ -160,6 +160,12 @@ export class HomePage {
       CoinbasePage
     };
     this.user$ = this.iabCardProvider.user$;
+    this.user$.subscribe(async user => {
+      if (user) {
+        this.bitPayIdUserInfo = user;
+        this.accountInitials = this.getBitPayIdInitials(user);
+      }
+    });
   }
 
   private showNewFeatureSlides() {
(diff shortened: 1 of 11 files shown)


src/pages/home/home.ts
```diff
@@ -239,6 +243,7 @@ export class HomePage {
     this.user$.subscribe(async user => {
```
JohnathanWhite, line 243: Since this subscription is happening inside of the ionViewWillEnter it's causing a flash everytime you enter settings. Simple fix is to move the subscription to the constructor.

JohnathanWhite (changes requested):

src/pages/home/home.ts
JohnathanWhite: to be clear its just this block `this.user$.subscribe(async user => {
      if (user) {
        this.bitPayIdUserInfo = user;
        this.accountInitials = this.getBitPayIdInitials(user);
        this.changeRef.detectChanges();
      }
    });`

cmgustavo: Fixed! Thanks!

Gamboster (approved):
ACK

PR #11672, opened 2021-05-21
[FIX] settings id init

JohnathanWhite:


src/pages/settings/settings.ts
```diff
@@ -129,42 +128,37 @@ export class SettingsPage {
     this.isCordova = this.platformProvider.isCordova;
     this.isCopay = this.app.info.name === 'copay';
     this.user$ = this.iabCardProvider.user$;
```
cmgustavo, line 130: Getting compile error in this point: 
```
 typescript: src/pages/settings/settings.ts, line: 93 
            'user$' is declared but its value is never read.
```

JohnathanWhite: removed it thanks!

cmgustavo (approved):
ACK!

PR #11682, opened 2021-05-25
APP-253: Change name of walletConnect on local storage

cmgustavo:


src/providers/config/config.ts
```diff
@@ -222,7 +223,8 @@ export class ConfigProvider {
         buycrypto: true,
         exchangecrypto: true,
         giftcards: true,
-        walletConnect: false
+        walletConnect: false,
```
gonzaloleon, line 226: Then we could try to remove old/unused property

gonzaloleon (approved):

src/providers/config/config.ts
cmgustavo: It's necessary for backward compatibility. It has a scheme (data type) that could fail if it read a different object.

gabrielbazan7 (approved):

PR #11687, opened 2021-05-27
ref/Fix undefined address

gonzaloleon:


src/providers/address-book/address-book.ts
```diff
@@ -83,8 +83,10 @@ export class AddressBookProvider {
             if (ab && _.isString(ab)) ab = JSON.parse(ab);
             ab = ab || {};
             _.each(ab, contact => {
-              const ctc = this.getContact(contact);
-              if (ctc) contacts.push(ctc);
+              if (contact.address) {
```
cmgustavo, line 86: What happen if a contact doesn't have address? I think it's mandatory when creating new contact.

src/providers/address/address.ts
```diff
@@ -40,6 +40,7 @@ export class AddressProvider {
     str: string,
     network: string = 'livenet'
   ): CoinNetwork {
+    if (!str) return null;
```
cmgustavo, line 43: That function should never been called without 'str'. We should check the source of the problem.

cmgustavo (changes requested):
Just left a few comments.

src/providers/address-book/address-book.ts
cmgustavo: +1 (it's only for migration to new address book)

cmgustavo (commented):
Suggestions:

* Change the **key name** to something like: `coin-address` , instead `address (coin)` in order to avoid spaces and brackets.
* Normalize name of the key on the storage: `addressbook-v2-livenet`
* Run migration when the app start (app.component) only if new key in the storage doesn't exist. This should speedup the app when opening AddressBook or WalletDetails views.

src/app/app.component.ts
```diff
@@ -442,6 +444,8 @@ export class CopayApp {
         }
       });
     }
+
+    await this.addressBookProvider.migrateOldContacts();
```
cmgustavo, line 448: We don't need to wait for this. It's safe to remove await in order to do the process asynchronous.

src/providers/address/address.ts
cmgustavo: I think it is not necessary anymore since we have a conditional here: https://github.com/bitpay/wallet/pull/11687/files#diff-e20f5602886c255ee30a38778ad8d5c2a889c2dfe2fedf1126d732efd59e9f82R86

cmgustavo (approved):
ACK! Nice work 👍🏼

PR #11699, opened 2021-06-07
APP-251: Fix connect to Coinbase (Android)

cmgustavo:
* Connect to Coinbase with Android requires a redirect function to work. It only takes `code` and `state` params and open the BitPay App.
* Added STATE random param to improve security.


rastajpa (dismissed):
ACK! great work! tested on Android 10 device

rastajpa (approved):
ACK! good solution!

gabrielbazan7 (approved):

PR #11734, opened 2021-07-02
[FEAT] Litecoin Support

gabrielbazan7:


src/providers/incoming-data/incoming-data.ts
```diff
@@ -482,6 +494,21 @@ export class IncomingDataProvider {
     } else this.goSend(address, amount, message, coin);
   }
 
+  private handleLitecoinUri(data: string, redirParams?: RedirParams): void {
+    this.logger.debug('Incoming-data: Litecoin URI');
+    let amountFromRedirParams =
+      redirParams && redirParams.amount ? redirParams.amount : '';
+    const coin = Coin.LTC;
+    let parsed = this.bwcProvider.getBitcoreDoge().URI(data);
```
Gamboster, line 502: we should use `getBitcoreLtc()` here

Gamboster:
Should we add a ninth URL scheme to config-template.xml or is it something to be done in a separate PR?
`<variable name="NINTH_URL_SCHEME" value="litecoin" />`

gabrielbazan7:
> Should we add a ninth URL scheme to config-template.xml or is it something to be done in a separate PR?
> `<variable name="NINTH_URL_SCHEME" value="litecoin" />`

will do it now

cmgustavo:
> > Should we add a ninth URL scheme to config-template.xml or is it something to be done in a separate PR?
> > `<variable name="NINTH_URL_SCHEME" value="litecoin" />`
> 
> will do it now

Also it's needed for Electron: https://github.com/bitpay/wallet/blob/master/electron/main.js#L140

package.json
```diff
@@ -130,7 +130,7 @@
     "apple-wallet-ng": "1.1.1",
     "base64-js": "1.3.0",
     "bitauth": "git+https://github.com/bitpay/bitauth.git#68cf0353bf517a7e5293478608839fa904351eb6",
-    "bitcore-wallet-client": "8.25.11",
+    "bitcore-wallet-client": "8.25.12",
```
cmgustavo, line 133: Should be 8.25.13

cmgustavo (approved):

PR #11837, opened 2021-08-23
Fix new feature slide

cmgustavo:


src/providers/app/app.ts
```diff
@@ -146,10 +146,14 @@ export class AppProvider {
...
-    );
+    const previousVersion =
+      parseInt(version.major, 10) +
+      parseInt(version.minor, 10) +
+      parseInt(version.patch, 10);
+    const currentVersion =
+      parseInt(targetVersion.major, 10) +
+      parseInt(targetVersion.minor, 10) +
+      parseInt(targetVersion.patch, 10);
+    return previousVersion < currentVersion;
```
JohnathanWhite, line 157: should be <=

src/providers/app/app.ts
```diff
@@ -146,10 +146,14 @@ export class AppProvider {
       return false;
     }
 
-    return (
-      version.major === targetVersion.major &&
-      version.minor === targetVersion.minor &&
-      version.patch === targetVersion.patch
-    );
+    const previousVersion =
```
JohnathanWhite, line 149: would rename this targetVersion

src/pages/home/home.ts
```diff
@@ -215,7 +215,7 @@ export class HomePage {
               }
             });
           } else {
-            this.persistenceProvider.setNewFeatureSlidesFlag(currentVs);
+            this.persistenceProvider.setNewFeatureSlidesFlag(dismissFlag);
```
JohnathanWhite, line 218: good catch

src/providers/app/app.ts
cmgustavo: Fixed. Thanks!

JohnathanWhite (approved):

PR #11855, opened 2021-08-31
[FIX] feature slide 

JohnathanWhite:
Ignores patches and solely checks that current major minor version === dismiss flag major minor


cmgustavo (approved):
LGTM

PR #11861, opened 2021-09-01
Ref: Tokens search

Gamboster:
The exact search result will be placed at the top of the results
Correction of errors in the styles of some icons
Fixed icons not showing (like EURT)
Some fixes in general


cmgustavo (approved):
LGTM! 💯

PR #11866, opened 2021-09-03
Fix: Coinbase integration

cmgustavo:
* Do not list duplicated accounts
* Filter custom ERC-20 tokens
* Get ordered accounts from Coinbase API


rkarendt (approved):

Gamboster (approved):
ACK

PR #11874, opened 2021-09-09
[REF] show spender and use correct icons

gabrielbazan7:
Copay_🔊


JohnathanWhite:
can we make the primary and secondary buttons here blue?

gabrielbazan7:
> can we make the primary and secondary buttons here blue?

They are blue for the BitPay build. That's a Copay screenshot.

BitPay

cmgustavo (approved):
LGTM

PR #11888, opened 2021-09-18
[FIX] invoice url scan

JohnathanWhite:
Invoice url's are not being opened from all scan points. This PR adds the check at the top of the handler.


cmgustavo (approved):
ACK

PR #11897, opened 2021-09-23
Fix: verify redirect host

lanchana:


src/components/info-sheet/info-sheet.html
```diff
@@ -527,6 +527,14 @@
       <span sheet-button-text translate>CONTINUE</span>
     </info-sheet-template>
 
+    <info-sheet-template *ngSwitchCase="'pairing-required'" type="warning">
+      <span sheet-title-text translate>Connect Your BitPay ID</span>
```
JohnathanWhite, line 531: let's change this to Verification Required

src/providers/incoming-data/incoming-data.ts
```diff
@@ -1561,20 +1574,13 @@ export class IncomingDataProvider {
...
-                {
-                  message: 'pairingOnly',
-                  payload: { paymentUrl: data }
-                },
-                () => {}
-              );
-            }, 100);
+          authRequiredInfoSheet.onDidDismiss(async () => {
+            await this.externalLinkProvider.open(
+              `https://${host}/id/verify?context=unlockv&id=${invoiceId}`
```
JohnathanWhite, line 1582: need to fix host here - host does not have link. subdomain removed so it does not redirect correctly

JohnathanWhite (approved):

PR #11899, opened 2021-09-24
Fix fluctuating home balance

msalcala11:
Incoming blocks are currently triggering an "update home balance" calculation, but this calculation only includes wallets that match the currency of the incoming block (resulting in an incorrect total balance shown on the home tab). This PR prevents incorrect home balance calculations by ensuring that the home balance only updates when all wallets (of all currencies) are included in the total balance calculation.


cmgustavo (approved):
ACK. Nice catch! 👍🏼

Gamboster (approved):
ACK!

PR #11935, opened 2021-10-18
[FEAT] RBF

gabrielbazan7:
BitPay

BitPay

BitPay

BitPay


rastajpa (approved):
ACK!

Gamboster (approved):
LGTM

JohnathanWhite:
Copy adjustments ->

If enabled, your transactions will be marked as non-final, and you will have the possibility, while they are unconfirmed, to replace them with transactions that pay higher fees. Note that some merchants do not accept non-final transactions until they are confirmed.

also 

This transaction can be accelerated using a higher fee.

Gamboster (approved):
ACK

PR #11939, opened 2021-10-21
[FIX] keep card IAB alive if analytics fails

cnguyen812:
Usually when the IAB emits a loaderror, we disable the IAB. Certain devices occasionally receive untrusted CA loaderrors for non-essential analytics URLs like this:  
```
{
  "type": "loaderror",
  "url": "https://px.ads.linkedin.com/collect/?time=12345&pid=12345&url=https%3A%2F%2Fbitpay.com%2Fwallet-card%2Fget-started&pageUrl=https%3A%2F%2Fbitpay.com%2Fwallet-card%2Fget-started&ref=&fmt=js&s=1",
  "code": 0,
  "sslerror": 3,
  "message": "The certificate authority is not trusted"
}
```

As a workaround I've added a whitelist to prevent the IAB from being disabled for known problematic URLs so those users can still access the card feature:
`['https://www.googletagmanager.com/gtag/js', 'https://px.ads.linkedin.com/collect/']`

### DIFF
--- a/src/providers/in-app-browser/in-app-browser.ts
+++ b/src/providers/in-app-browser/in-app-browser.ts
@@ -69,38 +69,50 @@ export class InAppBrowserProvider {
         }
       };
       ref.addEventListener('loadstop', initCb);
-      ref.addEventListener('loaderror', (err: InAppBrowserEvent & { sslerror?: number }) => {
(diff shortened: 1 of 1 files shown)


cmgustavo (approved):
LGTM

PR #11957, opened 2021-11-16
Fix: Prevent undefined decimals or infinite amounts for Tokens

Gamboster:
The error is given by the following:
We were allowing to create tokens already supported by bitpay as "Custom tokens". (USDC, DAI, ...)
There was an error in the creation of tokens when selecting a previously supported one (such as USDC) from the crypto swap feature. A confusion between the attribute ''decimal" and "decimals".
It is something rare, so there were few users reporting the error.
As a consequence, a "USDC Custom token" was saved with undefined decimals, so unitToSatoshi was undefined and when wanting to make calculations it resulted in infinite amounts.

Solution:

I have removed the tokens already supported by Bitpay from the list of custom tokens. Since they can be added on previous pages.

The data of the tokens already supported will have priority over those of the custom tokens (To prevent them from continuing to use undefined decimal places and causing errors when sending funds)

Fixed the use of the function "createToken" and "createCustomToken"

### DIFF
--- a/src/pages/add/custom-token/custom-token.ts
+++ b/src/pages/add/custom-token/custom-token.ts
@@ -76,11 +76,16 @@ export class CustomTokenPage {
     this.tokenSearchResults = this.filteredTokens;
     this.TOKEN_SHOW_LIMIT = 10;
     this.currentTokenListPage = 0;
+    const bitpaySupportedTokens: string[] = this.currencyProvider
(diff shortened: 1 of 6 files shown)


rastajpa (approved):
LGTM! Nice work!

cmgustavo (approved):

PR #11966, opened 2021-11-22
[FIX] keep nonce updated - take nonce from linked if erc20

gabrielbazan7:


Gamboster (approved):
ACK
I did several tests. It seems to be working very well

PR #11977, opened 2021-12-07
Feat: shib support for Simplex

Gamboster:
Also fixes icon in "buy crypto" and "Swap" button in wallet details


cmgustavo (approved):
LGTM

PR #11979, opened 2021-12-08
The android and ios trc fail in send

LastDragoon:
The coins are recieved but not to send so they stuck


paperchase4218:
How do I find and get my cash

omii2711 (approved):

PR #11988, opened 2021-12-21
Ref: enable XRP for non-US countries

cmgustavo:


rastajpa (approved):
ACK!

Gamboster:
It is also enabled for Crypto Swaps. We should remove it from there?

Gamboster:
Should we allow NON-US users to do Crypto Swaps with XRP? It is currently disabled for all

cmgustavo:
Closed in favor of https://github.com/bitpay/wallet/pull/11990

PR #12024, opened 2022-04-07
Fix: Changelly supported chains for swaps

Gamboster:
This PR solves the case in which tokens with the same currency abbreviation, but from a different chain than ERC20 are marked as supported, giving an error when obtaining the Changelly's rates for a certain pair, for example USDT


cmgustavo (approved):
LGTM

